PR #1, opened 2021-03-31
AWG and real signals

eggerdj:
### Summary

Changed the output of `BaseSignal.value` to return the real part of the signal only. Added a function to produce the I and Q signals from an AWG.

### Details and comments

The value of a signal is now always the real part. The plotting was also adapted accordingly and now gives the user the option to give axis.


qiskit_ode/converters/pulse_to_signals.py
```diff
@@ -123,3 +123,51 @@ def get_signals(self, schedule: Schedule) -> List[PiecewiseConstant]:
...
+            if_modulation: The intermediate frequency with which the AWG
+
+        Returns:
+            iq signals: A list of signals which is twice as long as the input list of signals.
+                For each input signal get_awg_signals returns two
+        """
+        new_signals = []
+
+        for sig in signals:
+            new_freq = sig.carrier_freq + if_modulation
```
DanPuzzuoli, line 147: For this line: can you clarify what the inputs/outputs to this function are meant to represent? It looks like it's outputting equations (12) and (13), with the input signal representing the complex valued `Omega(t) = f(t)exp(iwt)`? If yes, should this be subtracting or adding `if_modulation`? (Asking as in the notes it's a minus sign.)

qiskit_ode/signals/signals.py
```diff
@@ -88,39 +93,47 @@ def __add__(self, other):
     def __radd__(self, other):
         return signal_add(self, other)
 
-    def plot(self, t0: float, tf: float, n: int):
+    def plot(self, t0: float, tf: float, n: int, axis=None):
```
DanPuzzuoli, line 96: good idea

DanPuzzuoli (commented):
Aside from my comments this looks good. The plots in the demo look really good!

One thing I'm wondering with the AWG stuff (that is beyond the scope of this PR): how do we imagine a user modelling these things? I can imagine two cases:
1. A user is working with Pulse, so they start with pulse sequences that they import. These signals, `Re[f(t)exp(iwt)]`, are the desired outputs of the IQ mixer. In this PR (correct me if I'm wrong) you've given the function that generates the output of the AWG signal, with the input: the complex-valued signal representing the envelope of the signal you want to produce from the IQ mixer. Do we also need a function then that, given the signal `Re[f(t)exp(iwt)]`, intermediate frequency, and LO frequency that the mixer adds, prepares the input to your AWG function and maybe a model for the IQ mixer (or something along these lines), to help the user set up their modelling of the signal generation/mixing process?
2. Alternatively the user could just be directly building the inputs to this AWG function, in which case this is already enough!

qiskit_ode/converters/pulse_to_signals.py
```diff
@@ -123,3 +123,51 @@ def get_signals(self, schedule: Schedule) -> List[PiecewiseConstant]:
...
+        signals: List[PiecewiseConstant], if_modulation: float
+    ) -> List[PiecewiseConstant]:
+        """
+        Create signals that correspond to the output ports of an Arbitrary Waveform Generator
+        to be used with IQ mixers. For each signal in the list the number of signals is double
+        to create the I and Q components.
+
+        Args:
+            signals: A list of signals for which to create I and Q.
+            if_modulation: The intermediate frequency with which the AWG
```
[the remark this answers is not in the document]
eggerdj: Fixed here: e6c3fc3

qiskit_ode/converters/pulse_to_signals.py
eggerdj: Docstring has been updated 9c4eb56 . The notes subtracted the IF frequency but I updated it to an addition. For IQ mixers changing from plus to minus will typically switch from the right sideband to the left sideband.

DanPuzzuoli (approved):

PR #2, opened 2021-03-31
Improve performance on numpy backend 

vvilpas:
### Summary

Improve performance on numpy backend reducing the number of array objects construction and improving performance on array construction for numpy arrays. Construction performance is improved bypassing  `setattr` when we have a numpy array and we don't need to transform it to other data type or order. 

### Details and comments

Some plots showing performance gains when running **Numpy lsoda** section of **Benchmarking-CR_drive** notebook:

numpy_array_times_comp
numpy_array_times

### DIFF
--- a/qiskit_ode/dispatch/array.py
+++ b/qiskit_ode/dispatch/array.py
@@ -75,13 +75,9 @@ class Array(NDArrayOperatorsMixin):
         # set _data and _backend directly
         if (
             isinstance(data, numpy.ndarray)
-            and (backend == "numpy" or (not backend and Dispatch.DEFAULT_BACKEND == "numpy"))
+            and _is_numpy_backend(backend)
             and (not dtype or dtype == data.dtype)
-            and (
-                not order
-                or (order == "C" and data.flags["C_CONTIGUOUS"])
-                or (order == "F" and data.flags["F_CONTIGUOUS"])
-            )
+            and _is_equivalent_numpy_array(data, dtype, order)
         ):
             self.__dict__["_data"] = data
             self.__dict__["_backend"] = "numpy"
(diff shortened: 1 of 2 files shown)


DanPuzzuoli:
Still some linting issues, but the speed improvements look awesome!

One concern I have that we need to figure out is: I don't know if we can remove the `Array()` wrap on the outputs of the signal calling functions. How much is that particular change responsible for the speed improvements? 

We potentially need that to ensure proper dispatching occurs when doing computations on the outputs of the signals. I'm wondering though if we can define two versions of these functions, one that does the wrapping, and one that doesn't, and internally we can try to use the non-wrapped version as much as possible. E.g. in the `GeneratorModel` class, when calling `np.tensordot` to multiply/sum the signals with the operators, maybe if we make the operators the first argument (which it's no problem to wrap in `Array` as this only needs to be done once), then we don't need to wrap the signal outputs as the first argument determines the dispatching. Is this correct @chriseclectic ?

vvilpas:
I see. Most of performance gains comes from avoiding creating Arrays. I guess there is some mismatch in the code then, because in `BaseSignal` base class `envelope_value` and `value` methods are marked as returning `complex` values instead of `Array`s, and `Constant` for example returns a `complex` value as well. So just to confirm, even in the case of returning single values, we need to wrap them into `Array`s?

DanPuzzuoli:
Chris and I discussed this a bit on Friday, and we're thinking that, in light of the speed gains you've shown, it doesn't make sense to require `envelope_value` to output an `Array`. Internally `envelope_value` is actually only called from the `value` method, which I believe will output an `Array` anyway as the `arg` variable (computed in `value`) will be an `Array`. Also, the only time it is actually necessary to use the `Array` class is when using `jax`, and we can just offload the responsibility of wrapping the envelope function in an `Array` to the user.

All this is to say: it _isn't_ necessary to wrap the outputs of `envelope_value` in an `Array`, and so your PR will be good to merge once the lint tests are passing :).

DanPuzzuoli (approved):

PR #3, opened 2021-04-09
Signal sums and signal interface tweaks

DanPuzzuoli:
### Summary

This PR introduces the `SignalSum` object, containing a list of `Signal` instances and representing their sum. The purpose of this is to naturally preserve frequency information compartmentalization - i.e. if two different `Signal` instances are added together, they still remain internally separated in `SignalSum`, so that frequency-dependent operations/transformations (like RWA) can still analyze them separately. An additional `PiecewiseConstantSignalSum` object is introduced, which represents a sum of `PiecewiseConstant`s that have compatible sample structure; i.e. the same `dt`, `duration` and `start_time`. These both subclass `Signal`, to inherit much if the same functionality.

Using `SignalSum`, multiplication of two `Signal` instances is implemented to return a `SignalSum` that expands out the multiplication into two terms: one whose frequency is the sum of the frequencies of the product terms, and one which is the difference. Multiplication also has some `Signal`-type aware behavior: e.g. if the two signals being multiplied are `PiecewiseConstant` with compatible sample structure, it will implement the multiplication via multiplication of the sample arrays. This is in contrast to the general case: two arbitrarily user-defined `Signal`s, whose envelopes must be multiplied as functions.

Additionally, this PR makes modifications to the original `Signal` interface, the biggest of which is moving to vectorized evaluation.

Note:
- Due to significant changes, this code is put into a new file `new_signals.py`. Demos of the new functionality are presented in `example_notebooks/new_signals.ipynb`.
- The number of line changes is not representative of the actual number of changes: most are in the demo notebook, and the majority of lines in `new_signals.py` already exist, or are modifications of lines in `signals.py`. 

## Details

`Signal` interface
- Main methods:
  - `envelope(t)` - Vectorized evaluation of the complex envelope
  - `complex_value(t)` - Vectorized evaluation of `f(t)exp(i(2 * pi * v * t + phi))`
  - `__call__(t)` - Vectorized evaluation of `Re[f(t)exp(i(2 * pi * v * t + phi))]`
  - `conjugate()` - Return the signal with the complex conjugate of `f(t)` and phase `-phi`
  - `plot`, `plot_envelope`, `plot_complex_value` - plotting functions
- Main attributes:
  - `carrier_freq` - returns carrier freq
  - `phase` - returns phase
- This class also contains dunder methods for algebraic operations

Note: removed `value` in favour of `__call__`, and changed `envelope_value` to `envelope`.

`SignalSum(Signal)` interface
- Class representing a sum of any subclass of `Signal` (aside from `SignalSum` itself).
- Evaluation functions are modified naturally for a "sum":
  - `envelope` returns an array for each component in the sum for every time
  - `complex_value` and `__call__` return the sum of all signals in the sum.
- Additional attributes:
  - `components` - storage of the list signals in the sum.
- Additional methods:
  - `collapse()` - collapse the sum into a single `Signal`
  - `__getitem__(idx)` - returns the signal at the `idx` location in the sum
  - `__len__` - number of terms in the sum

`PiecewiseConstantSignalSum(SignalSum, PiecewiseConstant)`
- Class representing a sum of `PiecewiseConstant` instances, all of which have the same sample structure: same `dt`, same `duration`, and same `start_time`.
- Implements same interface as `SignalSum` with added `PiecewiseConstant` attributes.
- Stores envelope samples for all components in a single array, and all envelopes are evaluated simultaneously with a single array look up.

### Algebraic operations implementation

Algebraic operations are implemented via dunder methods in the base `Signal` class:
- `__add__`, `__radd__`, `__mul__`, `__rmul__`, `__neg__`, `__sub__`, and `__rsub__`
- Where relevant, the above operations also work with scalars (they get wrapped in `Constant`s)
- Addition is implemented via direct construction of a `SignalSum` object
  - If one (or both) of the signals being added together are a `SignalSum`, they will be appropriately unrolled.
- Multiplication is performed by constructing a `SignalSum` via the relation:
  - `Re[f(t)exp(i(wt + phi))] * Re[g(t)exp(i(vt + psi))] =` 
       `Re[0.5 *f(t)g(t)exp(i( (w+v)t + (phi + psi)))] + Re[0.5 *f(t)conj(g(t))exp(i( (w-v)t + (phi - psi)))]`
  - The above relation works for `SignalSum`s by performing the above operation for all pairs in the sums (distributivity of multiplication over sums).
- _Multiplication of envelopes_: the above formula involves multiplication of envelopes. Where possible, special rules apply: E.g. constants are multiplied by just multiplying their values. If compatible (based on start_time, dt, and number of samples), piecewise constants are multiplied by multiplying their sample arrays. If no special rules apply, the envelopes are multiplied as functions (e.g. any multiplication involving a generic `Signal`.)


DanPuzzuoli:
Going to remake this PR as a draft - forgot to make it a draft initially, but tests are still being run after changing it to a draft.

PR #4, opened 2021-04-19
Updating Signal interface and adding SignalSum to enable algebraic operations on Signals

DanPuzzuoli:
### Summary

This PR introduces the `SignalSum` object, containing a list of `Signal` instances and representing their sum. The purpose of this is to naturally preserve frequency information compartmentalization - i.e. if two different `Signal` instances are added together, they still remain internally separated in `SignalSum`, so that frequency-dependent operations/transformations (like RWA) can still analyze them separately. An additional `DiscreteSignalSum` object is introduced, which represents a sum of `DiscreteSignal`s that have compatible sample structure; i.e. the same `dt`, `duration` and `start_time`. These both subclass `Signal`, to inherit much if the same functionality.

Using `SignalSum`, multiplication of two `Signal` instances is implemented to return a `SignalSum` that expands out the multiplication into two terms: one whose frequency is the sum of the frequencies of the product terms, and one which is the difference. Multiplication also has some `Signal`-type aware behavior: e.g. if the two signals being multiplied are `DiscreteSignal` with compatible sample structure, it will implement the multiplication via multiplication of the sample arrays. This is in contrast to the general case: two arbitrarily user-defined `Signal`s, whose envelopes must be multiplied as functions.

Additionally, this PR makes modifications to the original `Signal` interface, the biggest of which is moving to vectorized evaluation.

## Details

`Signal` interface
- Main methods:
  - `envelope(t)` - Vectorized evaluation of the complex envelope
  - `complex_value(t)` - Vectorized evaluation of `f(t)exp(i(2 * pi * v * t + phi))`
  - `__call__(t)` - Vectorized evaluation of `Re[f(t)exp(i(2 * pi * v * t + phi))]`
  - `conjugate()` - Return the signal with the complex conjugate of `f(t)` and phase `-phi`
  - `plot`, `plot_envelope`, `plot_complex_value` - plotting functions
- Main attributes:
  - `carrier_freq` - returns carrier freq
  - `phase` - returns phase
- This class also contains dunder methods for algebraic operations

Note: removed `value` in favour of `__call__`, and changed `envelope_value` to `envelope`.

Changed `PiecewiseConstant` to `DiscreteSignal`.

`SignalSum(Signal)` interface
- Class representing a sum of any subclass of `Signal` (aside from `SignalSum` itself).
- Evaluation functions are modified naturally for a "sum":
  - `envelope` returns an array for each component in the sum for every time
  - `complex_value` and `__call__` return the sum of all signals in the sum.
- Additional attributes:
  - `components` - storage of the list signals in the sum.
- Additional methods:
  - `flatten()` - collapse the sum into a single `Signal`. This destroys frequency information so should be used selectively. This is particularly useful initially as none of the rest of the code has been written to support `SignalSum`s.
  - `__getitem__(idx)` - returns the signal at the `idx` location in the sum. Based on input, this has been written to enable numpy-style indexing. E.g. `idx` can be an integer, a slice, a list of integers, or an array of integers.
  - `__len__` - number of terms in the sum
  - `draw(t0, tf, N, function, axis)` - plot the signal. The `function` option specifies which "function" to plot: i.e. the full signal, the complex envelope, or the complex value.

`DiscreteSignalSum(SignalSum, DiscreteSignal)`
- Class representing a sum of `DiscreteSignal` instances, all of which have the same sample structure: same `dt`, same `duration`, and same `start_time`.
- Implements same interface as `SignalSum` with added `DiscreteSignal` attributes.
- Stores envelope samples for all components in a single array, and all envelopes are evaluated simultaneously with a single array look up.

`SignalList`
- This replaces `VectorSignal`, and is meant to store a list of `Signal`s (and any instances of its subclasses). This has more of a limited interface and currently I only expect this to be used internally.
- `complex_value` - evaluate the complex value (this returns a vector for each element in the list)
- `__call__` - evaluate the list.
- `flatten` - flattens the components.
- `drift` - Evaluates only the constant components.
- `__getitem__` indexing - similar to `SignalSum`, is set up to do fancy indexing.
- `__len__` number of components.

Additional note:
- The `to_pwc` method of `Signal` has been removed, and the functionality has been moved into the constructor `DiscreteSignal.from_Signal`. A similar static method works for `SignalSum`s: `DiscreteSignalSum.from_SignalSum`.

### Algebraic operations implementation

Algebraic operations are implemented via dunder methods in the base `Signal` class:
- `__add__`, `__radd__`, `__mul__`, `__rmul__`, `__neg__`, `__sub__`, and `__rsub__`
- Where relevant, the above operations also work with scalars (they get wrapped in `Constant`s)
- Addition is implemented via direct construction of a `SignalSum` object
  - If one (or both) of the signals being added together are a `SignalSum`, they will be appropriately unrolled.
- Multiplication is performed by constructing a `SignalSum` via the relation:
  - `Re[f(t)exp(i(wt + phi))] * Re[g(t)exp(i(vt + psi))] =` 
       `Re[0.5 *f(t)g(t)exp(i( (w+v)t + (phi + psi)))] + Re[0.5 *f(t)conj(g(t))exp(i( (w-v)t + (phi - psi)))]`
  - The above relation works for `SignalSum`s by performing the above operation for all pairs in the sums (distributivity of multiplication over sums).
- _Multiplication of envelopes_: the above formula involves multiplication of envelopes. Where possible, special rules apply: E.g. constants are multiplied by just multiplying their values. If compatible (based on start_time, dt, and number of samples), piecewise constants are multiplied by multiplying their sample arrays. If no special rules apply, the envelopes are multiplied as functions (e.g. any multiplication involving a generic `Signal`.)

### Additional review notes

I've updated all other files that used `Signal` and its subclasses, along with all existing tests, which are now passing. I've also updated the example notebooks so they all run with the current interface. 

The notebook `new_signals.ipynb` demos the new interfaces and functionality. 

Still to do:
- Add new tests for new functionality (especially for vectorized evaluation). 
- To get things running I've had to do a bit of `Array` hackery again. It might be good to start putting explicit jax implementations of some components to avoid this as much as possible. Should we just check the default backend and use jax logic if it's `'jax'`?
- Once things are settled I'll update the original/merge the original `Signals.ipynb` demo with some new stuff in `new_signals.ipynb` to finish off the PR.

I haven't implemented absolutely everything we discussed, but I think at this point, if we're happy with this structure, we can leave subsequent iterative changes for subsequent PRs that can be worked on in parallel with other things that depend on the `Signal` structure.


DanPuzzuoli:
To do since last discussion (crossed out means completed):

- Rename PiecewiseConstant -> DiscreteSignal
    - Same with PiecewiseConstantSignalSum
- ~Remove `np.vectorize` option from Signal - put in the docs that this is assumed~
- Plotting (may leave this for the end)
    - Return axis object?
    - Defaults: add a grid
    - Defaults: Add legend for real/imaginary where relevant
    - Rename plot -> draw
    - Have a single draw function and just specify which via a “method” kwarg or something like this
- repr
    - Make it look like the default python repr (with angle brackets and what not) but a bit more concise
- Change collapse -> flatten or reduce
- Change to_pwc to discretize
    - Make it a class method of Discrete signal and of DiscreteSignalSum
- Properly flesh out `SignalList` object and update locations that used `VectorSignal`
- Write implementation/test everything with jax!!!
    - Can we use dispatch default backend to choose jax behaviour?
- Add everything to docs

qiskit_ode/signals/signals.py
```diff
@@ -27,167 +28,196 @@
...
 
-        Args:
-            dt: Time increment to use.
-            n_samples: number of steps to resample with.
-            start_time: start time from which to resample.
+    @carrier_freq.setter
+    def carrier_freq(self, carrier_freq: float):
+        """Carrier frequency setter."""
+        if type(carrier_freq) == list:
+            carrier_freq = [Array(entry).data for entry in carrier_freq]
```
eggerdj, line 102: Something seems inconsistent here between the type `carrier_freq: float` and the immediate list check.

qiskit_ode/signals/signals.py
```diff
@@ -27,167 +28,196 @@
...
-        """Return a new signal that is the complex conjugate of self."""
+        if callable(envelope):
+            self._envelope = envelope
 
-    @abstractmethod
-    def envelope_value(self, t: float = 0.0) -> complex:
-        """Evaluates the envelope at time t."""
+        # initialize internally stored carrier/phase information
+        self._carrier_freq = None
+        self._phase = None
```
eggerdj, line 80: Is this needed if you set them immediately after? Or does lint or the editor complain?

qiskit_ode/signals/signals.py
```diff
@@ -27,167 +28,196 @@
...
+        return self._phase
 
-        return PiecewiseConstant(
-            dt, samples, start_time=start_time, carrier_freq=self.carrier_freq, phase=self.phase
-        )
+    @phase.setter
+    def phase(self, phase: float):
+        """Phase setter."""
+        if type(phase) == list:
+            phase = [Array(entry).data for entry in phase]
```
eggerdj, line 115: Same comment as above.

qiskit_ode/signals/signals.py
```diff
@@ -27,167 +28,196 @@
...
+    def __call__(self, t: Union[float, np.array, Array]) -> Array:
+        """Vectorized evaluation of the signal at time t."""
+        return np.real(self.complex_value(t))
 
-    def __radd__(self, other):
-        return signal_add(self, other)
+    def __str__(self) -> str:
+        """Return string representation."""
+        if self.name is not None:
+            return str(self.name)
```
eggerdj, line 135: Would it perhaps not be nicer to return
```
Signal(carrier_freq={freq}, phase={phase}, name={name})
```
if it has a name?

qiskit_ode/signals/signals.py
```diff
@@ -27,167 +28,196 @@
...
 
-        Args:
-            t0: initial time
-            tf: final time
-            n: number of points to sample in interval.
-            axis (AxesSubplot): the axis to use for plotting.
-        """
-        x_vals = np.linspace(t0, tf, n)
+    def __neg__(self) -> "SignalSum":
+        return -1 * self
```
eggerdj, line 154: Does this have to be a `SignalSum`? Could it not be a `Signal`?

qiskit_ode/signals/signals.py
```diff
@@ -247,8 +279,53 @@ def __init__(
...
+        """
+        Constructs a ``DiscreteSignal`` signal object by sampling a ``Signal``.
+
+        Args:
+            signal: Signal to sample.
+            dt: Time increment to use.
+            n_samples: Number of steps to resample with.
+            start_time: Start time from which to resample.
+            sample_carrier: Whether or not to keep the carrier analog or include it in the
+                            sampling.
```
eggerdj, line 310: `keep the carrier along` is a bit vague

qiskit_ode/signals/signals.py
```diff
@@ -247,8 +279,53 @@ def __init__(
...
+
+        Returns:
+            DiscreteSignal: The discretized ``Signal``.
+        """
+
+        times = start_time + (np.arange(n_samples) + 0.5) * dt
```
eggerdj, line 316: Perhaps add in the docstring what sampling methodology we use. Here, it appears to be at the mid-point.

qiskit_ode/signals/signals.py
```diff
@@ -282,24 +359,27 @@ def start_time(self) -> float:
...
-        return PiecewiseConstant(
+        return DiscreteSignal(
             dt=self._dt,
             samples=np.conjugate(self._samples),
             start_time=self._start_time,
             duration=self.duration,
-            carrier_freq=-self.carrier_freq,
+            carrier_freq=self.carrier_freq,
             phase=-self.phase,
         )
```
eggerdj, line 384: How do we interpret `conjugate` for a real signal?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
-        \Omega(t)*exp(2\pi i \nu t)
+        return "DiscreteSignal(dt={dt}, carrier_freq={freq}, phase={phase})".format(
+            dt=self.dt, freq=str(self.carrier_freq), phase=str(self.phase)
+        )
 
-    multiplication of two signals implements
 
-    .. math::
+class SignalSum(Signal):
+    """Class representing a sum of ``Signal`` objects."""
```
eggerdj, line 421: I think this requires more details. Perhaps a few words on its use and how it is implemented.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
 
-        \Omega_1(t)*\Omega_2(t)*exp(2\pi i (\nu_1+\nu_2) t)
+    def __init__(self, *args, name: Optional[str] = None):
+        """Initialize with a list of Signal objects through ``args``.
 
+        Args:
+            args: ``Signal`` subclass objects.
```
eggerdj, line 427: Why not call this `signals`?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        for sig in self.components:
+            carrier_freqs.append(sig.carrier_freq)
+
+        phases = []
+        for sig in self.components:
+            phases.append(sig.phase)
+
+        # set carrier and phase
+        self.carrier_freq = carrier_freqs
+        self.phase = phases
```
eggerdj, line 464: So this explains the list check in the setter. I think the type hints need to be changed in `Signal`.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+
+    def envelope(self, t: Union[float, np.array, Array]) -> Array:
+        """Evaluate envelopes of each component. For vectorized operation,
+        last axis indexes the envelope, and all proceeding axes are the
+        same as the ``t`` arg.
+        """
```
eggerdj, line 470: This a bit unclear.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        # to do: jax version
+        # not sure what the right way to do this is, here we actually need
+        # to get it to use np/jnp
```
eggerdj, line 473: Can you some how do something like checking if jax is installed and then point to it? E.g. a bit like what is often done with `HAS_MATPLOTLIB`?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        return np.moveaxis(Array([env(t) for env in self._envelopes]), 0, -1)
+
+    def complex_value(self, t: Union[float, np.array, Array]) -> Array:
```
eggerdj, line 476: docstring?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
-    r"""Implements mathematical addition between two signals.
-    Since a signal is represented by
-    .. math::
+        return default_str
 
-        \Omega(t)*exp(2\pi i \nu t)
+    def flatten(self) -> Signal:
+        """Merge into a single ``Signal``. The output frequency is given by the
+        average.
+        """
```
eggerdj, line 525: Would it make sense to have the average frequency as the default but then let the user override this? For example, with a mixer I probably do not want to have the average as the carrier frequency. We could also think of methods to have the carrier at the frequency where the power spectral density of the signal is highest.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        return Signal(envelope=merged_env, carrier_freq=ave_freq, name=str(self))
 
-    Returns:
-        signal: The type will depend on the given base class.
 
-    Raises:
-        Exception: if signals cannot be added.
+class DiscreteSignalSum(DiscreteSignal, SignalSum):
+    """Represents a sum of piecewise constant signals, all with the same
+    time parameters: dt, number of samples, and start time.
```
eggerdj, line 544:
```suggestion
    """Represents a sum of discretized signals, all with the same
    time parameters: dt, number of samples, and start time.
```

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        dt: float,
+        samples: Union[Array, List],
+        start_time: float = 0.0,
+        duration: int = None,
+        carrier_freqs: float = None,
+        phases: float = None,
+        name: str = None,
+    ):
+        """Samples array has 0th axis corresponding to a signal, 1st axis corresponding to samples
+        for each signal."""
```
eggerdj, line 558: This needs more detail.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        self.phase = phases
 
-    def __init__(
-        self,
-        envelope: Callable,
-        carrier_freqs: Array,
-        phases: Array,
-        drift_array: Optional[Array] = None,
+    @classmethod
+    def from_SignalSum(
```
eggerdj, line 596: Does this need to do any validation to check the assumptions of DiscreteSignal sum or does it enforce them by sampling the given `SignalSum`? Perhaps this can be clarified in the docstring?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
-            t: time
+class SignalList:
+    """A list of ``Signal``s, with functionality for simultaneous evaluation."""
 
-        Returns:
-            Array: the signal envelope at time t
-        """
-        return self.envelope(t)
+    def __init__(self, signal_list: List[Signal]):
+        self.components = signal_list
```
eggerdj, line 689: Docstring

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
-            VectorSignal: the complex conjugate of self
-        """
-        return VectorSignal(
-            lambda t: np.conjugate(self.envelope_value(t)),
-            -self.carrier_freqs,
-            -self.phases,
-            np.conjugate(self.drift_array),
+    @property
+    def drift(self) -> Array:
+        """Return the 'drift' Array, i.e. the constant part of the ``SignalList``."""
```
eggerdj, line 714: By `constant part` do you mean the sub-list of all constant signals? Can you clarify? So if understand this part it will allow us to conveniently separate drift and control parts of a Hamiltonian?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+            for term in sig_entry:
+                if isinstance(term, Constant):
+                    val += term._value.data
+
+            drift_array.append(val)
+
+        return Array(drift_array)
+
+    def __getitem__(self, idx: Union[int, List, np.array, slice]) -> Signal:
+        """Enables numpy-style subscripting, as if this class were a 1d array."""
```
eggerdj, line 733: This looks like a copy of `__getitem__` from `SignalSum`. It might make sense to introduce an abstract base class to remove some of this redundency.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+        if isinstance(sublist, Signal):
+            return sublist
+        else:
+            return SignalList(*sublist)
+
+    def __len__(self):
+        return len(self.components)
+
+
+def signal_add(sig1: Signal, sig2: Signal) -> SignalSum:
```
eggerdj, line 763: To keep things more modular it might make sense to put these type of functions in a `signals_operations.py` file.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+    """Add two signals."""
+
+    # generic routine
+    # convert to SignalSum instances
+    try:
+        wrapped_sig1 = to_SignalSum(sig1)
+        wrapped_sig2 = to_SignalSum(sig2)
+    except QiskitError as e:
```
eggerdj, line 771: Does our lint allow variables name as a single letter?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+
+    Special cases:
+
+        - Multiplication of two ``Constant``s returns a ``Constant``.
+        - Multiplication of a ``Constant`` and a ``DiscreteSignal`` returns a ``DiscreteSignal``.
+        - If two ``DiscreteSignal``s have compatible parameters, the resulting signals are
+        ``DiscreteSignal``, with the multiplication being implemented by array multiplication of
+        the samples.
+        - Lastly, if no special rules apply, the two ``Signal``s are multiplied generically via
+        multiplication of the envelopes as functions.
```
eggerdj, line 846: What happens to carrier phase and frequency?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+    )
+    prod2 = Signal(
+        envelope=new_env2,
+        carrier_freq=sig1.carrier_freq - sig2.carrier_freq,
+        phase=sig1.phase - sig2.phase,
+    )
+    return prod1 + prod2
+
+
+def sort_signals(sig1: Signal, sig2: Signal) -> Tuple[Signal, Signal]:
```
eggerdj, line 923: What does this enable?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +407,570 @@ def add_samples(self, start_sample: int, samples: List):
...
+
+    Returns:
+        SignalSum
+
+    Raises:
+        QiskitError: If the input type is incompatible with SignalSum.
+    """
+
+    if isinstance(sig, (int, float, complex)) or (isinstance(sig, Array) and sig.ndim == 0):
+        sig = Constant(sig)
```
eggerdj, line 968: Is `Constant` a `SignalSum`? Does this work with the error raised below?

eggerdj (changes requested):
Looking good. See comments. It might also be worth it to
* Add a legend to the plot based on the signal names if defined.
* Add the name of the signal to the `__repr__`

qiskit_ode/signals/signals.py
DanPuzzuoli: Ah yeah, right. I had added the list check to support subclasses `SignalSum` and `DiscreteSignalSum`, where `carrier_freq` and `phase` are either lists/Arrays. I could add `list` as an acceptable type here but maybe it makes more conceptual sense to just have `SignalSum` directly implement the list/array versions of these getter/setters.

qiskit_ode/signals/signals.py
DanPuzzuoli: Just removed these lines and the linter doesn't complain, so I'll get rid of them!

qiskit_ode/signals/signals.py
DanPuzzuoli: Same reply: moved the list handling version to `SignalSum`.

qiskit_ode/signals/signals.py
DanPuzzuoli: This does actually return a `SignalSum` - the expression `-1 * self` will result in a call to `self.__rmul__`, which returns a `SignalSum`.

qiskit_ode/signals/signals.py
DanPuzzuoli: This is a good point:
- In this PR, for a signal `Re[f(t)exp(i(wt + phi))]`, conjugation maps the signal to one implementing the function `Re[conj(f(t))exp(i(wt - phi))]`; basically taking the conjugate of the envelope, and conceptually including the phase as part of that. 
- I chose this as it is useful in implementing multiplication of signals, as we do really need to just take the conjugate of the envelope/phase. Also, since signals are explicitly real now, also taking `w` -> `-w` would result in the same real-valued function.
- However, this behaviour is definitely weird with the method `complex_value`, as with this definition `sig2 = sig1.conjugate()` won't satisfy `sig2.complex_value(t) == np.conj(sig1.complex_value(t))`.

Some options: 
- We could change `.conjugate()` to `.conjugate_envelope()` to be explicit about this. 
- Or `.conjugate()` could be expanded to have optional arguments about which parameters in the complex representation are being "conjugated". E.g. it could have a signature: 
```
def conjugate(self, envelope=True, carrier_freq=True, phase=True):
```

My preference would be the latter option.

qiskit_ode/signals/signals.py
DanPuzzuoli: Updated this, as well as the corresponding text in `DiscreteSignalSum.from_SignalSum` to be more explicit.

qiskit_ode/signals/signals.py
DanPuzzuoli: Added this.

qiskit_ode/signals/signals.py
DanPuzzuoli: Updated.

qiskit_ode/signals/signals.py
DanPuzzuoli: Changed to `signals`

qiskit_ode/signals/signals.py
DanPuzzuoli: Changed them in `Signal`.

qiskit_ode/signals/signals.py
DanPuzzuoli: I've updated this. I've actually removed most of the description - what I was describing was basically what "vectorized" means, which maybe isn't actually necessary here.

qiskit_ode/signals/signals.py
DanPuzzuoli: yeah - I think the way we need to do this is to just look at `jax.default_backend()`. Something we need to be careful about though is what types we expect the user-defined envelopes to return. (i.e. do we demand they return an `Array` if using jax?)

qiskit_ode/signals/signals.py
DanPuzzuoli: Added.

qiskit_ode/signals/signals.py
DanPuzzuoli: I agree, though I think this requires enough thought that it warrants its own subsequent PR. Currently I'm imagining:
- `flatten` is a brute force function that forces a sum into a single `Signal`, and should be avoided where possible. I view this mainly as a stopgap for the near-term until we update the rest of the repo to work with `SignalSum` objects.
- I think we should add a `simplify` method (in a subsequent PR) that more cleverly attempts to merge signals in a sum based on either their PSD, or on their `carrier_freq` values (and we can have a tolerance here).

I think for AWG implementation we want to use `simplify`, so that it doesn't merge components that are too far apart in frequency space.

qiskit_ode/signals/signals.py
DanPuzzuoli: Updated this doc string.

qiskit_ode/signals/signals.py
DanPuzzuoli: Added.

qiskit_ode/signals/signals.py
DanPuzzuoli: I've updated this to try to be more clear.

And yes, you're correct, the ultimate use case is to be able to determine the "constant part" of a model (like a Hamiltonian).

qiskit_ode/signals/signals.py
DanPuzzuoli: Yeah possibly - we could abstract out `__len__` as well. I'll think a bit, but it's probably worth it just for `__getitem__` alone since it's so long.

qiskit_ode/signals/signals.py
DanPuzzuoli: Just tried but got circular import issues, so I think we'll need to keep these together.

qiskit_ode/signals/signals.py
DanPuzzuoli: No but I disabled name checking for the whole file this as the argument name `t` doesn't conform to this. Updated though to be called `qe`.

qiskit_ode/signals/signals.py
DanPuzzuoli: Updated comment to explain.

qiskit_ode/signals/signals.py
DanPuzzuoli: This is just to enable more succinct type checking for binary operations on signals that care about special cases. E.g. multiplication has a special rule for if one `sig1` and `sig2` is a `Constant` and the other is a more general `Signal`. By sorting `sig1` and `sig2` first, the multiplication routine only needs to check the single case: 
```
type(sig1) is Constant and type(sig2) is Signal
```

qiskit_ode/signals/signals.py
DanPuzzuoli: The subsequent check wraps any subclass of `Signal` that is not a `SignalSum` in a `SignalSum`, so this will handle the `Constant` case.

qiskit_ode/signals/signals.py
DanPuzzuoli: My thinking was that if it has a name it should just be displayed to make it more succinct, but I could go either way on this.

haggaila:
I have two small questions:
1. When `conjugate`'ing, shouldn't the conjugate of exp{i v t + phi} negate v as well (in addition to phi)?
2. In the new_signals.ipynb demo, the plots of the discretized signals have a slope between changing values, which is not the way I understand a discrete signal (which must jump sharply). Could it be an artifact of the plotting?

DanPuzzuoli:
@haggaila 

> I have two small questions:
> 
> 1. In the documentation above it is not mentioned that `conjugate()`, negates the frequency as well (like the code does).

`.conjugate()`  _used_ to negate frequency as well, but this PR has changed it so that it only conjugates the envelope/phase. I'm going to change it back to also negating the frequency though, based on your and @eggerdj 's comments. (I had a reason for changing this, but I'm realizing now that for the application I had in mind it doesn't make a difference, and your comments have made it clear that it is kind of confusing.)

> 2. In the new_signals.ipynb demo, the plots of the discretized signals have a slope between changing values, which is not the way I understand a discrete signal (which must jump sharply). Could it be an artifact of the plotting?

No this is the correct behaviour. This is arising in plots of `DiscreteSignal`s in which only the envelope has been discretized (so the carrier wave is still analog). The notebook also shows instances of discretizing the carrier wave as well: when plotted they are clearly fully piecewise constant. It's useful to have both behaviours.

DanPuzzuoli:
I've made updates based on suggestions. @eggerdj I've replied to your comments - either directly implementing your suggestions or giving some explanation as to what is going on.

Other changes of note:
- Created `SignalCollection` class for abstracting the internal list handling (`__len__`, `__getitem__`, etc) from `SignalSum`, `SignalList`, and `DiscreteSignalSum`.
- Updated `draw` to have a title/legend. The title is based on which function is being + the `str(self)`, and the legend shows Real/Imag when plotting complex data.
- JAX/`Array` handling
   - I've changed all of the signal evaluation functions (`envelope`, `complex_value`, and `__call__`) to only do `Array` wrapping when the backend is `jax`. This way, there shouldn't be any `Array` overhead when using straight numpy. I think maybe this can be used as a "general principle" when coding with the `dispatch`: for functions that contain both numpy/jax implementations, the return type can be `Union[np.array, Array]`, with the presumption that `numpy` implementation may return an `np.array`, and `jax` implementation (or another if added) will return an `Array`. This makes the type hinting a bit less concise but I do think it's perhaps "natural" in that, obviously, the `Array` wrapping is totally unnecessary for the pure `numpy` implementation.
   - In relation to the above, there is now a helper function at the end of `signals.py`, `array_funclist_evaluate`, which sets evaluation of a list of functions for which `Array` wrapping is necessary. `SignalSum` and `SignalList` both use this to setup how they evaluate `envelope`, `complex_value`, and `__call__` when operating in JAX-mode. This function is super small/simple, but it's useful to abstract this out into a helper function as it needs to occur in multiple places.

Question @eggerdj : 
- For `__repr__`, did you mean to actually implement `__repr__` or was that question related to `__str__`? I replied to your comment about `__str__` that I want the `name` attribute to override the behaviour of `__str__` as a mechanism for making the printing of `Signal` names a bit more concise. Thoughts?

haggaila:
I have a small suggestion, to have two setters and getters for the carrier frequency in `BaseSignal`, one for frequency and the other for angular frequency (with the 2 \pi factor). To backup the necessity of this - backends report frequencies in both formats (and even for the same physical variable);
For example the qubit anharmonicity is a frequency in `backend.properties().qubit_property(1, 'anharmonicity')`, but an angular frequency in `backend.configuration().hamiltonian['vars']['delta0']`.

qiskit_ode/signals/signals.py
eggerdj: Ok, sounds good.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +447,693 @@ def add_samples(self, start_sample: int, samples: List):
...
+
+    def __getitem__(
+        self, idx: Union[int, List, np.array, slice]
+    ) -> Union[Signal, "SignalCollection"]:
+        """Get item with Numpy-style subscripting, as if this class were a 1d array."""
+
+        if type(idx) == np.ndarray and idx.ndim > 0:
+            idx = list(idx)
+
+        sublist = None
```
eggerdj, line 488:
```suggestion
```

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +447,693 @@ def add_samples(self, start_sample: int, samples: List):
...
-        def env_func(t):
-            return Array([sig.envelope_value(t) for sig in signal_list])
+        times = start_time + (np.arange(n_samples) + 0.5) * dt
 
-        # construct carrier frequency and phase list
-        # if signal doesn't have a carrier, set to 0.
-        carrier_freqs = Array(
-            [Array(getattr(sig, "carrier_freq", 0.0)).data for sig in signal_list]
+        freq = signal_sum.carrier_freq
+        samples = None
```
eggerdj, line 726:
```suggestion
```

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +447,693 @@ def add_samples(self, start_sample: int, samples: List):
...
+            samples = signal_sum.envelope(times) * exp_phases
+        else:
+            samples = signal_sum.envelope(times)
+
+        return DiscreteSignalSum(
+            dt, samples, start_time=start_time, carrier_freq=freq, phase=signal_sum.phase
         )
 
-        phases = Array([Array(getattr(sig, "phase", 0.0)).data for sig in signal_list])
+    def complex_value(self, t: Union[float, np.array, Array]) -> Union[complex, np.array, Array]:
```
eggerdj, line 739: docstring?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +447,693 @@ def add_samples(self, start_sample: int, samples: List):
...
     def __init__(
         self,
-        envelope: Callable,
-        carrier_freqs: Array,
-        phases: Array,
-        drift_array: Optional[Array] = None,
+        dt: float,
+        samples: Union[List, Array],
+        start_time: float = 0.0,
+        duration: int = None,
```
eggerdj, line 638: It does not seem like this class is using duration. Is this needed here?

test/ode/signals/test_signals.py
```diff
@@ -17,213 +17,843 @@
...
+    def test_conjugate(self):
+        """Verify conjugate() functioning correctly."""
+        discrete_conj = self.discrete2.conjugate()
+        self.assertAllClose(discrete_conj.samples, np.conjugate(self.discrete2.samples))
+        self.assertAllClose(discrete_conj.carrier_freq, -self.discrete2.carrier_freq)
+        self.assertAllClose(discrete_conj.phase, -self.discrete2.phase)
+        self.assertAllClose(discrete_conj.dt, self.discrete2.dt)
+
+
+class TestSignalSum(QiskitOdeTestCase):
```
eggerdj, line 488: Perhaps add a test to check that `self.sig_sum1 - self.sig_sum2` behaves as expected?

eggerdj (changes requested):
Looking good. I noticed a few minor things.

qiskit_ode/signals/signals.py
```diff
@@ -11,183 +11,230 @@
...
 
+    def conjugate(self):
+        """Return a new signal whose complex value is the complex conjugate of this one."""
 
-class Signal(BaseSignal):
-    """The most general mixed signal type, represented by a callable
-    envelope function and a carrier frequency.
-    """
+        def conj_env(t):
+            return np.conjugate(self.envelope(t))
```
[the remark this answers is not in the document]
DanPuzzuoli: Yeah: in the `__init__` I've set it so that if jax is being used, the output of the envelope function will always be wrapped in an `Array`, so here the dispatch will implement `np.conjugate` via `jnp.conjugate`.

qiskit_ode/signals/signals.py
```diff
@@ -11,183 +11,230 @@
...
-    def conjugate(self):
-        """Return a new signal that is the complex conjugate of self."""
+    The envelope function can be complex-valued, and the frequency and phase must be real.
 
-    @abstractmethod
-    def envelope_value(self, t: float = 0.0) -> complex:
-        """Evaluates the envelope at time t."""
+    Note: this class assumes that the envelope function is vectorized. If it isn't, it can
+    be vectorized automatically by calling ``numpy.vectorize``\, or, if using JAX, by
+    calling ``jax.numpy.vectorize``\.
```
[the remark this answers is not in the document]
DanPuzzuoli: Added.

qiskit_ode/signals/signals.py
DanPuzzuoli: Yeah it isn't ever used. It only serves as a possible convenience argument for specifying a `DiscreteSignal` of a given length where all samples are `0`. I'll remove it for now and we can see if we ever end up actually needing that.

test/ode/signals/test_signals.py
DanPuzzuoli: Added `self.double_sig_sum = self.sig_sum1 - self.sig_sum2` in the `setUp` and then added the equivalent checks for `self.double_sig_sum` in each test case.

qiskit_ode/signals/signals.py
DanPuzzuoli: added

qiskit_ode/signals/signals.py
```diff
@@ -196,23 +243,28 @@ def __init__(self, value: complex, name: str = None):
...
         """
-        self._value = value
-        super().__init__(name)
-
-    def envelope_value(self, t: float = 0.0) -> complex:
-        return self._value
+        self._name = name
+        self._value = np.real(Array(value))
+        self.phase = 0.0
+        self.carrier_freq = 0.0
```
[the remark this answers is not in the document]
DanPuzzuoli: Hmm interesting. Yeah it works, changed it to this. I also removed some of the methods as they become redundant after making this change. This has left `Constant` pretty much redundant. 

Should we just get rid of `Constant`? I liked having it be obviously different from `Signal`, but given that the class is basically empty, maybe that's not a good enough reason. I think one potential way of doing this is to add an `is_constant` property to `Signal` (set to `True` if the envelope is specified as a number with 0 freq), which can be used to check if something is constant (as opposed to the type checks that check for `Constant` instances).

Thoughts?

DanPuzzuoli:
> I have a small suggestion, to allow setting and getting the carrier frequency both as a frequency and as an angular frequency (so the user wouldn't have to mess with the 2 \pi factor). To backup the necessity of this - backends report frequencies in both formats (and even for the same physical variable);
> For example the qubit anharmonicity is a frequency in `backend.properties().qubit_property(1, 'anharmonicity')`, but an angular frequency in `backend.configuration().hamiltonian['vars']['delta0']`.
> However, I see that currently the frequency can only be set in Signal, and only via the constructor, so it could be complicated to support both.

Yeah I think for now I'd rather keep it simpler and just have one way of specifying it. If it ends up being annoying we can change it later.

qiskit_ode/signals/signals.py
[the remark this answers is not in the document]
haggaila: I can't see any argument in favor of one or the other..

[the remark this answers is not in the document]
DanPuzzuoli: I think I may try getting rid of `Constant`. My only initial resistance to it was that I didn't like the idea of a user specifying `Signal(1.)` to represent the constant `1.` - the usage of the word "signal" when setting a constant seemed weird. I think though that we can just completely eliminate the need for the user to ever construct a `Signal` object for constants: they can just leave them as numbers (floats, complex, whatever), and the conversion to `Signal` objects that are constant can all happen behind the scenes. This already occurs for addition/multiplication: if you do `4 + signal`, the `4` is automatically converted to a `Constant`. The only other place this auto conversion needs to be added is in `SignalList` (relevant for specifying model coefficients).

DanPuzzuoli:
@eggerdj 

ready for re-review

test/ode/signals/test_signals.py
eggerdj: Thanks!

test/ode/signals/test_signals.py
```diff
@@ -17,213 +17,943 @@
...
+        )
+        self.sig_sum2 = DiscreteSignalSum.from_SignalSum(
+            self.signal2 - self.signal3, dt=0.5, start_time=0, n_samples=10
+        )
+
+        self.signal1 = DiscreteSignal.from_Signal(self.signal1, dt=0.5, start_time=0, n_samples=10)
+        self.signal2 = DiscreteSignal.from_Signal(self.signal2, dt=0.5, start_time=0, n_samples=10)
+        self.signal3 = DiscreteSignal.from_Signal(self.signal3, dt=0.5, start_time=0, n_samples=10)
+
+        self.double_sig_sum = self.sig_sum1 - self.sig_sum2
```
eggerdj, line 722: Is this used anywhere?

qiskit_ode/signals/signals.py
eggerdj: Ok. We can leave as is for now. It is easy to change later on and will not have a big impact.

qiskit_ode/signals/signals.py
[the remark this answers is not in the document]
eggerdj: Thanks.

qiskit_ode/signals/signals.py
```diff
@@ -11,216 +11,262 @@
...
-    Constant signal that has no carrier and may appear in a model.
-    """
+        return Signal(conj_env, -self.carrier_freq, -self.phase)
+
+    def draw(
+        self,
+        t0: float,
+        tf: float,
+        n: int,
+        function: Optional[str] = "signal",
```
eggerdj, line 209: This is a bit of a nit-pick but I'm not convinced by the name `function` here. I don't have a good alternative to offer either. Perhaps, `element` or `component`?

qiskit_ode/signals/signals.py
```diff
@@ -11,216 +11,262 @@
...
-        return self._value
-
-    def value(self, t: float = 0.0) -> float:
-        return np.real(self._value)
 
-    def conjugate(self):
-        return Constant(self._value.conjugate())
+        plotter = plt
+        if axis:
+            plotter = axis
```
eggerdj, line 231:
```suggestion
        plotter = axis if axis else plt
```

qiskit_ode/signals/signals.py
```diff
@@ -231,24 +277,75 @@ def __init__(
...
 
-        if samples is not None:
-            self._samples = Array(samples)
-        else:
-            self._samples = Array([0.0] * duration)
+        # initialize internally stored carrier/phase information
+        self._carrier_freq = None
+        self._phase = None
+        self._carrier_arg = None
+        self._phase_arg = None
```
eggerdj, line 294: Could this not be done in the base class? It feels strange to not call `super().__init__` here but I see why you did it since `Signal.__init__` wants to define a `self._envelope` which `DiscreteSignal` does not have. Can we not bring these classes closer together? E.g. by renaming `self._samples` to `self._envelope` in `DiscreteSignal` and letting `Signal.__init__` accept an `Array` as well as `Callable, complex, float, int` (you already do a conversion in the `__init__` of signal)

qiskit_ode/signals/signals.py
```diff
@@ -231,24 +277,75 @@ def __init__(
...
+        freq = signal.carrier_freq
+
+        if sample_carrier:
+            freq = 0.0
+            samples = signal(times)
+        else:
+            samples = signal.envelope(times)
+
+        return DiscreteSignal(
+            dt, samples, start_time=start_time, carrier_freq=freq, phase=signal.phase
```
eggerdj, line 347: Why no name?

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +433,696 @@ def add_samples(self, start_sample: int, samples: List):
...
+                sig = SignalSum(*sig)
+
+            if isinstance(sig, SignalSum):
+                components += sig.components
+            elif isinstance(sig, Signal):
+                components.append(sig)
+            elif isinstance(sig, (int, float, complex)) or (
+                isinstance(sig, Array) and sig.ndim == 0
+            ):
+                components.append(Signal(sig))
```
eggerdj, line 542: Seems like `Signal` should accept `Array` as input too? See my comment above.

qiskit_ode/signals/signals.py
```diff
@@ -11,216 +11,262 @@
...
+        Initializes a signal given by an envelope and a carrier.
 
         Args:
-            dt: Time increment to use.
-            n_samples: number of steps to resample with.
-            start_time: start time from which to resample.
-
-        Returns:
-            A piecewiseConstant signal.
+            envelope: Envelope function of the signal, must be vectorized.
```
eggerdj, line 80: From the docstring I would expect init to raise an error if the envelop is not vectorized but I cannot find one. Why?

qiskit_ode/signals/signals.py
```diff
@@ -11,216 +11,262 @@
...
+    .. code-block:: python
 
-    def to_pwc(self, dt: float, n_samples: int, start_time: float = 0.0) -> "PiecewiseConstant":
+        vectorized_func = np.vectorize(non_vectorized_func)
+        signal = Signal(envelope=vectorized_func, carrier_freq=5.)
+    """
+
+    def __init__(
+        self,
+        envelope: Union[Callable, complex, float, int],
```
eggerdj, line 71: Can this not also take an `Array` as input? See my comment below on `DiscreteSignal`. The if-else structure in this init also suggests the we should accept Array.

eggerdj (changes requested):
Almost there. I think we might be able to better align the internals of `Signal` and `DiscreteSignal`

test/ode/signals/test_signals.py
DanPuzzuoli: Yeah - it inherits from `TestSignalSum`, so all of the test cases from that class are being run on these signals.

qiskit_ode/signals/signals.py
DanPuzzuoli: I don't believe it's possible to check for this at instantiation. I think the only way to really find out that something _isn't_ vectorized is if it fails to produce arrays of the correct shape on some input. I'm not sure what we could do here aside from adding checks in every signal evaluation function that the input shape = output shape, but this seems like unnecessary overhead. I'm tempted to not worry about this for now: if something isn't vectorized I believe it should result in pretty clear errors (of course, I could be wrong).

qiskit_ode/signals/signals.py
DanPuzzuoli: Not sure what name to assign to this:

- If the original signal has no name then I think it's fine to not assign a name.
- If the original signal does have a name, if we want to assign a name we could do something like `'Discretized(dt=0.1, name)'`? Or do you think we should just set the name to the name of `signal`?

qiskit_ode/signals/signals.py
DanPuzzuoli: I agree in general with this comment, though I've tried a different approach than your suggestion: 

- I've instead changed `DiscreteSignal.__init__` to construct the envelope evaluation function (basically just the code that was in `DiscreteSignal.envelope`), then call `Signal.__init__` with this envelope. 
- Following this, I changed `DiscreteSignalSum.__init__` to call `DiscreteSignal.__init__`, as the envelope evaluation functions are the same in both of these cases.
- I also changed `SignalSum.__init__` to construct the envelope function (again, from code that was originally in `SignalSum.envelope`), then again called `Signal.__init__` with this envelope (as in `DiscreteSignal`).
- As a result of the above changes I was able to eliminate:
   - Several redundant lines of code in various `__init__` functions   
   - Overriding `envelope` methods in `DiscreteSignal` and `SignalSum`.
   - Unrelated to inheritance changes, but I realized the `complex_value` method in `DiscreteSignalSum` was redundant, as it inherits the correct form of this method from `SignalSum`, so this was also removed.

I think this might clean things up now - everything calls back to the original `Signal.__init__`, with array evaluation logic set up in the respective inits.

qiskit_ode/signals/signals.py
DanPuzzuoli: Yeah I'm not crazy about the name `function` either, but I've also been unable to come up with a better alternative. I don't like `element` or `component` as I think there is a bit of naming collision with `SignalSum`/`DiscreteSignalSum`, which actually have elements/components. 

The only other alternative I can think of that seems okay is `method`. 

- Pro: this arg is literally choosing which method (in the class sense) that is being plotted 
- Con: "method" sounds like it's referring to a "plotting method"

qiskit_ode/signals/signals.py
DanPuzzuoli: See response to above comment.

qiskit_ode/signals/signals.py
DanPuzzuoli: See response to above comment.

qiskit_ode/signals/signals.py
eggerdj: I think we can just set the name to the name of the signal.

eggerdj (changes requested):
A few more small comments.

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +427,678 @@ def add_samples(self, start_sample: int, samples: List):
...
-        envelope: Callable,
-        carrier_freqs: Array,
-        phases: Array,
-        drift_array: Optional[Array] = None,
+        dt: float,
+        samples: Union[List, Array],
+        start_time: float = 0.0,
+        carrier_freq: Union[List, np.array, Array] = None,
+        phase: Union[List, np.array, Array] = None,
+        name: str = None,
```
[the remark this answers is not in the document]
DanPuzzuoli: good catch, added it to the call to `DiscreteSignal.__init__`

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +427,678 @@ def add_samples(self, start_sample: int, samples: List):
...
-    Returns:
-        signal: The type will depend on the given base class.
+        phases = []
+        for sig in self.components:
+            phases.append(sig.phase)
 
-    Raises:
-        Exception: if signals cannot be added.
-    """
+        Signal.__init__(self, envelope, carrier_freqs, phases)
```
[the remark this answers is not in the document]
DanPuzzuoli: I assume this is solved with the type hints in your other suggestions

qiskit_ode/signals/signals.py
```diff
@@ -327,202 +427,678 @@ def add_samples(self, start_sample: int, samples: List):
...
-        self.phases = phases
 
-        self._im_angular_freqs = 1j * 2 * np.pi * carrier_freqs
+        if carrier_freq is None:
+            carrier_freq = np.zeros(samples.shape[-1], dtype=float)
+
+        if phase is None:
+            phase = np.zeros(samples.shape[-1], dtype=float)
+
+        DiscreteSignal.__init__(self, dt, samples, start_time, carrier_freq, phase)
```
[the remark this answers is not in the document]
DanPuzzuoli: I assume this is also solved by the type hints in your comments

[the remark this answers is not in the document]
DanPuzzuoli: Not sure if I messed up committing one of your suggestions, but I just manually added the updated type hints/doc description for `DiscreteSignal` - which should resolve this.

eggerdj (approved):

PR #8, opened 2021-06-08
Update GitHub linguist's calculation to better reflect the source language of the package.

paniash:
### Summary

It is clear that the source of the package is written in python as `.py` files. However, the GitHub linguist reports that close to 75% of the repository consists of Jupyter notebooks, which does a poor job at denoting the source language.

I suggest adding a `.gitattributes` file to better reflect the source language of the package. 

### Details and comments

This is just me being pedantic but would be a nice addition in my opinion.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ipynb linguist-vendored


DanPuzzuoli (approved):
Good idea! Thanks for this.

PR #12, opened 2021-07-22
[WIP] Simulation building

haggaila:

### DIFF
--- a/examples/qubit_chain.py
+++ b/examples/qubit_chain.py
@@ -1,0 +1,80 @@
+from qiskit_dynamics.builders.dynamical_simulation import *
+from qiskit_dynamics.builders.mpo_simulation_builder import *
+from qiskit_dynamics.signals import Signal
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+import matplotlib.pyplot as plt
+
+
+N = 4
+Jz = 1.
+g1 = 0.1
+t_f = 2.2
+step = 0.05
+t_span = [0, t_f]
+t_eval = np.arange(0., t_f, step)
+
+r_qubits = range(N)
+subsystems = OrderedDict()
+H = .5 * Sz(0)
+rho_0 = .5 *(Sid(0) + Sz(0))
+noise_signals = [Signal(g1)] * N
+noise_operators = []; obs_1q = []; obs_2q = []
+for i_qubit in r_qubits:
+	subsystems[i_qubit] = 2
+	obs_1q.append(Sz(i_qubit))
+	if i_qubit > 0:
+		H += .5 * Sz(i_qubit)
+		H += .5 * Jz * Sz(i_qubit - 1) * Sz(i_qubit)
+		rho_0 *= .5 *(Sid(i_qubit) - Sx(i_qubit))
+		obs_2q.append(Sz(i_qubit - 1) * Sz(i_qubit))
+	noise_operators.append(Sp(i_qubit))
+
+sim_def = SimulationDef(t_span = Array(t_span), t_eval = Array(t_eval), initial_state = rho_0,
+						hamiltonian_operators = [H], hamiltonian_signals = [Signal(1.)],
+						noise_operators = noise_operators, noise_signals = noise_signals,
+						observable_operators = obs_1q + obs_2q)
+sim_full = DenseSimulationBuilder(sim_def)
+sim_full.build(subsystems)
+sim_full.solve()
+
+tmp = 2
+
+# Exceptions below, WIP
+
+sim_prune = DenseSimulationBuilder(sim_def)
+subsystems_pruned = subsystems.copy()
+subsystems_pruned[2] = 0
+subsystems_pruned[3] = 0
+sim_prune.build(subsystems_pruned)
+sim_prune.solve()
+
+sim_mpo = MPOSimulationBuilder(sim_def)
+sim_mpo.build(subsystems[0:2])
+sim_mpo.solve()
+
+# plot comparative graphs
+
+
+# def plot_space_time(data: np.ndarray, t_steps: np.ndarray,  # t_ticks: np.ndarray, qubits: np.ndarray,
+# 					ax = None, fontsize = 16, b_save_figures = True, s_file_prefix = ''):
+# 	if ax is None:
+# 		_, ax = plt.subplots(figsize = (14, 9))
+# 	plt.rcParams.update({'font.size': fontsize})
+# 	im = ax.imshow(data, interpolation = 'none', aspect = 'auto')
+# 	divider = make_axes_locatable(ax)
+# 	cax = divider.append_axes('right', size = '5%', pad = 0.05)
+# 	plt.colorbar(im, cax = cax)
+# 	# plt.colorbar(im)
+# 	ax.set_xlabel('$t$', fontsize = fontsize)
+# 	ax.set_xticks(t_steps)
+# 	# ax.set_xticklabels(t_ticks, fontsize = fontsize)
+# 	# ax.set_yticks(qubits)
+# 	# ax.set_yticklabels(qubits, fontsize = fontsize)
+# 	ax.set_ylabel('qubits', fontsize = fontsize)
+# 	ax.set_title('$\\langle\sigma_z(t)\\rangle$')
+# 	if b_save_figures:
+# 		plt.savefig(s_file_prefix + '.sigma_z.png')
+#
+#
+# plot_space_time(sim_dense.obs_data[0:4, :], np.asarray(t_eval), b_save_figures = False)
+# plt.show()
--- a/qiskit_dynamics/builders/dynamical_operators.py
+++ b/qiskit_dynamics/builders/dynamical_operators.py
@@ -1,0 +1,283 @@
+from typing import OrderedDict, Dict, List, Union, Any
+from abc import ABC, abstractmethod
+import numpy as np
+from copy import deepcopy
+from qiskit.quantum_info.operators.base_operator import BaseOperator
+
+
+class DynamicalOperator(ABC):
+
+	DEFAULT_ALIASES =\
+	{
+		# Note that both strings must be lower case
+		'id': 'i',
+		'sx': 'x',
+		'sy': 'y',
+		'sz': 'z',
+	}
+
+	def __init__(self, s_id: Any = '', s_type = '', aliases = None):
+		self.s_id = s_id
+		self._s_type = ''
+		self._s_type_unique = ''
+		if aliases is None:
+			aliases = self.DEFAULT_ALIASES
+		self.aliases = aliases  # must be assigned before setting self.s_type
+		self.s_type = s_type
+		self.compound_type = ''
+		self.compound_ops = None
+
+	def __deepcopy__(self, memo = None):
+		cc = DynamicalOperator(self.s_id, self.s_type, self.aliases)
+		cc.compound_type = self.compound_type
+		cc.compound_ops = deepcopy(self.compound_ops, memo)
+		return cc
+
+	def new_operator(self):
+		return DynamicalOperator(aliases = self.aliases)
+
+	@property
+	def s_type(self) -> str:
+		return self._s_type
+
+	@property
+	def s_type_unique(self) -> str:
+		return self._s_type_unique
+
+	@s_type.setter
+	def s_type(self, s_type):
+		self._s_type = s_type
+		s_type = self.s_type.lower()  # should be here after assigning _s_type with user's string as is
+		self._s_type_unique = self.aliases.get(s_type, s_type)
+
+	def __add__(self, other):
+		if not isinstance(other, DynamicalOperator):
+			raise Exception("Both operands in an addition must be instances of a DynamicalOperator.")
+		result = self.new_operator()
+		result.compound_type = '+'
+		result.compound_ops = [self, other]
+		return result
+
+	# def __pow__(self, power, modulo=None):
+	# 	if type(power) is not int:
+	# 		raise Exception("Only integer powers are currently supported for operators.")
+	# 	result = DynamicalOperator(aliases = self.aliases)
+	# 	result.compound_type = '**'
+	# 	result.compound_ops = [self, power]
+	# 	return result
+
+	def __sub__(self, other):
+		return self.__add__(-other)
+
+	def __mul__(self, other):
+		result = self.new_operator()
+		if isinstance(other, DynamicalOperator):
+			result.compound_type = '@'
+			result.compound_ops = [self, other]
+			return result
+		else:
+			other_type = type(other)
+			if other_type is complex or other_type is float or other_type is int:
+				result.compound_type = '*'
+				result.compound_ops = [self, other]
+				return result
+		raise Exception("The second operand of a multiplication must be a DynamicalOperator class or a scalar.")
+
+	def __rmul__(self, other):
+		result = self.__mul__(other)
+		return result
+
+	def __neg__(self):
+		result = self.__rmul__(-1.)
+		return result
+
+	def __pos__(self):
+		return self
+
+	def get_operator_matrix(self, s_type_unique, dim: int) -> np.ndarray:
+		# TODO: replace qubit creation with Qiskit operator.from_label()
+		# TODO: support 'i x y z sp sm n p q a a_ n n^2 0 1 p_0_1'
+		if dim == 2:
+			if s_type_unique == 'i':
+				return np.identity(dim)
+			elif s_type_unique == 'x':
+				return np.asarray([[0, 1], [1, 0]], complex)
+			elif s_type_unique == 'y':
+				return np.asarray([[0, -1j], [1j, 0]], complex)
+			elif s_type_unique == 'z':
+				return np.asarray([[1, 0], [0, -1]], complex)
+			elif s_type_unique == 'sp':
+				return np.asarray([[0, 1], [0, 0]], complex)
+			elif s_type_unique == 'sm':
+				return np.asarray([[0, 0], [1, 0]], complex)
+			elif s_type_unique == '0':
+				return np.asarray([[1, 0], [0, 0]], complex)
+			elif s_type_unique == '1':
+				return np.asarray([[0, 0], [0, 1]], complex)
+		# Just a partial implementation for now
+		raise Exception(f"Operator type {self.s_type} unknown or unsupported for matrix generation with dimension {dim}.")
+
+
+class DynamicalOperatorKey:
+
+	def __init__(self, op: DynamicalOperator):
+		self.s_id = op.s_id
+		self.s_type_unique = op.s_type_unique
+
+	def __hash__(self):
+		return hash((self.s_id, self.s_type_unique))
+
+
+class OperatorBuilder(ABC):
+
+	def __init__(self):
+		self._ids = None
+		self._identity_matrices = None
+		self._subsystem_dims = None
+		self._total_dim = 0
+		self._dyn_op = None
+
+	def build_dictionaries(self, operators: Union[DynamicalOperator, List[DynamicalOperator]]) -> List[Dict]:
+		results = []
+		b_flatten = False
+		if type(operators) != list:
+			b_flatten = True
+			operators = [operators]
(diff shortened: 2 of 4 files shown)


DanPuzzuoli:
A record of comments based on our discussion:

For `DynamicalOperator`, I think where the design should go should be motivated by the following (we can add to the list as necessary):
- How do users register new `DynamicalOperator` subclasses?
    - It would be nice if all they had to do is create a new subclass and override the behaviour of the `get_operator_matrix` method, or could even instantiate `DynamicalOperator` directly with a concrete operator (without even needing to ever define a string alias)
- How should we handle operators defined on multiple subsystems?

General `DynamicalOperator` design comments: 
- The `DynamicalOperator` class is being used both to represent specific operators (e.g. `'x'`, `'y'`, `'z'`, or composite operators) acting on specific subsystems, as well as a "global" registry of string aliases for operators (in the static method `get_operator_matrix`). I think it might be good to clean up this behaviour:
    - Instead of having `get_operator_matrix` be a static method, each subclass should implement this directly (i.e. what the `Sz` operator should be concretely only needs to be contained in the `Sz` class. I think the general registry of behaviour currently existing in `get_operator_matrix` could be moved into a `DynamicalOperator.from_label` static method, that just knows how to reroute to the correct subclass. This will also allow us to retain some similarity with `Operator.from_label`.
- I think the functionality in the classes `DynamicalOperatorKey` and `OperatorBuilder` should be moved into `DynamicalOperator`
    - `DynamicalOperatorKey` essentially enables comparison between `DynamicalOperator` instances, so maybe we can just compare them directly, e.g. by implementing the `__eq__` method in `DynamicalOperator`? It can compare class and subsystem data, or for `DynamicalOperator`s defined purely in terms of a concrete matrix, can check equality of the matrices. I think this will require some thought but it can potentially completely remove the use/dependence on string aliases.
    - The functionality in `OperatorBuilder` I think can just be directly moved into methods of `DynamicalOperator` as is (it'll lose the list functionality but I think that's fine).

haggaila:
To record our recent discussion:
``DynamicalOperator`` will keep a unique string for the type (necessary for the dictionary building and to make connection with non-matrix-based solvers such as the MPO solver. However, each operator type will be implemented by a subclass and override ``get_operator_matrix``, which will make the aliases unnecessary (will be removed). Also, an operator can be instantiated with a specific matrix and be given an id automatically if necessary. ``DynamicalOperatorKey`` will be retained for dictionary keys (to not have dictionaries of object references), and ``_build_dictionary`` will automatically generate a dict pointing to the ``DynamicalOperator`` instances according to keys. The key will be extended such that multiple subsystem ids can be stored. Reordering subsystems will be considered (using the ``qinfo`` implementation). The functionality in ``OperatorBuilder`` will remain there as there are precomputation steps and member fields used internally for building matrices.

haggaila:
@DanPuzzuoli I pushed an implementation of the above discussed revision. It's ready for review.

DanPuzzuoli:
For now I just have comments on `DynamicalOperator` and it's related functionality.

This iteration is a definite improvement, though I think this still needs more interface/usage clarity.

Design comments:
- The usage of the `s_type` attribute is still a bit strange to me. I like that the concrete matrix generation has been moved to the `DynamicalOperator` subclasses, but it feels strange that it checks the `s_type` attribute. E.g. I can instantiate an `Sx` instance, but then change `s_type` to something other than `'x'`, which yields the instance totally useless. This really seems like something that shouldn't be changeable _in principle_.
     - I know `s_type` is being used elsewhere in the capacity of a unique identifier, so it has a purpose, but it feels like something that shouldn't really be part of the user experience or have the ability to actually modify the behaviour of instances.
     - I don't have a clear suggestion for this problem, but this is a fundamental aspect of the class and how it behaves so it is important.
- The handling of singleton v.s. composite operators seems unclear. The only way to create a composite operator instance directly is to instantiate an empty `DynamicalOperator`, then manually set the relevant attributes. After that, the class has no functionality for working with composite operators (the only thing being modified is the behaviour of `OperatorBuilder`). This somewhat uncouples the data storage from the functionality of composite operators, which I think makes their behaviour unclear.
    - Here I'd like to see some kind of class structure for representing composite operators. Some version of this is in the `Signal` class/subclass, but e.g. there is a `SignalSum` class which represents a sum of signals. It exposes the same interface as `Signal` itself, but is explicit that it is representing a composite object, rather than a single object, and methods can be overridden as necessary to handle this. This is basically the problem described/solved by the [composite design pattern](https://en.wikipedia.org/wiki/Composite_pattern): we want to have "singleton" operators, as well as composite operators built from expressions of singletons, but we want them to both to expose the same interface when it comes to algebraic operations and concretization methods. 
    - What I would like to see here is a class (or classes) which explicitly represents composite operators (e.g. a `DynamicalOperatorSum` and/or `DynamicalOperatorProduct`) which can be instantiated directly with lists of operators. These composite classes can, e.g., override `get_operator_matrix` to call the operator builder functionality.
- The above point is connected to: there isn't a need for the `OperatorBuilder` class.
    - We discussed this before but the functionality is static - the methods do modify internal attributes, but they're only used during that single function call, so the usage of instance attributes (in my opinion) only serves to obfuscate where the information is going and how it is being used.
    - I still think the dictionary building and matrix building functionality should be exposed as methods of `DynamicalOperator` (e.g. so that composite operators can be built). E.g. for a `DynamicalOperatorSum`, we can override `get_operator_matrix` to return the operator. 
    - For the above point I am suggesting implementing class methods that build a _single_ composite operator, and I know there are efficiency concerns with calling this as a method on each individual operator v.s. calling a function to do a list of operators all at once, but I think a solution that would enable both modes of operation, and solves my issue in the first point is:
         - Move the functionality in `OperatorBuilder` into standalone utility functions for building the dictionaries and for building the matrices. This can still function on lists, and be used internally where necessary. We can however override `get_operator_matrix` for composite operators to just call the utility function. This way a user can get the matrix for a specific operator without having to import/call the utility functions.
         - I know this isn't too big a deal but it just really feels like every dynamical operator (whether singleton or composite) should "know" how to turn itself into a matrix (with the relevant subsystem data of course).

Some more concrete comments:
- `DynamicalOperator.is_scalar` should be a utility function not associated with any class or instance.
     - There is a file in the root folder of the package called `type_utils.py` that contains functions like this, so may be worth putting it there for general usage.
- `DynamicalOperator.new_operator` just does `return DynamicalOperator()`, so any code that calls this should just directly instantiate `DynamicalOperator`.
- I think `Null(DynamicalOperator)` should be renamed to `Zero(DynamicalOperator)`, or something like this. I don't think `null` is used in python but it suggests something like `None`.
- The same exception is being raised in every subclass of `DynamicalOperator`. I wonder if we can make `DynamicalOperator.get_operator_matrix` call `DynamicalOperator._get_operator_matrix`, then raise the exception if this doesn't do anything. Subclasses can then implement `DynamicalOperator._get_operator_matrix` and don't need to individually raise the error every time.

haggaila:
For the first list above:

1. I agree, s_type should not be modifiable, it is protected by a property for that purpose and the existence of the setter is a left-over from the earlier version, that I will now remove.
2. Can be done of course, but will make the code more cumbersome with more class types and more validations, and I I don't right now see that it's required - am I missing? Do you see a use case to directly instantiating a product composite class (rather than writing it with *)?
3. (i) I think it's a matter of taste.
   (ii) The method _build_one_dict() could move to class DynamicalOperator because it's a self-contained (recursive) function. The wrapper build_dictionaries() could be made static. However, the matrix building should not IMO be moved to DynamicalOperator because of the precomputation of 5 variables (stored as builder instance fields as you note), global cached information of the building, which get_operator_matrix() is ignorant of.
   (iii) So if you insist, I would add static wrapper functions, but I would not implement the building of matrices in class DynamicalOperator.

For the second list above:
Ok for all except (2), new_operator is required in order that subclasses of DynamicalOperator could return instances of the right subclass.

haggaila:
I am pushing a revision of the code that implements: dictionary building as part of ``DynamicalOperator``, discarding of ``OperatorBuilder`` in favor of functions, other changes discussed above, but not composite ``DynamicalOperator`` containers. If necessary, those can be added later without change of interface. Some further cosmetic changes and rearrangements also included, and further development of the simulation builders (that remain in a preliminary state).

DanPuzzuoli:
Summary of discussion we just had about the `SimulationDef` and builder interfaces:
- Parameter handling
    - The goal of this component is to define a common set of simulation tasks in a unified way, which can then be "compiled" to different simulation pathways (standard solvers, MPO solver, perturbation solvers) while trying to minimize user intervention. I.e. these different simulation pathways are composed of different "lower level" pieces that have similar but different setups/interfaces, and this will allow streamlining the process of setting things up and passing to those solvers (of course, for a more narrow definition of "simulation task").
    - Given this goal, we need to be intentional about what parameters go into the generic `SimulationDef`, and which parameters need to be passed as arguments to the specific builders.
    - E.g. Times are handled differently by different solvers:
         - `solve_lmde` allows for specification of arbitrary starting/endpoints, and the state can be returned at arbitrary intermediate time points.
         - perturbation solvers require a fixed `dt` to compile to begin with which can't be changed afterwards, and the simulation routine allows for an arbitrary initial time, but the full simulation interval must be specified as an `int` number of time steps, and intermediate evaluation points must also be `int`s
         - MPO I don't believe requires a commitment to a `dt` value at compile-time. (How is simulation time specified?)
    - As raised by @levbishop , however we handle this, we need to make sure that we are striking the right balance between combining common procedures but not over-merging interfaces. (And also make sure that it's worth doing this - though I think at this point we all roughly agree that it is.)
    - Options:
          - Currently, everything is passed to `SimulationDef`, even disparate arguments that apply to some solvers and not others, and @haggaila prefers this.
          - My preference is that things that are not common amongst simulation pathways should be passed directly to the point at which they are used - i.e. additional parameters that must be committed to when compiling a certain method must be supplied when `build` is called, and solver-specific parameters like time-parameters of varying formats should be passed to the `solve` function directly.
          - We can continue discussing, just recording here.
- Things to look out for in the design of these components:
    - Simulating state vector v.s. density matrix v.s. unitary v.s. lindbladian propagator
    - Integration with quantum info types
        - Make sure to take advantage of specialized methods, e.g. pauli observables
    - Clear organization/exposure of what _can_ or _can't_ be set after compiling to a given simulation pathway

haggaila:
@DanPuzzuoli , the MPO solver does not really have a truly distinct compile-time step - you invoke it with a list of parameters (including ``dt``), and it evolves the solution until the final time (saving intermediate results and observables to disk as requested).
About ``SimulationDef``, just to stress that I think its design should be guided by being able to describe a set of problems (here, Hamiltonian/Lindblad evolution with the scope that we discussed), and not by solver details of execution or supported parameters. In other words, if the same problem can be solved by a fixed-time-step solver that digests a ``dt`` parameter and ``n_steps``, and by a variable-time-step [not sure about all this hyphenation] solver that takes in a bit of different parameters, both should use one ``SimulationDef`` class, that'll encapsulate the two options, with relevant validations.
I'll try to generate a working useful implementation next week for a concrete review.

DanPuzzuoli:
> About `SimulationDef`, just to stress that I think its design should be guided by being common to a defining set of problems (here, Hamiltonian/Lindblad evolution with the scope that we discussed), and not by solver details of execution or supported parameters. In other words, if the same problem can be solved by a fixed-time-step solver that digests a `dt` parameter and `n_steps`, and by a variable-time-step [not sure about all this hyphenation] solver that takes in a bit of different parameters, both should use one `SimulationDef` class, that'll encapsulate the two options, with relevant validations. As an "ad absurdum" counterexample (IMO at least), since there may be solvers that support both fixed and variable timesteps, we don't want to write two `SimulationDef` (and related builder) classes for each.

I agree that we should not have multiple `SimulationDef`s - I don't think anyone is suggesting that. The question though is what is our strategy for handling differences between solvers: 
- The option you are proposing is to have `SimulationDef` store every possible argument/configuration option that one might want to pass to a solver. 
    - The benefit you're advocating is: the user can do this once and then never have to pass arguments to `.solve()` again no matter how many builders they use.
    - The drawbacks I see are: 
        - This makes the meaning of `SimulationDef` less clear. It's defining a general simulation problem, but it's also a data container that needs to store every possible arg/configuration parameter for solvers. So in some sense it's supposed to be general and solver independent, but if we try to include all possible solver configuration arguments, it's going to need to know a lot about the specifics of the solvers. E.g. not counting `t0` and `n_steps`, the perturbative solvers actually have 7 highly specialized configuration arguments, not just `dt`; do I need to add every one of them to `SimulationDef`? What if two solvers require configuration parameters that have similar sounding names but are actually different - will we need to specialize the argument names, e.g. `dt_solver1` and `dt_solver2`? We'll need to ask ourselves these questions now for the anticipated solvers, as well as later whenever we want to add a new one we are not anticipating.
        - Another issue is: what if a user specifies different types of arguments for different solvers that are actually incompatible, e.g. if `tf != t0 + n_steps * dt`. You might say this is fine: it's up to the user to be consistent and they can be inconsistent if they want. On some level I would agree with that, but on principle I don't want to enable such inconsistencies.
- The option I am proposing is: 
    - Remove from `SimulationDef` arguments that are not truly standardized across the problem and across solvers. I understand that time parameters feel like they should go here, but there is so much variation amongst how solvers handle them, even with just 3 solvers (MPO, solve_lmde, and perturbative), that it doesn't seem feasible.
    - When calling a builder, parameters that are necessary for compilation must be supplied at that moment.
    - `.solve` will need to be called with the arguments that the underlying solver needs, e.g. time values to `solve_lmde`, `t_eval`, everything.
    - The drawback of this approach is that it doesn't have the benefit of yours: the user will now need to supply these arguments to different places. 
    - Benefits: Much clearer where arguments are going and how they're being used. `SimulationDef` is well defined, and only the only place that solver-specific details need to go is in the builder associated with that solver. I.e. we avoid the mixed validation and maintainability issues as described above.

I don't know if you think that this will reduce the usefulness of the `SimulationDef` and builders, but I don't think it does at all. `SimulationDef` is still defining a general problem, and the builders are still organizing/collecting functionality across the package into a semi-standardized interface (managing `DynamicalOperator` -> different representations, storing/organizing compiled data, and knowing where all the relevant functionality in the package is).

haggaila:
@DanPuzzuoli, I will implement your suggestion. To be sure I understand - do you mean that ``solve()`` will take the time parameters among its ``**kwargs``? Or that there will not be a ``SimulationBuilder`` base class (or that ``solve()`` will not be an abstract function of it? Could you specify which parameters to remove from ``SimulationDef`` and should they go to ``solve`` or ``build``?

DanPuzzuoli:
> @DanPuzzuoli, I will implement your suggestion. To be sure I understand - do you mean that `solve()` will take the time parameters among its `**kwargs`? Or that there will not be a `SimulationBuilder` base class (or that `solve()` will not be an abstract function of it? Could you specify which parameters to remove from `SimulationDef` and should they go to `solve` or `build`?

The `SimulationBuilder` base class should define a minimal interface _in association_ with what the `SimulationDef` class is intending to represent. I.e.:
- Define a model
- Simulate over an interval
- Evaluate state at intermediate time points
- Evaluate observables at intermediate time points
Hence, the `SimulationBuilder` base class needs to define:
- A constructor, which at the very least takes in a `SimulationDef` (I guess we are calling this `build`). Subclasses can have additional arguments for things that must be committed to when building.
- Setter methods for adjustable parameters. (I guess the base class won't necessarily have or need this.)
- A `.solve` method, which must take in, at the very least, arguments associated with the task defined in `SimulationDef` - integration time parameters, evaluation times, observable evaluation times. We allow each subclass to make these things more concrete (e.g. is `t_eval` a `float` or an `int`, etc).

For something like this it's possible that the abstract base class and typing is going to need to be super loose - e.g. for `solve` we want it to accept the time parameters defining the problem, but we may want to give them different names from solver to solver (instead of `t_span` the perturbative solver will need `t0` and `n_steps`). The base class can at least document the intended use.

In terms of specifically where the parameters go, I'm thinking:
- All time arguments can be removed from `SimulationDef`.
- When calling `build` (or constructing the builder, whatever the right term is), any parameters that need to be committed to at compilation time need to be specified.
    - I guess for MPO it sounds like there isn't anything that needs to be committed to.
    - For `solve_lmde`-based solvers maybe we can have some optional arguments here setting rotating wave cutoff frequency, and generally some optional arguments for what kind of evaluation mode to operate in (e.g. dense v.s. sparse v.s. vectorized). These are basically things that go into setting up the model; they're not super expensive or anything but they seem in line with the kind of thing that the builders are setting up.
    - For perturbative solvers this will require a large list of parameters that I won't lay out here, but among them is the `dt` of the solver.
- `.solve` I think, as described above, should accept time arguments as described above, along with whatever associated optional arugments there are for the specific solver
    - E.g. for `solve_lmde`, we can have `.solve(t_span, t_eval, method, atol, rtol, ...)` (basically the one argument that doesn't need to be supplied from normal usage of `solve_lmde` is the model). This can then do additional processing on the results to output the required simulation results as defined by `SimulationDef`, like compute expectation values.

DanPuzzuoli:
I like the inclusion of time variables being passed directly to `solve`, but I feel like the `SimulationTimes` object is reproducing the issues I had with including the times in the `SimulationDef` initially. Having a class which stores all time variables forces consolidating argument names and meanings and makes it unclear which time variables the specific solver is using. It also requires maintaining the class and its internal logic and validation procedures. Making the time variables explicit arguments to `.solve` allows each builder to directly document what it needs, and to define things how it wants (e.g. setting `t_eval` to ints v.s. floats). I don't see why saving on the number of arguments to `solve` is worth these trade offs. 

With regards to `DenseSimulation` and `DenseSimulationBuilder` - why are these separate classes? Can `DenseSimulation` not just accept `SimulationDef` directly and do the build step at construction and store things internally?

In terms of naming conventions, I don't think that `DenseSimulation`/`DenseSimulationBuilder` need to specifically reference "dense". The handling of sparse v.s. dense v.s. vectorized is all controlled by a single argument to the model classes, so this can just be supplied as an argument to the build function.

haggaila:
@DanPuzzuoli I don't think that grouping these common time variables to one class creates a problem, I think that it gives a unified interface for slightly different but very related variables and behaviors of solvers. Since there are multiple solvers that are either fixed-step or variable-step ones, it allows code reuse realizing one convenient interface ("define once use everywhere") and there is no complicated logic to document - in any case a futuristic class can implement a solve method that takes other arguments.
Take the final decision on it.
``DenseSimulationBuilder`` is separated because of the discussed one-to-many relation: Like that a user can build the operators once (a possibly costly step) and then run different simulations based on it (with different parameters etc.), with different storage for each one. As discussed, there should be a fast interface for a "define-build-simulate". I'll push some code later.
Do you want the prefix "Dense" replaced with "Matrix", "Dynamical", something else or just removed?

haggaila:
I think that the simulation building classes have somewhat lost their meaning, as they currently don't serve well the purpose of a unified interface for a common problem, the parameters and options are spread over a few data structures (``SimulationDef``, ``SimulationTimes`` that is actually targeted to be discarded), and various function parameters [class constructors, ``build()``, ``solve()``], and also there many parameters are being duplicated [e.g. the list of ``MatrixBuilder.solve()`` parameters that are just delegated to ``solve_lmde()``, and many others].
I am therefore closing this PR. I will open a separate PR just for the dynamical operators as you suggested @DanPuzzuoli . In addition, let's discuss the builders workflow.

I suggest the builder classes to support the following workflow:
1. Define Hamiltonian/Lindblad operators.
2. Do (potentially costly) precomputations (build operators), targeting different solvers and/or varying build parameters while using a single definition of the model operators from step (a).
3. Specify an initial state, simulation times, solver parameters, and solve, possibly multiple times (varying input), using one set of prebuilt operators from step (b).
4. Potentially serialize a solution from step (c) to files.
5. Potentially load a serialized solution from files.
6. Specify and calculate observables using a solution.
7. Extract observables from multiple (different) simulations, do simple manipulations on the data, and plot.

Builders should also balance code reuse of similar tasks between different solvers. Specifically, try to encompass the important similarities when solving Hamiltonian/Lindblad models (despite differences in details between solvers).

haggaila:
Implementation deprecated.

PR #13, opened 2021-07-23
Groundwork for operator_collections

averyparr:
### Summary
1. `OperatorCollection`s provide a modular way for `Model` classes to use and easily switch between different evaluation modes. 
    
    a. Easier to implement new evaluation modes without cluttering every `Model` class
    
    b. Allows `Models` to work in four steps, each of which is modular and separated from the main `Model` implementation. 

2. Rotating Wave Approximation is now handled by a function mapping `Model` classes to other `Model` classes

    a. Clearer interface on the user side with easy signal handling for translating pre-RWA to post-RWA. 

    b. Core functionality is implmented in helper functions, which makes future RWA implementations easier

    c. Most of the `perform_rotating_wave_approximation` function is simply casework to determine which type of model it is 
        dealing with, along with a smal bit of special handling for each `Model` class. This makes adding new `Model` classes
        relatively straightforward. 

3. Additional features added over the past few weeks

    a. Non-vectorized `LindbladModel` evaluation and `RotatingFrame` support for `LindbladModel`s. 

    b. Basic sparse matrix operator storage and calculation (`csr_matrix`)

    c. Rotating frames & basis transformations for vectorized evaluation are implemented in a manner that's simpler to the user 
        (e.g. passing only $F$, and the `Frame` object figuring out how to deal with vectorized operators).

    d. Diagonal frame operators no longer waste time changing bases. 

    e. Optimizations to speed in a few existing places, as well as several rounds of optimization within `OperatorCollection`s.

    f. `evaluate_hamiltonian` functionality for `LindbladModel`


qiskit_dynamics/models/operator_collections.py
```diff
@@ -0,0 +1,291 @@
...
+        for the hamiltonian terms. 
+        Args: 
+            signal_values: length-2 list of Arrays. 
+            components: 
+                signal_values[0]: hamiltonian signal values, s_j(t)
+                    Must have length self._num_ham_terms
+                signal_values[1]: dissipator signal values, \gamma_j(t)
+                    Must have length self._num_dis_terms
+            y: density matrix [(n,n) array] representing the state at time t
+        """
```
DanPuzzuoli, line 282: I think here it may be worth combining the left mult of the Hamiltonian with left mult of the dissipator commutator, and similarly for the right mult parts. This will turn two matrix mults into two matrix additions. Somewhat minor but it should be strictly more efficient.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": true
```
DanPuzzuoli, line 2: Did you add this file? I don't think this should be added

DanPuzzuoli (commented):
Looks cool!

I have a couple of general comments:
- I really like conceptually what you've done with `DenseGeneralOperatorCollection`, however I'm concerned that the complexity introduced by its generality (i.e. checking different input types and setting flags for behaviour, then conditionals check these flags when evaluating) is not necessarily balanced with the expected uses of that generality. I.e., we really need:
     - A left multiplication-only version, and
     - One for representing Lindblad dissipators.
     - In the future, we may need one for the dissipation portion of the Bloch-Redfield equation, but that is also just one more
 
Again to be clear, I like what you've done to facilitate that generality, and while in principle generality is good, there is a cost to code complexity in terms of readability (and hence future maintenance), and given how few anticipated special cases there are, my immediate feeling is that the generality is not worth the complexity. So my recommendation here is to axe this.

- If the above is axed, I'd rename `DenseSingleSideOperatorCollection` to `DenseOperatorCollection`, and get rid of the `LMult`/`RMult` options, and just have it do left multiplication

- For `DenseLindbladCollection`:
    - See inline comment for evaluating things
    - This generally looks good to me though

qiskit_dynamics/models/operator_collections.py
averyparr: Agreed. Will change implementation for next commit.

.vscode/settings.json
averyparr: It wasn't added intentionally. Will find a way to remove.

qiskit_dynamics/models/generator_models.py
```diff
@@ -444,4 +450,4 @@ def _evaluate_in_frame_basis_with_cutoffs(self, sig_vals: Array):
         return 0.5 * (
             np.tensordot(sig_vals, self._ops_in_fb_w_cutoff, axes=1)
             + np.tensordot(sig_vals.conj(), self._ops_in_fb_w_conj_cutoff, axes=1)
-        )
+        )
```
averyparr, line 453: Could be slightly faster (by one element-wise multiplication) if we just define the operators to be half as large when passing them to the operator collections.

qiskit_dynamics/models/generator_models.py
```diff
@@ -311,137 +276,94 @@ def frame(self) -> Frame:
...
         Raises:
-            QiskitError: If model cannot be evaluated.
-        """
+            QiskitError: If model cannot be evaluated."""
 
         if self._signals is None:
             raise QiskitError("""GeneratorModel cannot be evaluated without signals.""")
 
-        sig_vals = self._signals.complex_value(time)
+        sig_vals = np.real(self._signals.complex_value(time))
```
DanPuzzuoli, line 311: For this you can directly call `self._signals(time)` (as it will evaluate the real part).

qiskit_dynamics/models/generator_models.py
```diff
@@ -65,70 +61,52 @@ def frame(self) -> BaseFrame:
...
+        Takes care of putting all operators into the basis in which the frame
+        matrix F is diagonal.
+        """
         pass
 
-    @cutoff_freq.setter
     @abstractmethod
-    def cutoff_freq(self, cutoff_freq: float):
-        """Set cutoff frequency."""
+    def evaluate_with_state(
```
DanPuzzuoli, line 72: What are your thoughts on changing the names of these functions to:
- `evaluate_generator`, and
- `evaluate_rhs`

I'm just wary of the naming collision between the "state" of a differential equation (which is what this is referring to more generally) and a "quantum state". I think the above suggested names are very concrete/explicit (but am of course open to discussion).

qiskit_dynamics/models/generator_models.py
```diff
@@ -168,22 +146,17 @@ def frame(self, frame: Union[Operator, Array, Frame]):
...
-
-    @cutoff_freq.setter
-    def cutoff_freq(self, cutoff_freq: float):
-        """Cutoff frequency not supported for generic."""
-        if cutoff_freq is not None:
-            raise QiskitError("""Cutoff frequency is not supported for function-based generator.""")
-
     @property
     def drift(self) -> Array:
+        """Gets the drift term"""
```
DanPuzzuoli, line 151: Should this be being pulled from the operator collection?

qiskit_dynamics/models/generator_models.py
```diff
@@ -311,137 +276,94 @@ def frame(self) -> Frame:
...
-        """
-
-        # for now if the frame operator is not None raise an error
-        if self.frame.frame_operator is not None:
-            raise QiskitError(
-                """The drift is currently ill-defined if
-                               frame_operator is not None."""
-            )
+        # Apply rotations e^{-Ft}Ae^{Ft} in frame basis where F = D
+        if self.frame.frame_diag is not None:
```
DanPuzzuoli, line 317: this block should just be calling a method (or method(s) if it requires multiple calls) of `Frame` directly, rather than implement it within `GeneratorModel`. If the `Frame` object is doing it less efficiently than what you've figured out with the array optimizations, then feel free to change the implementation in `Frame`!

qiskit_dynamics/models/lindblad_models.py
```diff
@@ -127,6 +130,71 @@ def from_hamiltonian(
...
+            is returned in the frame or not."""
+
+        if not in_frame_basis:
+            y = self.frame.operator_into_frame_basis(y)
+
+        hamiltonian_sig_vals = np.real(self._hamiltonian_signals.complex_value(time))
+        dissipator_sig_vals = np.real(self._dissipator_signals.complex_value(time))
+
+        # Need to check that I have the differences chosen correctly
+        if self.frame.frame_diag is not None:
```
DanPuzzuoli, line 182: This should be implemented purely by making calls to methods of `self._frame` (see other comment related to this).

qiskit_dynamics/models/operator_collections.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def apply_function_to_operators(self, function_to_apply: Optional[Callable]):
+        """Applies a function to all operators in the collection.
+        Args:
+            function_to_apply: the function to be applied to all operators"""
+        self._operators = function_to_apply(self._operators)
+        if self.drift is not None:
+            self.drift = function_to_apply(self.drift)
+
+    def evaluate_without_state(self, signal_values: Array) -> Array:
```
DanPuzzuoli, line 137: Similar to the comment for `GeneratorModel`, I'd rather call this somethign like: `evaluate_linear_combination`, and the `evaluate_with_state` to something like `apply_map`. Just to be more concrete about what's being "evaluated", and to avoid naming collisions for meanings of the word "state".

qiskit_dynamics/models/operator_collections.py
```diff
@@ -0,0 +1,287 @@
...
+
+    @property
+    def num_operators(self):
+        return self._hamiltonian_operators.shape[-3], self._dissipator_operators.shape[-3]
+
+    @property
+    def hilbert_space_dimension(self) -> int:
+        return self._hamiltonian_operators.shape[-1]
+
+    def apply_function_to_operators(self, function_to_apply: Optional[Callable]) -> None:
```
DanPuzzuoli, line 198: nice function!

qiskit_dynamics/models/frame.py
```diff
@@ -625,6 +641,10 @@ def _conjugate_and_add(
         Note: B is added in the frame basis before any potential final change
         out of the frame basis.
         """
+        if self.vectorized_operators:
+            # faster than the flatten operation later.
```
DanPuzzuoli, line 645: What does this comment mean?

qiskit_dynamics/models/frame.py
```diff
@@ -672,6 +696,25 @@ def operators_into_frame_basis_with_cutoff(
...
         # create difference matrix for diagonal elements
         dim = len(ops_in_frame_basis[0])
+        cutoff_array = self.calculate_cutoff_filter(carrier_freqs, cutoff_freq, dim)
+
+        return (
+            cutoff_array * ops_in_frame_basis,
+            cutoff_array.transpose([0, 2, 1]) * ops_in_frame_basis,
+        )
+
+    def calculate_cutoff_filter(
```
DanPuzzuoli, line 706: I wonder if it makes sense to just have this be a utility function in the file that contains the rotating wave approximation, since this is very specific to that.

qiskit_dynamics/models/generator_models.py
```diff
@@ -50,85 +50,107 @@ class BaseGeneratorModel(ABC):
...
-          evaluation of :math:`G'(t)` to remove terms above a given frequency.
-          In this abstract class the exact meaning of this is left unspecified;
-          it is left to concrete implementations to define this.
     """
 
     @property
     @abstractmethod
-    def frame(self) -> BaseFrame:
-        """Get the frame."""
+    def hilbert_space_dimension(self) -> int:
```
DanPuzzuoli, line 57: I would rather just call this `dim`. While it's technically a Hilbert space, that terminology is somewhat overpowered in this context.

qiskit_dynamics/models/generator_models.py
```diff
@@ -231,37 +261,64 @@ class GeneratorModel(BaseGeneratorModel):
...
-        # initialize internal operator representation in the frame basis
-        self.__ops_in_fb_w_cutoff = None
-        self.__ops_in_fb_w_conj_cutoff = None
+    @property
+    def hilbert_space_dimension(self) -> int:
+        return self._operators.shape[-1]
+
+    @property
+    def evaluation_mode(self) -> str:
+        return super().evaluation_mode
```
DanPuzzuoli, line 315: Is this necessary if you're just straight up calling `super().evaluation_mode`? I think this should just be inherited.

qiskit_dynamics/models/generator_models.py
```diff
@@ -231,37 +261,64 @@ class GeneratorModel(BaseGeneratorModel):
...
+
+    @evaluation_mode.setter
+    def evaluation_mode(self, new_mode: str):
```
DanPuzzuoli, line 318: this is great!

DanPuzzuoli (commented):
Some other comments

For frame:

- Should the `vectorized_operators` property be a property of the class or just an optional argument to the evaluation functions?
    - If it’s only one function I would recommend just making it an optional argument to that function
    - How exactly does this work and what exactly does it mean?

GeneratorModel
- evaluate_mode
    - Looks good!
    - We should just shorten “dense_operator_collection” to “dense”
    - “dense_lindblad_collection” should also just be shortened to “dense”, and “dense_vectorized_lindblad_collection” to “dense_vectorized”

qiskit_dynamics/models/frame.py
averyparr: Ah–the reshape (n**2,) -> (n,n) is about three orders of magnitude faster than the reshape (n,n) -> (n**2,) in the tests I've done. I can remove the comments, if they're not clear.

qiskit_dynamics/models/frame.py
averyparr: I think this function and operators_into_frame_with_cutoff are now obsolesced by the new Model -> Model RWA functions. I had originally used this because I wanted to be able to access _just_ this functionality from outside the Frame object in pervious versions. As of now, I'm not particularly attached of having this be a separate function.

averyparr: I think this function and operators_into_frame_with_cutoff are now obsolesced by the new Model -> Model RWA functions. I had originally used this because I wanted to be able to access _just_ this functionality from outside the Frame object in pervious versions. As of now, I'm not particularly attached of having this be a separate function.

qiskit_dynamics/models/generator_models.py
averyparr: It really is not necessary–in fact, I'd like not to have it. It's only there because I haven't found a way to use the @evaluation_mode.setter decorator in this class to override the setter method without either 1) also overriding the getter method, with something like this, or 2) getting some linting "error" (not sure if it will actually error if I try to run it, or if I can just add a #pylint ignore or something)

PR #14, opened 2021-07-30
Removed constant amplitude function vs constant

averyparr:
### Summary
Removed discrepancy in Signal class between the following scenarios: 
1) User passes complex amplitude c to the Signal constructor
2) User passes function f(t) = c as the amplitude

Adjusted unit tests to reflect this.


DanPuzzuoli (approved):
Good catch and fix.

PR #16, opened 2021-08-10
Add tutorials to doc builds

mtreinish:
### Summary

This commit adds the tutorials to the documentation builds. Previously
they were in a self contained directory and weren't compiled with the
documentation or executed in CI. This commit fixes both of those
oversights so we include the documentation in sphinx builds and execute
them as part of that build process.

At the same time the CI configuration for publishing documentations to
https://qiskit.org/documenation/dynamics at release time is added.
Automatically when a tag for qiskit-dynamics is pushed it will trigger a
documentation build and upload that.


DanPuzzuoli (approved):
Passing tests and tutorials build. The rendering of tutorial thumbnails is off (locally) but we can deal with that when we review the tutorials themselves.

PR #18, opened 2021-08-20
Add scan_parameters for numpy.

averyparr:
### Summary
Adds a scan_parameters function that makes evaluation of a function at all combinations of several inputs simpler.

### Details and comments

scan_parameters(func,*args, method="parallel",**kwargs) evaluates func, which takes len(args) arguments, where each arg in args is an array type, at all possible combinations of these input arguments, using the method specified by method.


DanPuzzuoli:
Closing this as its old and we'll likely need to think more about how this gets done.

PR #27, opened 2021-09-16
Qiskit pulse and dynamics tutorial

eggerdj:
### Summary

This PR shows a tutorial connecting pulse and dynamics.

### Details and comments
We simulate the dynamics of a pulse schedule that implements a sx followed by a sy rotation.


DanPuzzuoli:
Looking really good. To recap our discussion from earlier: I've been working on a tutorial for [optimizing pulses using JAX here](https://github.com/Qiskit/qiskit-dynamics/blob/2eb9af7e82e32e45b05278081d1fab710b765195/docs/tutorials/optimizing_pulse_sequence.ipynb) as part of PR #26 , and we should try to converge on a consistent style.

I think this tutorial and the optimization one above already very similar writing-wise: they are broken into small, relatively self-contained steps/accomplishments for the user to follow. In terms of technical formatting: I've numbered my sections and also outlined all of the steps the tutorial goes through at the beginning, in a numbered list. Take a look and see if you are good with using this formatting or want to change it.

The only other thing I think is where to put the big import block. I kind of like putting them _after_ the introduction, or even as you go (which maybe gives the reader a bit more context for what/where things are being imported from and why). I think mainly though I like to see the title/text introduction at the top of the file. Let me know what you think.

eggerdj:
That makes sense. I updated accordingly in 638e609

DanPuzzuoli (changes requested):
Can you make one final change:
- Instead of importing `HamiltonianModel`, `rotating_wave_approximation`, and `solve_lmde`, can you just do `from qiskit_dynamics import Solver`.
- When constructing the model, instead do the following. Note that I'm setting the rotating frame to the drift so that it will solve in that frame, which shouldn't change your results but should make it faster. (Let me know if this isn't the case.)
```
hamiltonian_solver = Solver(drift=drift, hamiltonian_operators=operators, signals=signals, rotating_frame=drift, rwa_cutoff_freq=2*w)
```
- When solving, instead of calling `solve_lmde`, do:
```
%time sol = hamiltonian_solver.solve(t_span=[0., 2*T], y0=y0, atol=1e-10, rtol=1e-10)
```

The rest should work as before.

DanPuzzuoli (approved):

PR #28, opened 2021-09-16
Improve type handling

### LINKED ISSUE #22: Handling of operator input types to models

There are a few type-handling issues that we need to sort out. Currently, the model classes support operation in both dense and sparse modes, however, regardless of the chosen mode of operation, at instantiation, everything is passed through the `to_array` function, which converts everything to a dense array. If a sparse mode is selected, it will then convert these arrays to sparse. We should set things up so that if a user specifies things in a sparse format then it never gets converted to a dense format.

To do:
- Add a `to_csr` analog of `to_array`, which does the same job but converts everything to the required format for the sparse types (see e.g. the constructor to `SparseOperatorCollection` for the required format).
- Add handling of qutip operator input types to both `to_array` and `to_csr`. This should be done without actually importing qutip, so need to decide how to detect this (`op.__name__` ?) .
- Make sure the use-case of specifying things operators in a sparse format, and requesting models to operate in sparse format, never converts to dense.

brosand:
### Summary
Create several new type handling functions, and tests for those functions, as well as some small improvements to `rotating_frame.py` and the testing framework.

Closes #22 

### Details and comments
Create the functions `to_csr` and `to_numeric_matrix_type`, which improve the handling of sparse matrices in the package. Previously they were always converted to Arrays, but now they maintain their state. In addition, there are some small changes to `rotating_frame.py` which reduce overhead and perform more automatic type conversion. 

Finally, there is a small change to the testing framework, which eliminates some excess code in `__init__.py`, the function `load_tests`, which was causing some issues with test runners.


DanPuzzuoli:
Generally looking good.

For testing of these functions and general type handling behaviour, the main thing we need is tests that actually do some type checking. I think existing tests verify correctness, but don't actually verify type (I may not be 100% correct about this but just perusing the test files I'm not seeing any specific type checks). Specifically, we need tests for:
- `type_utils.py`
    - For `to_array`: Verify that `np.array` -> `np.array`, `Array` -> `Array`, `Operator` -> `Array`, and `List[Operator]` -> `Array` (any others)?
    - For `to_csr`: Verify the corresponding list
    - For `to_numerical_matrix_type`: The same idea
    - For the above, we really want to make sure to probe the logic of the choices, e.g. does it accidentally trigger the `isinstance(x, Iterable)` block when it shouldn't?
- `models/rotating_frame.py` 
    - do the functions now have the correct type handling behaviour? For a diagonal frame, if we call one of the frame transformations with a sparse matrix, do we get sparse out? Does np.array go to to np.array, does Array go to Array? 
- `models/operator_collections.py`
    - Do the `evaluate` functions for `SparseOperatorCollection` and `SparseVectorizedLindbladCollection` return sparse matrices?
- The model classes in `models`
    - This would be more of an "integration test" of several of the above tests, but if we define, e.g. a `HamiltonianModel` to be in sparse mode, does `hamiltonian.evaluate` return a sparse matrix?

Some comments on the structure of the tests:
- I'm noticing that `test_type_utils.py` only contains a single test class (+ the jax version), but `type_utils.py` has multiple unrelated utility functions/classes. The class `TestTypeUtils` needs to be broken up into separate test classes corresponding to each individual class/function in `type_utils`. E.g. one for `StateTypeConverter`, one for `to_array`, one for `to_csr`, etc.
     - An important note in the above is that we do not need JAX versions of tests involving sparse matrices, like `to_csr`, as these don't involve JAX at all. So, an additional benefit of breaking things up is that we can target more specifically what is being tested in JAX mode. (E.g. just don't make a test class inheriting from `TestJaxBase` and from `Testto_csr` (or whatever you end up calling it).)
- The previous sub-comment about not having JAX versions of tests that only work with CSR matrices also applies to all the other tests I mentioned above. E.g. you should move tests for sparse handling in `RotatingFrame` into another test class called something like `TestRotatingFrameSparse`, and again, not make a JAX version of this class. 

Other non-testing related comments:
- Make sure the sparse operator collections are using `to_csr` to do the sparse conversions in the constructors.

test/dynamics/common.py
```diff
@@ -81,3 +102,20 @@ def jit_grad_wrap(self, func_to_test: Callable) -> Callable:
         wf = wrap(lambda f: jit(grad(f)), decorator=True)
         f = lambda *args: np.sum(func_to_test(*args)).real
         return wf(f)
+
+
+class TestQutipBase(unittest.TestCase):
+    """Base class with setUpClass and tearDownClass for setting jax as the
```
DanPuzzuoli, line 108: This is good, but update the doc string to indicate that it's for qutip. The `tearDownClass` isn't necessary for the qutip case, so remove this from the doc string as well.

DanPuzzuoli:
Just took a look at the latest version. For now I'll leave commenting on the changes to operator collections/models for later.

- The changes to `rotating_frame.py` look good, assuming `to_numeric_matrix_type` working.

- I just did some testing and it turns out that that for converting `Operator` and `Qobj` types to array we may not need to do any special checking at all: `Array(op)` works for `op` either an `Operator` or `Qobj`. Both objects implement the `__array__` method, so `np.array()` and `jnp.array()` work on both, and hence `Array()` works on both.
- With the above we can simplify the logic of `to_array` as follows:
```
def to_array(op: Union[Operator, Array, List[Operator], List[Array], spmatrix]):
    """Convert an operator or list of operators to an Array.
    Args:
        op: Either an Operator to be converted to an array, a list of Operators
            to be converted to a 3d array, or an array (which simply gets
            returned)
    Returns:
        Array: Array version of input
    """
    if op is None:
        return op

    if isinstance(op, np.ndarray):
        if dispatch.default_backend() in [None, "numpy"]:
            return op
        else:
            return Array(op)

    if isinstance(op, Array):
        return op
    
    if issparse(op):
        return Array(op.toarray())

    if isinstance(op, Iterable):
        op = Array([to_array(sub_op) for sub_op in op])
    else:
        op = Array(op)
        
    if op.backend == "numpy":
        return op.data
    else:
        return op
```

- In the above:
   - The special cases of `np.ndarray`, `Array`, and `issparse` (for which `np.array()` doesn't work on) are all at the top (as before).
   - The last type check is to see if it's an `Iterable` - if not it just tries to call `Array` on it (this handles the `Operator` and `Qobj` cases).
   - If it _is_ an `Iterable`, it loops through the entries recursively calling `to_array` on them. This allows the `Iterable` case to be handled in a single `if` statement, and all the special cases for different lists in different formats gets sent back to the individual case handling already present in `to_array`.
- Looking at `to_csr`, I see that you're handling iterables like this there already. I'd suggest changing `to_csr` to the following though:
```
def to_csr(op: Union[Operator, Array, List[Operator], List[Array], spmatrix]) -> csr_matrix:
    """Convert an operator or list of operators to a sparse matrix.
    Args:
        op: Either an Operator to be converted to an sparse matrix, a list of Operators
            to be converted to a 3d sparse matrix, or a sparse matrix (which simply gets
            returned)
    Returns:
        csr_matrix: Sparse matrix version of input
    """
    if op is None:
        return op

    if isinstance(op, csr_matrix):
        return op 
    if isinstance_qutip_qobj(op):
        return op.data
    if (isinstance(op, Array) or isinstance(op, np.ndarray)) and op.ndim < 3:
        return csr_matrix(op)
    
    if isinstance(op, Iterable):
        return [to_csr(item) for item in op]
    else:
        return csr_matrix(op)
```
- In the above:
    - We leave the `Qobj` data as is: it's a special subclass of `csr_matrix`, so I think it's fine to not do `csr_matrix()` on it.
    - The `Operator` case is now just handled by the final `else` statement.
    - For the `Iterable` case, we call back to `to_csr` rather than using `csr_matrix`, so that all the special type handling for individual matrices gets inherited to this case.

For the tests:
- `TestQutipBase` looks good - though I added a single comment to update the doc string.
- For the tests for `to_array`, as tedious as this may seem, I think it would be cleaner to have short tests that explicitly test one type of conversion, e.g.:
     - `test_to_array_Operator`, `test_to_array_ndarray`, `test_to_array_Operator_list`, etc. These tests can both test the type of the output, as well as the value. This is in contrast to the current tests, where one is testing the value of all different types of inputs, and the other is testing type.
     - This should make it much more obvious what the tests are doing, as they will be much simpler and the name indicates more clearly what the test case is meant to be doing. I.e. the name `test_to_array` doesn't indicate anything about what the test is doing aside from that it's for the function `to_array`. (Also, I know this setup existed before your changes :P .)
- The same applies to the tests for `to_csr`.

qiskit_dynamics/type_utils.py
```diff
@@ -320,39 +322,125 @@ def vec_dissipator(L: Array):
     )
 
 
-def to_array(op: Union[Operator, Array, List[Operator], List[Array], spmatrix]):
-    """Convert an operator or list of operators to an Array.
+def isinstance_qutip_qobj(obj):
```
DanPuzzuoli, line 325: Don't forget type hints here. You can use: `any` to refer to `obj` being able to literally be anything. Also the doc string needs proper:
```
Args:
     obj: Object to check.

Returns:
 ....
```

DanPuzzuoli, line 325: Don't forget type hints here. You can use: `any` to refer to `obj` being able to literally be anything. Also the doc string needs proper:
```
Args:
     obj: Object to check.

Returns:
 ....
```

qiskit_dynamics/type_utils.py
```diff
@@ -320,39 +322,125 @@ def vec_dissipator(L: Array):
...
+    return False
 
+
+def to_array(op: Union[Operator, Array, List[Operator], List[Array], spmatrix], no_iter=False):
+    """Convert an operator or list of operators to an Array.
     Args:
         op: Either an Operator to be converted to an array, a list of Operators
             to be converted to a 3d array, or an array (which simply gets
             returned)
+        no_iter (Bool): Boolean determining whether to look inside iterables to make multiple
```
DanPuzzuoli, line 345: I think this explanation of `no_iter` is a bit unclear. You could say something as simple as: "Whether to recursively unroll `Iterables`."

test/dynamics/common.py
```diff
@@ -82,4 +98,21 @@ def jit_grad_wrap(self, func_to_test: Callable) -> Callable:
...
+
+    Test cases that inherit from this class will automatically work with jax
+    backend.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        try:
+            # pylint: disable=import-outside-toplevel
+            import qutip
```
DanPuzzuoli, line 116: Just disable all pylint warnings associated with this import. I really don't think there's anything else that can be done.

test/dynamics/test_type_utils.py
```diff
@@ -256,22 +259,189 @@ def test_sparse_commutator_dissipator(self):
             )
         )
 
-    def test_to_array(self):
-        """Tests for to_array"""
+    # def test_to_array(self):
```
DanPuzzuoli, line 262: Don't forget to delete this.

test/dynamics/test_type_utils.py
```diff
@@ -256,22 +259,189 @@ def test_sparse_commutator_dissipator(self):
...
+        self.assertAllClose(to_array(ndarray), ndarray)
+
+    def test_to_array_sparse_matrix(self):
+        """Tests for to_array with a single sparse matrix"""
+        op = Operator.from_label("X")
+        spm = csr_matrix(op)
+        ar = Array(op)
+        self.assertAllClose(to_array(spm), ar)
+
+    def test_to_array_Operator_list(self):
```
DanPuzzuoli, line 292: The input here isn't a list of operators.

test/dynamics/test_type_utils.py
```diff
@@ -256,22 +259,189 @@ def test_sparse_commutator_dissipator(self):
...
+    def test_to_csr_sparse_matrix(self):
+        """Tests for to_csr with a single sparse matrix"""
+        op = Operator.from_label("X")
+        spm = csr_matrix(op)
+        self.assertAllCloseSparse(to_csr(spm), spm)
+
+    def test_to_csr_Operator_list(self):
+        """Tests for to_csr with a list of operators"""
+        list_of_ops = [[[0, 1], [1, 0]], [[0, -1j], [1j, 0]], [[1, 0], [0, -1]]]
+        sparse_matrices = [csr_matrix(op) for op in list_of_ops]
```
DanPuzzuoli, line 346: Not an Operator list.

test/dynamics/test_type_utils.py
```diff
@@ -256,22 +259,189 @@ def test_sparse_commutator_dissipator(self):
...
+        self.assertAllClose(to_numeric_matrix_type(list_of_arrays), normal_array)
+        self.assertAllClose(to_numeric_matrix_type(op_arr), list_of_arrays)
         for i in range(3):
-            self.assertAllClose(sparse_matrices[i].toarray(), normal_array[i])
+            self.assertAllCloseSparse(
+                to_numeric_matrix_type(sparse_matrices)[i], sparse_matrices[i]
+            )
+
+
+class TestTypeUtilsQutip(TestTypeUtils, TestQutipBase):
```
DanPuzzuoli, line 411: This test shouldn't inherit from `TestTypeUtils`, as it will rerun all of those tests, even though it's only adding a single test.

Set it to to just inherit from the usual dynamics base class.

test/dynamics/test_type_utils.py
```diff
@@ -256,22 +259,189 @@ def test_sparse_commutator_dissipator(self):
...
+        qutip_qobj = [qutip.Qobj(op) for op in list_of_ops]
+        self.assertAllCloseSparse(to_numeric_matrix_type(qutip_qobj)[0], sparse_matrices[0])
+        self.assertAllCloseSparse(to_csr(qutip_qobj)[0], sparse_matrices[0])
+        self.assertAllCloseSparse(to_array(qutip_qobj)[0], normal_array[0])
 
 
 class TestTypeUtilsJax(TestTypeUtils, TestJaxBase):
     """Jax version of TestTypeUtils tests.
 
     Note: This class has no body but contains tests due to inheritance.
```
DanPuzzuoli, line 432: Delete this note.

Also there is similar redundancy in this test class as with the qutip case, there is no need to rerun the `to_csr` of `TestTypeUtils` tests with the JAX backend set to default.

Really the tests in this file need to be broken into classes:
- Tests for `StateTypeConverter` + a JAX version, e.g. `TestStateTypeConverter` and `TestStateTypeConverterJax`
- Tests for `to_array` + a JAX version, `Test_to_array` and `Test_to_array_JAX`
- Tests for `to_csr` (no need for a JAX version), `Test_to_csr`
- Tests for qutip input types `Test_qutip_type_handling`

test/dynamics/models/test_rotating_frame.py
```diff
@@ -462,6 +463,108 @@ def test_vectorized_frame_basis(self):
...
+        """Test frame transformations with no frame."""
+
+        # rotating_frame = RotatingFrame(Array(np.zeros(2)))
+        rotating_frame = RotatingFrame(None)
+
+        t = 0.123
+        y = csr_matrix([1.0, 1j])
+        out = rotating_frame.state_into_frame(t, y)
+        self.assertAllCloseSparse(out, y)
+        assert isinstance(out, csr_matrix)
```
DanPuzzuoli, line 478: use `self.assertTrue(x)` instead of `assert x`, so that this is recorded as a failure rather than an error. Same with all below.

DanPuzzuoli (changes requested):
Additional changes:
- For operator collections: make sure that `to_csr` is being called on the inputs to _all_ constructors for sparse types. E.g. `SparseVectorizedLindbladCollection` and `SparseLindbladCollection`. `to_array` is still being called on them before they are converted to sparse. I'm a bit worried that `np.round` may only work on arrays. If you can't figure out something quickly then we should make a note of it in an issue and figure it out in a subsequent PR.
- We still don't have any type testing for operator collections but that's fine for now.

test/dynamics/test_type_utils.py
```diff
@@ -256,22 +259,189 @@ def test_sparse_commutator_dissipator(self):
...
+
+    def test_to_array_nparray_list(self):
+        """Tests for to_array with a list of numpy arrays"""
+        list_of_ops = [[[0, 1], [1, 0]], [[0, -1j], [1j, 0]], [[1, 0], [0, -1]]]
+        ndarray_list = [np.array(op) for op in list_of_ops]
+        list_of_arrays = [Array(op) for op in list_of_ops]
+        self.assertAllClose(to_array(ndarray_list), list_of_arrays)
+
+    def test_to_array_3d_array(self):
+        """Tests for to_array with a list of numpy arrays"""
```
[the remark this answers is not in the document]
brosand: this is for converting list of arrays to 3d array, is my naming confusing? I will change

DanPuzzuoli (approved):

PR #30, opened 2021-09-21
Resolving small API changes

### LINKED ISSUE #29: Collection of small model API issues
A group of small API changes:

- Argument naming conventions
    - Change `drift` kwarg to `static_hamiltonian` or `static_operator` where applicable
- Somewhat related to the above:
    - Add static dissipator for lindblad?
    - All model classes should be able to be instantiated with/without various operator types. E.g. currently `operators` is required for `HamiltonianModel`, but we should be able to instantiate with just a drift. Similarly, a `LindbladModel` should be able to be instantiated with _just_ dissipators (no Hamiltonian part).
        - It should be "simple" to change this but may be somewhat non-trivial as the current structure implicitly makes these assumptions in a bunch of places.
- Add validation to Lindblad signal setter
    - None currently exists.
- Add validation to Hamiltonian operators in Lindblad
    - None currently exists.
- For some reason Lindblad constructor needs Hamiltonian signals.
    - This should not be necessary.

Depending on the scope of these they could be handled in a single PR, but if any require substantial changes they should be done in subsequent PRs.

DanPuzzuoli:
### Summary

Closes #29 

### Issue recap

Recreating list for record keeping purposes - so things can be crossed out.

- (partial) Argument naming conventions
    - Change `drift` kwarg to `static_hamiltonian` or `static_operator` where applicable
- Somewhat related to the above:
    - Add static dissipator for lindblad?
    - All model classes should be able to be instantiated with/without various operator types. E.g. currently `operators` is required for `HamiltonianModel`, but we should be able to instantiate with just a drift. Similarly, a `LindbladModel` should be able to be instantiated with _just_ dissipators (no Hamiltonian part).
        - It should be "simple" to change this but may be somewhat non-trivial as the current structure implicitly makes these assumptions in a bunch of places.
- (partial) Add validation to Lindblad signal setter
    - None currently exists.
- (partial) Add validation to Hamiltonian operators in Lindblad
    - None currently exists.
- (partial) For some reason Lindblad constructor needs Hamiltonian signals.
    - This should not be necessary.

While trying to resolve some of the above I realized there is another issue we should address with this:
- Duplication of `operator` and `static_operator` information in the models and operator collections. This isn't such a big deal but it does make tracking down behaviour somewhat confusing. It would be good to eliminate this duplication, which should eliminate some of the circularity in the way the property setters are utilized in the constructor.

### Details and comments

Current status:
- Changed `drift` kwarg -> `static_operator` or `static_hamiltonian` depending on context. Tests are passing.
    - Remaining open issue: Due to the divergence between the `static_operator` and `static_hamiltonian` names, some inheritance structure has been messed up across the models and operator collections. I'm going to leave this unresolved at the moment, as decisions about handling this are going to be affected by adding a static dissipator to `LindbladModel` and the corresponding operator collections.
    - No need to add new tests for this as a the "drift" handling is already heavily tested.
- Added validation to `LindbladModel` signal setter. 
    - More or less already existed in the constructor, so just moved it to the signals setter.
    - Need to add tests for this, but will wait as the automatic setting of dissipator signals to value `1.` should be revisited when static dissipator is added.
- Added validation to Hamiltonian operators in `LindbladModel`
    - Refactored the validation step in `HamiltonianModel` to use an `is_hermitian` function which verifies if a 2d array is Hermitian or if a 3d array represents a list of Hermitian operators.
    - Added validation of hamiltonian terms in `LindbladModel`
    - Added tests for both `HamiltonianModel` and `LindbladModel` that this validation step is actually working, as well as verifying that it can be overridden by setting `validate=False`.
    - Left this part marked as (partial) as this validation will ultimately need to handle csr_matrix types as well, and whether that will be handled in this PR depends on which of this and #28 is completed first.
- Allow `hamiltonian_signals` to be optional in `LindbladModel` constructor.
    - Made this optional but need to test. 
- Eliminate duplication of operator storage in models and Operator collections
    - Just did this for `GeneratorModel` and `HamiltonianModel`, and for their corresponding operator collections
    - Created two helper functions: one to update the internally stored static operators and operator list from one frame to another, and one to setup the operator collection for a given evaluation mode. These are now called by the constructor and property setters where required. (Originally this code was solely in property setters, so it required setting properties in certain orders, and required accessing data in weird ways, which is why the duplication of the operator data was necessary.)
    - Added tests for `setup_operators_in_frame` but need to add tests for `construct_operator_collection`
    - Need to clarify the interfaces for `BaseOperatorCollection` and `BaseGeneratorModel`, and the `DenseOperatorCollection` and `SparseOperatorCollection`.
         - Mainly, with other proposed changes to argument/property names, need to re-evaluate some of the abstract property names
    - Need to do the same for `LindbladModel` and its corresponding operator collections

Another general note:
- Need to make sure the model constructors make correct use of `to_array` v.s. `to_numeric_matrix_type`, though depending on merge order this responsibility may be with this PR or with #28 .


DanPuzzuoli:
This PR is getting somewhat sprawling but it should hopefully simplify/clarify things. From a user perspective very little has changed (renaming `drift` to `static_operator` or `static_hamiltonian`), but this (and other things) are modifying/simplifying the whole model and operator collection structure, so it's impacting a lot of lines of code and getting a bit hard to think about in the simple terms of each of the isolated points in the main comment.

Plan going forward:
- Finish simplifying `BaseGeneratorModel`, `GeneratorModel`, `HamiltonianModel`, `BaseOperatorCollection`, `DenseOperatorCollection`, and `SparseOperatorCollection`
    - Knowing the naming structure we want now, remove any abstract properties from the base classes that no longer apply to the Lindblad case
    - Set the new helper functions as static class methods of `GeneratorModel`, and set `HamiltonianModel` to use them where appropriate instead of importing
    - Document and test them (in particular, configurations in which some args are `None`.) Some of these tests may need to wait until the type handling PR is in, but I can at least write what I expect they should be and temporarily skip them until I can rebase on that.
- Do the same to `LindbladModel` and its corresponding operator collections

Once the above two points are in:
- Modify the model classes and operator collections so that `operators`/`hamiltonian_operators` can be supplied as `None`
    - Perhaps this should be done in conjunction with the top two?
- Add the type handling/testing for the points in the main comment

Additionally:
- I think I may delete `CallableGenerator`. It's no longer used anywhere and just adds more confusion with the model interfaces
- Going to scrap the plan to add a `static_dissipator` in this PR. This should be easy-ish to do after this PR and can be a standalone PR if we decide we really need it.

DanPuzzuoli:
Note:
- Getting `GeneratorModel.transfer_operators_between_frames` to work for csr matrices required modifying `RotatingFrame._conjugate_and_add` to work for CSR matrices (if input is csr then it will convert the `op_to_add_in_fb` to a csr). This is a little bit of a hack, but the type handling at the moment is still a little bit unclear. It's definitely getting better, but still needs some work. That being said, my main concern at the moment is: if the user starts with sparse, wants to work with sparse, and only builds/transforms models using standard functions, the matrices should _never_ be turned into dense arrays. This at least achieves that.

DanPuzzuoli:
Just added option for Dense/SparseOperatorCollection to be defined with `None` for `operators`, and evaluate with only a static operator. Added tests for this case as well as tests for error modes if collections are evaluated with neither of `operators` and `static_operators`.

DanPuzzuoli:
Finished up with allowing GeneratorModel and HamiltonianModel to have only a static component.
- Added tests to verify evaluation works.
- Added validation check and corresponding test: these classes require at least one of `static_operator` and `operators` to be non-empty.
- Modified signal setting validation to handle case that `operators is None`. If `operators is None` it now raises an error if `signals` is set to anything other than `None`. Added test case for this.

I think the last thing that needs to be done with these classes is to make the Hamiltonian operator validation step work for both arrays and CSR (right now it auto-converts to array).

DanPuzzuoli:
Current status:
- All required changes have been made to `GeneratorModel` and `HamiltonianModel`, and the corresponding operator collections.
- Need to do the same for LindbladModel and associated operator collections:
    - Eliminate redundant operator storage (follow setup for `GeneratorModel`)
    - Make interface for lindblad associated operator collections cleaner, similar to the others (maybe a new abstract base class?)
    - Add ability for lindblad collections to function by specifying only one of all the possible operator types.
        - Add static dissipators at this point?
    - Update LindbladModel to accept only one of all possible operator types.
    - Change signal setter/getter for LindbladModel into two, one for hamiltonian_signals and one for dissipator_signals.

DanPuzzuoli:
Current status:
- `LindbladModel` has been setup and tested to work in various configurations of having different types of operators specified.
- `LindbladModel.signal` setter has been updated to align with the signal setter of `GeneratorModel` in terms of validation. It still sets the dissipator coefficients to a list of ones if they are specified as `None`, but this will be changed once the static dissipators are added. 
- Linblad-associated operator collections have been revamped. These could still use improvement, but I think they are starting to converge.
     - The vectorized versions have been setup now to inherit essentially all of their functionality from other classes. The MRO is a bit weird for these so I don't think this is yet the "right" version, but whether or not this should be improved upon in this PR or in subsequent work will be a matter of discussion in review.
     - There are not a lot of conditional checks in RHS evaluation - a potential thing to modify for speed in the future.

To do:
- Add static dissipators + testing
- Could always add more tests generally with all of these changes, but i'm starting to feel fairly satisfied with where the tests are at.

DanPuzzuoli:
Note: still using the above list of items as the current to do list, but currently ran into the issue that `rotating_wave_approximation` doesn't work natively on sparse representations. Next thing to do is to clean this up and test it.

DanPuzzuoli:
Note: I've decided to not update `rotating_wave_approximation` to work directly with sparse matrices. This is a standalone task that deserves some thought. For now I've modified `rotating_wave_approximation` to explicitly convert everything into dense arrays. Note that it still work son sparse models, it just uses dense arrays intermediately while doing the approximation.

qiskit_dynamics/models/generator_model.py
```diff
@@ -529,3 +420,141 @@ def evaluate_rhs(self, time: float, y: Array, in_frame_basis: Optional[bool] = F
             return op_combo @ y
 
         return out
+
+    @classmethod
+    def transfer_static_operator_between_frames(
```
chriseclectic, line 425: Is there a particular reason this and the other transfer operators func are class methods of this class rather than just a function you use directly? They don't seem to make any use of the class instance.

qiskit_dynamics/models/generator_model.py
```diff
@@ -529,3 +420,141 @@ def evaluate_rhs(self, time: float, y: Array, in_frame_basis: Optional[bool] = F
...
+            # For sparse case, if list, loop
+            if isinstance(operators, list):
+                operators = [new_frame.operator_into_frame_basis(op) for op in operators]
+            else:
+                operators = new_frame.operator_into_frame_basis(operators)
+
+        return operators
+
+    @classmethod
+    def construct_operator_collection(
```
chriseclectic, line 529: Does this need to be part of the public API or can it be `_construct_operator_collection`. If public does it need to be a method (do subclasses overload it?) or can it be a separate function.

qiskit_dynamics/models/hamiltonian_model.py
```diff
@@ -34,14 +38,14 @@ class HamiltonianModel(GeneratorModel):
...
-    where :math:`H_j` are Hermitian operators, :math:`H_d` is the drift component,
+    where :math:`H_j` are Hermitian operators, :math:`H_d` is the static component,
     and the :math:`s_j(t)` are either :class:`~qiskit_dynamics.signals.Signal` objects or
     are numerical constants. Constructing a :class:`~qiskit_dynamics.models.HamiltonianModel`
     requires specifying the above decomposition, e.g.:
 
     .. code-block:: python
 
-        hamiltonian = HamiltonianModel(operators, signals, drift)
+        hamiltonian = HamiltonianModel(operators, signals, static_operator)
```
chriseclectic, line 48: The signature of HamiltonianModel below is `(static_operator, operators, signals)` so the above would give an error. You either need to change the order or use kwargs:
```suggestion
        hamiltonian = HamiltonianModel(static_operator=static_operator, operators=operators, signals=signals)
```

qiskit_dynamics/models/hamiltonian_model.py
```diff
@@ -110,25 +112,70 @@ def rotating_frame(self) -> RotatingFrame:
...
+        adj = None
+        if operators.ndim == 2:
+            adj = np.transpose(np.conjugate(operators))
+        elif operators.ndim == 3:
+            adj = np.transpose(np.conjugate(operators), (0, 2, 1))
+        return np.linalg.norm(adj - operators) < tol
+    elif issparse(operators):
+        return spnorm(operators - operators.conj().transpose()) < tol
+    elif isinstance(operators, list) and issparse(operators[0]):
+        return all(spnorm(op - op.conj().transpose()) < tol for op in operators)
```
chriseclectic, line 179: This could potentially be faster if `all` was swapped to `any` and the logic flipped so it can terminate as soon as one fails
```suggestion
        return not any(spnorm(op - op.conj().transpose()) >= tol for op in operators)
```

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,616 @@
...
+          operator, and
+        - :math:`\gamma_j(t)` denotes the signal corresponding to the
+          :math:`j^{th}` Lindblad operator.
+
+    Instantiating an instance of :class:`~qiskit_dynamics.models.LindbladModel`
+    requires specifying the above decomposition:
+
+    .. code-block:: python
+
+        lindblad_model = LindbladModel(static_hamiltonian,
```
chriseclectic, line 67: Might be good to use kwargs instead of args in docs here too

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,616 @@
...
+        """
+
+        if (
+            static_hamiltonian is None
+            and hamiltonian_operators is None
+            and static_dissipators is None
+            and dissipator_operators is None
+        ):
+            raise QiskitError(
+                self.__class__.__name__
```
chriseclectic, line 122:
```python
raise QiskitError(
    f"{type(self).__name__} requires at least one of static_hamiltonian "
    "hamiltonian_operators, static_dissipators, or dissipator_operators "
    "to be specified at construction.")
```

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,616 @@
...
+        self._evaluation_mode = evaluation_mode
+        self.vectorized_operators = "vectorized" in evaluation_mode
+
+        self._rotating_frame = None
+        self.rotating_frame = rotating_frame
+
+        self.signals = (hamiltonian_signals, dissipator_signals)
+
+    @classmethod
+    def from_hamiltonian(
```
chriseclectic, line 152: You could also add an optional `hamiltonian` kwarg to init to implement this as part of regular init? Maybe getting a lot of args for init at this point though...

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,616 @@
...
+
+            # verify signal length is same as operators
+            if len(dissipator_signals) != len(self.get_dissipator_operators()):
+                raise QiskitError(
+                    "Dissipator signals need to have the same length as dissipator operators."
+                )
+
+            self._dissipator_signals = dissipator_signals
+
+    def get_static_hamiltonian(self, in_frame_basis: Optional[bool] = False) -> Array:
```
chriseclectic, line 265: Since we have some methods that use setters, and others that use these set/get functions an idea to make them all use setters would be to add a `in_frame_basis` property with setter, so that in these classes you could have something like:

```python
@property
def in_frame_basis(self):
    return self._in_frame_basis

@in_frame_basis.setter
def in_frame_basis(self, value):
    self._in_frame_basis = value

@property
def static_hamiltonian(self):
    op = self._operator_collection.static_hamiltonian
    if self._in_frame_basis or self.rotating_frame is None:
        return op
    else:
        return self.rotating_frame.operator_out_of_frame_basis(op)
 
@static_hamiltonian.setter
def (self, hamiltonian):
     # etc
```

I'm not sure how often these methods are actually used by users (didn't see them anywhere in tutorial notebooks), but after typing above I saw that mostly they were used in internal codes where you need the argument so maybe this change would be inconvenient there unless you also did something like:

```python
def _get_static_operator(self, in_frame_basis):
 
@property
def static_operator(self):
   return self._get_static_operator(self._in_frame_basis)
# etc
```

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,616 @@
...
+                time,
+                ham,
+                operator_in_frame_basis=True,
+                return_in_frame_basis=in_frame_basis,
+                vectorized_operators=self.vectorized_operators,
+            )
+
+        return ham
+
+    def evaluate(self, time: float, in_frame_basis: Optional[bool] = False) -> Array:
```
chriseclectic, line 468: If `in_frame_basis` was changed to property all the evaluation methods / call methods could be changed to look that up too.

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,616 @@
...
+
+        else:
+            rhs = self._operator_collection.evaluate_rhs(
+                hamiltonian_sig_vals, dissipator_sig_vals, y
+            )
+
+        return rhs
+
+    @classmethod
+    def construct_operator_collection(
```
chriseclectic, line 570:
```suggestion
    def _construct_operator_collection(
```

qiskit_dynamics/models/operator_collections.py
```diff
@@ -124,138 +153,299 @@ class SparseOperatorCollection(BaseOperatorCollection):
...
+    @hamiltonian_operators.setter
+    def hamiltonian_operators(self, new_hamiltonian_operators: Optional[Array] = None):
+        self._hamiltonian_operators = to_array(new_hamiltonian_operators)
+
+    @property
+    def static_dissipators(self) -> Array:
+        return self._static_dissipators
+
+    @static_dissipators.setter
+    def static_dissipators(self, new_static_dissipators: Optional[Array] = None):
```
chriseclectic, line 387: The model classes would have same signature as these collection classes if you can make `in_frame_basis` a property rather than kwarg

qiskit_dynamics/models/operator_collections.py
```diff
@@ -561,31 +731,69 @@ def evaluate_rhs(self, ham_sig_vals: Array, dis_sig_vals: Array, y: Array) -> Ar
...
+                )
 
             out = left_mult_contribution + right_mult_contribution + both_mult_contribution
 
-        else:
+        elif hamiltonian_matrix is not None:
             out = (([hamiltonian_matrix] * y) - (y * [hamiltonian_matrix]))[0]
+        else:
+            raise QiskitError(
+                """SparseLindbladCollection with None for static_hamiltonian,
```
chriseclectic, line 793: Indents look weird here, better to split into multiple line single string:
```
raise QiskitError(
    "line 1 "
    "line 2")
```

qiskit_dynamics/models/generator_model.py
DanPuzzuoli: Can just make it a separate function, same as above.

qiskit_dynamics/models/lindblad_model.py
DanPuzzuoli: I think I'd like to keep this separate. If we add it to the init we'll need to add validation to check if redundant/inconsistent information is being passed and document the multiple modes of behaviour of the init. Not that huge a deal but it feels simpler to keep it structurally separate.

qiskit_dynamics/models/lindblad_model.py
DanPuzzuoli: Yeah they are almost exclusively used internally. I really like this suggestion. I think the approach of just making the `_get`... methods private seems best. Will think about it but will probably proceed with this.

qiskit_dynamics/models/lindblad_model.py
DanPuzzuoli: Good idea.

qiskit_dynamics/models/lindblad_model.py
DanPuzzuoli: Made this change.

qiskit_dynamics/models/lindblad_model.py
DanPuzzuoli: Agreed, changed.

qiskit_dynamics/models/generator_model.py
DanPuzzuoli: Done.

DanPuzzuoli:
I believe I've addressed all comments. Aside from small changes which I've commented on above, substantial changes are:

- Made `in_frame_basis` a property of the models, along with `static_operator`, `operators`, and the corresponding `LindbladModel` equivalents.
   - The original `get_static_operator`, etc... have been made private, and the properties call this, using the `in_frame_basis` property.
- Added `in_frame_basis` kwarg to constructors for the models and `Solver`, and added handling of this kwarg to `solve_lmde`, `solve_ode`, and `rotating_wave_approximation` (all of these functions do computations in the frame basis, and now they need to check whether to transform things into/out of the frame basis at the beginning/end)
- I've made all of the class methods into standalone helper functions.
- Not related to any comments, but something I realized was missed in the `Solver` class: added the `validate` kwarg to Solver construction, so that a Solver can be constructed with validation of hamiltonian operators being skipped.
    - Added a test to check that a `Solver` can be constructed within a function to be jitted.


Still to do:
Started to fix up the notebooks, but there seems to be an error now in the RWA for the lindblad model in `general_demo.ipynb`. Will need to fix this, and add a test to catch whatever is happening.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -88,53 +88,66 @@ class Solver:
...
             hamiltonian_signals: Coefficients for the Hamiltonian operators.
-            dissipator_operators: Optional dissipation operators.
+            static_dissipators: Constant dissipation operators.
+            dissipator_operators: Dissipation operators with time-dependent coefficients.
             dissipator_signals: Optional time-dependent coefficients for the dissipators. If
                                 ``None``, coefficients are assumed to be the constant ``1.``.
-            drift: Hamiltonian drift operator. If a ``rotating_frame`` is specified, the
-                   ``frame_operator`` will be subtracted from the drift.
             rotating_frame: Rotating frame to transform the model into.
+            in_frame_basis: Whether to represent the model in the basis in which the rotating
```
chriseclectic, line 116: Does this also mean its solved in the rotating frame basis? Might be good to make that clear in this docstring. Also should default be True instead of False? If you define a frame do you typically want to solve in that frame?

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -0,0 +1,654 @@
...
+            hamiltonian_operators: List of operators in Hamiltonian with time-dependent
+                                   coefficients.
+            hamiltonian_signals: Time-dependent coefficients for hamiltonian_operators.
+            static_dissipators: List of dissipators with coefficient 1.
+            dissipator_operators: List of dissipator operators with time-dependent coefficients.
+            dissipator_signals: Time-dependent coefficients for dissipator_operators.
+            rotating_frame: Rotating frame in which calcualtions are to be done.
+                            If provided, it is assumed that all operators were
+                            already in the frame basis.
+            in_frame_basis: Whether to represent the model in the basis in which the rotating
```
chriseclectic, line 111: What happens if you supply a rotating frame, but don't put "in_frame_basis"? Do we need `in_frame_basis` as a kwarg or should it automatically be set to `True` if you supply a rotating frame and then you can set it to false after construction if you don't want to go into that frame?

DanPuzzuoli:
To explain `in_frame_basis`:
- Setting it to `True` makes the model evaluation functions evaluate in the basis in which the frame is diagonal (as per the decomposition given by `np.linalg.eigh`).
    - If the `rotating_frame` is `None`, or was supplied in already diagonal form as a 1d array, this kwarg/property has no effect.
- I thought the default `False` was natural as "entering a rotating frame" is perfectly well defined irrespective of what basis things are written in, so setting `in_frame_basis=True` feels like an _additional_ action that the user should have to manually specify.
- In terms of which basis to solve in: I do believe that it's always better to solve in the basis in which the frame is diagonal.
    - It is strictly cheaper than in a basis in which the frame is non-diagonal (for this package everythign is internally stored in the frame basis, so if you compute things in a different basis there is the additional step of mapping the generator/state out of the frame basis).
    - I've written `solve_ode` and `solve_lmde` to _always_ solve in the frame basis, whether `model.in_frame_basis` is `True` or `False`. Before solving, if `model.in_frame_basis=False`, it will temporarily set it to `True`, map the state into this basis, solve, then map the output back to the original basis and set `model.in_frame_basis` back to `False`.

As explained in the second point it seems to me to default it to `True`, and due to the last point it shouldn't have any effect on the efficiency of solving one way or the other.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: See below - I can add somethign to the doc-string explaining the behaviour.

chriseclectic (approved):

PR #31, opened 2021-09-23
Improved documentation

eggerdj:
* Renamed the converters to pulse.
* Deleted tutorials by moving their core to the docs.

### Summary
This PR improves the docs and renames `converters` to `pulse`.

### Details and comments

A couple of tutorials have been deleted by moving their content to the documentation.


qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -31,13 +31,22 @@
 
 
 class InstructionToSignals:
-    """Converts pulse instructions to Signals for the Aer simulator."""
+    """Converts pulse instructions to Signals to be used in models.
+
+    Qiskit-dynamics simulates the time evolution using the :class:`Signal` class.
```
DanPuzzuoli, line 36: I think this general blurb about dynamics using `Signal`s, so simulating pulse schedules requires a means of conversion should go into the module doc string.

qiskit_dynamics/pulse/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+# that they have been altered from the originals.
+
+"""
+==============================================
+Pulse (:mod:`qiskit_dynamics.pulse`)
+==============================================
+
+.. currentmodule:: qiskit_dynamics.pulse
+
+This module contains converters that convert Qiskit pulse schedules into signals such that
```
DanPuzzuoli, line 22: Since we might be putting more stuff into this module it might be worth making this first sentence a bit more general, like:

"This module contains tools for interfacing qiskit pulse with qiskit dynamics."

Then have a section called "Converters" that talks more specifically about schedule -> Signal conversion. The general statement about the need for schedule -> Signal conversion in `InstructionToSignals` (mentioned below) I think would be better here than in the class doc string.

DanPuzzuoli (changes requested):
I just commented in some spots that I think some of the text could be moved around: e.g. general statements about requiring schedule -> Signal conversion should be in the module doc string instead of the `InstructionsToSignal` class doc string. Also the description of the module could be slightly more general as we will definitely be putting more things into this folder.

Other than this looks good though.

qiskit_dynamics/pulse/pulse_to_signals.py
eggerdj: See here: 094e9f2

qiskit_dynamics/pulse/__init__.py
eggerdj: See here: 094e9f2

DanPuzzuoli (approved):
Looks good to me, though we'll need to rebase/rerun these tests once the current issues with JAX are sorted out. Will mark this as Approved though.

PR #32, opened 2021-09-23
A general simulation tutorial

haggaila:
### Summary

Still missing some references at the bottom.
The call to plot_bloch_vector() from qiskit visualization does not show anything.

### DIFF
--- a/docs/tutorials/Lindblad_dynamics_simulation.ipynb
+++ b/docs/tutorials/Lindblad_dynamics_simulation.ipynb
@@ -1,0 +1,347 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Solving the Lindblad dynamics of a qubit chain\n",
+    "\n",
+    "In this tutorial we walk through an example of solving the time evolution of the density matrix of coupled qubits using `qiskit_dynamics`. The model that we solve consists of a time-independent Hamiltonian for qubits interacting with their nearest neighbors along a ring. In addition, energy relaxation terms acting on each qubit will be modeled by using a non-Markovian master equation in Lindblad form. We will discuss the system's steady state as a function of the model parameters.\n",
+    "\n",
+    "In the sections below we define a model, solve the dynamics and plot some observables using the following steps:\n",
+    "1. Define the number of qubits and precompute some matrix operators.\n",
+    "2. Define all relevant parameters and setup a `Solver` instance with the model of the system, consisting of the Hamiltonian and the jump operators of the Lindblad dissipator.\n",
+    "3. Define the initial state and simulation and output times for the initial value problem, and evolve the system state.\n",
+    "4. Define observables and calculate their values as a function of time.\n",
+    "5. Plot some single-qubit observables.\n",
+    "6. Analyze and discuss the results."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1. Prepare the single-qubit operators\n",
+    "\n",
+    "In `qiskit-dynamics`, dynamical simulations are defined using the operators that act the states appearing in the differential equations. We start by creating the single-qubit operators that act on each qubit in the simulation, and are represented as matrices in the Hilbert space of $N$ qubits. Using `qiskit` library routines, it is easy to create $N$-qubit operators that are products of basic single-qubit operators, by using labels (descriptive strings) with each character in the label defining one qubit.\n",
+    "\n",
+    "Below, we first set the number of qubits $N$ to be simulated, and then prepare and store the single-qubit Pauli operators that will be used in the rest of this tutorial."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\n",
+      "Single-qubit operators generated.\n"
+     ]
+    }
+   ],
+   "source": [
+    "from qiskit.quantum_info import Operator\n",
+    "\n",
+    "N = 6\n",
+    "\n",
+    "x_ops = []\n",
+    "y_ops = []\n",
+    "z_ops = []\n",
+    "qubits = range(N)\n",
+    "identities = ['I'] * N\n",
+    "\n",
+    "for i in qubits:\n",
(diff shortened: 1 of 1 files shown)


DanPuzzuoli:
This simulation is really cool. I feel though like this may be more of an "advanced" tutorial - the model and dynamics are very non-trivial. Is this what you were referring to originally about having two simulation tutorials? Maybe we can have this and a more basic one, e.g. observing rabi oscillations of a single qubit driven on resonance?

Regardless of the above,  I think this tutorial is good but just needs to be edited down a bit to be simpler:
- Given how advanced the model is, it might be worth giving more explanation of what it is at the beginning of the tutorial, so that when you get to section 2 setting up the `Solver`, there isn't as much text before getting into the code.
    - The Hamiltonian and dissipator information in section 2 can go at the beginning.
- Section 1 
     - I think we can simplify this by taking advantage of `Pauli` in quantum info (this is suggested by @chriseclectic ). Setting `z = Operator(np.zeros((dim, dim)))`, the Pauli X operator on qubit with index 3 for example, can be created via: `z + Pauli('X')(3)`. This can maybe even be incorporated directly into the Hamiltonian generation code for section 2. Two qubit operations can be created in the same way, e.g. `z + Pauli('XY')(i, j)`. The addition of `z` is a bit of a hack but I feel like it's more readable than the string manipulation and `''.join` that is currently being used.
- Section 2
     - Can be simplified according to the above two comments.
- Section 3
     - Less explanation in the text is required - I think you can get rid of everything but the first sentence.
     - DensityMatrix also has a `from_label` constructor, you can do: `DensityMatrix.from_label('0' * N)`
     - Use `np.linspace` to generate the time points. This will automatically include the endpoints and avoid requiring to check if the endpoint is in the list.
     - As the model is constant in time, you could use the `scipy_expm` method, though it will require setting up the solver to be in a vectorized evaluation mode. E.g. at construction add the initial kwarg: `evaluation_mode='dense_vectorized'`, then when calling `solve` do `method='scipy_expm', max_dt = ...`, where `...` is something at least as large as the size of the steps between time points. Alternatively, if using a standard ODE solver I think you will want to go into the frame of the drift. 
- Section 4
     - Interesting comments about the mean bloch vector v.s. individual qubits bloch vector. 
- Section 5 
     - Cool!
- Section 6
     - I think this should be made way shorter and maybe just be moved after the plot of Section 5. This stuff is interesting but the tutorial should be more concerned with demonstrating how the code can be used to perform a task, rather than interpretation of the results of this specific simulation. 
- No need to have print statements at the end of code blocks - we already know what the code block is for
- Since you're not using JAX there's no need to use the `Array` class, you can just use standard numpy arrays.

haggaila:
Thank you @DanPuzzuoli for the review, I'll apply it. I have two questions - I did not know how to use `np.linspace` since it accepts a number of steps and not the time step. Passing the number of steps would require a computation since IMO it should be determined by the time step and the total simulation time span. What do you suggest? About `Array` - without using it I got type hint warnings, what do you suggest? Thanks.

DanPuzzuoli:
Where are the type-hint warnings appearing? Depending on where it is we should create an issue for it, but whatever the case you should avoid using `Array` in this notebook.

For using `np.linspace` can you just choose `N` so you get a suitable `dt` for the time span you're using? I mainly just don't like the `if` statement adding a time value if necessary, and would prefer you just setting `N` directly to whatever value works.

haggaila:
@DanPuzzuoli I applied your corrections except two that I didn't know how:

1. I didn't know how to use the construct `z + Pauli('X')(3)`. I found the Pauli constructor `Pauli(label='X')`, but then I got an error that the Pauli class is not callable and didn't find a syntax that achieves it. What should it be?
2. About vectorization, I wasn't sure if you want me to use it. Since I expect it allows simulating only about half the number of qubits, I didn't apply this modification. About going to the drift frame - I tried adding to the `Solver` constructor the parameter `drift=H`, but that seemed to work slower (for 7 qubits on my laptop it took ~18 seconds instead of ~9 without this parameter), so I thought it may not be what you meant. Please let me know what you suggest to do.
3. I removed the `Array` usages. There are type warnings in the Solver constructor (with `hamiltonian_operators` and `dissipator_operators`) and the `solve()` method (with the `t_span` parameter).
4. I can work on a simpler tutorial following this one.

DanPuzzuoli:
The "Simulating rabi oscillations..." tutorial looks really good. I think the one main comment is that it'd be good to establish some notational consistency between tutorials:
- The pulse tutorial, and the one I'm writing for JAX, use X, Y, Z for Pauli operators instead of s_x, s_y, s_z (and \sigma_j in latex). I prefer this as it corresponds directly to the symbol used to construct it (e.g. `Operator.from_label('Z')`).
- Frequencies are usually written as either \omega (rad/s) or \nu (1/s). Can you use this instead of f?
- I'm wondering if you need to write the form of the lindblad equation, or if you should just state the jump operators?

Aside from this, some minor preferences:
- I know this is contradicting a bit what I said in the previous tutorial, but can you put the Hamiltonian description into the first section, and the dissipator description into the last section when you add dissipation to the model? In contrast to the other tutorial, I think the model here is "standard" enough to not require explicitly stating what it is in the intro.
- Can you change the title of section 1 to something like: "Setup the solver with the Hamiltonian model". I feel like the term "hamiltonian" should be used in it.
- Change the title of section 2 to just: "Solve the system"

Other than this looks good to me! (Will comment on the other tutorial in a separate comment.)

DanPuzzuoli:
For the spin chain tutorial:
- Same general notational convention comments from the previous comment apply.
- Similar to previous comment: Do we need to write the Lindblad equation here or can we just state what the dissipators are?
- Can you just put `2 * np.pi` everywhere instead of using the `_2_pi`.
- There's still some print statements to be eliminated.
- Can you put a comment line above the `DensityMatrix('0' * N)` saying: "# density matrix with all qubits in ground state" or something.
- For the issue with constructing the Pauli's and getting the "no callable" error, do you need to just update terra? It works fine for me when I try it.

haggaila:
@DanPuzzuoli I applied the changes in your two comments above.
For the question, I think that it is a bit better to write down the Lindblad equation explicitly with the full expressions for the dissipators, to avoid ambiguity that comes from different possible notations (where is the dagger - on the left or right jump operator, and is there a factor of 1/2 or a factor of 2 on the other term in the dissipator), and also just for completeness. Let me know whether to remove it nevertheless. Thanks.

PR #33, opened 2021-09-24
Add documentation for dispatch module

chriseclectic:
### Summary

Adds documentation for dispatch module.

* The init level documentation for the module is added to the User Guide section of docs instead of the API Reference like the existing modules (which should probably also be moved to user guide section).
* A separate rst file listing only the classes and functions in the module is instead added to the API reference section.
* Old tutorial file is removed
* Renames howto to User Guide


qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
 # that they have been altered from the originals.
 
-"""Array with numpy compatibility and multiple backend support.
+"""
+=================================================
+Dispatch Module (:mod:`qiskit_dynamics.dispatch`)
+=================================================
+
+.. currentmodule:: qiskit_dynamics.dispatch
+
```
DanPuzzuoli, line 21: Should this first sentence just be:

"This module contains an :class:`Array` class that wraps N-dimensional array
objects from different libraries for use with Numpy functions through a common
interface."

Not sure how to interpret the rest.

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+these arrays will raise an exception.
+
+
+Attributes and Methods
+----------------------
+
+The :class:`Array` class exposes the same methods and attributes of the wrapped
+array class and adds two additional attributes
+
+* :attr`~Array.data` which returns the wrapped array object
```
DanPuzzuoli, line 86:
```suggestion
* :attr:`~Array.data` which returns the wrapped array object
```

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+* :attr`~Array.backend` which returns the backend string of the wrapped array.
```
DanPuzzuoli, line 87:
```suggestion
* :attr:`~Array.backend` which returns the backend string of the wrapped array.
```

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+
+.. jupyter-execute::
+
+    # Call numpy.ndarray method
+    a.reshape((2, 5))
+
+Array Initialization
+--------------------
+
+An `Array` object can be initialized as `Array(data)` from any ``data`` that is
```
DanPuzzuoli, line 101:
```suggestion
An ``Array`` object can be initialized as ``Array(data)`` from any ``data`` that is
```

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+
+1. An ``array`` object of one of the supported backends.
+2. An object that can be used to initialize the backend array.
+3. An object of a class that defines a `__qiskit_array__` method with the
```
DanPuzzuoli, line 105:
```suggestion
3. An object of a class that defines a ``__qiskit_array__`` method with the
```

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+  the array backend.
+* ``backend`` (``str``): The array backend to use. If not specified this
+  will be inferred from the input data type if it is a backend array
+  instance, otherwise it will be the :func:`default_backend`.
+
+
+Using Arrays with other Libraries
+=================================
+
+The :class`Array` class is intended to be used with array functions in
```
DanPuzzuoli, line 132:
```suggestion
The :class:`Array` class is intended to be used with array functions in
```

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+    import library
+
+    # Wrap library function f
+    f = dispatch.wrap(library.f)
+
+    # Apply f to an Array
+    a = Array([0, 1, 2, 4])
+    b = f(a)
+
+The wrapped function will automatically unwrapping any Array args and kwargs
```
DanPuzzuoli, line 155:
```suggestion
The wrapped function will automatically unwrapping any ``Array`` args and kwargs
```

Also this sentence needs to be edited - should it just be "unwrap"?

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+for calling the wrapped function. If the result of the function is a backend array
+or an instance of a class that defines a `__qiskit_array__` method the wrapped
```
DanPuzzuoli, line 157:
```suggestion
or an instance of a class that defines a ``__qiskit_array__`` method the wrapped
```

qiskit_dynamics/dispatch/__init__.py
```diff
@@ -10,10 +12,189 @@
...
+function will return this as an Array object. Similarly if the result is a tuple
+this conversion will be applied to each element of the tuple where appropriate.
+
+Wrapping Decorators
+-------------------
+
+The :func:`wrap` function can also be used to wrap function decorators by setting
+the ``decorator=True`` kwarg.
+
+For example to wrap autograd and jit decorators from the Kax library to work with
```
DanPuzzuoli, line 167:
```suggestion
For example to wrap autograd and jit decorators from the JAX library to work with
```

We should try to consistently refer to JAX using all caps, as that is how the JAX team writes it in their own docs.

DanPuzzuoli (changes requested):
Aside from minor comments/edits looks great.

All of the other API docs put the list of functions/classes at the end, rather than at the beginning. Can you move this to the end (or argue why we should move all the others to the beginning)?

DanPuzzuoli (approved):

PR #34, opened 2021-09-24
Fix jax update errors

DanPuzzuoli:
### Summary

Fixing testing errors as a result of JAX update.

Current changes:
- Added `Array` wrapping to args of `assertAllClose` in the base test class. `jax.numpy` will now raise errors if things like lists are passed into most functions, so this wrapping is now necessary.
- Removed automatic array wrapping for `t_span` in `solve_lmde` and `solve_ode`. This fixed tests in `test_jax_transformations` which utilize these function. JAX is I think now more aggressive with turning things into tracers, and in these tests we were expecting these to be static. This wrapping was totally superfluous anyway as each solver handled things on its own.
    - This fixed some errors but broke other things - turns out it was due to some underlying solver functions assuming `t_span` was an `Array`, so they did `t_span.data` automatically. Removed those and those errors are fixed.
- There is a problem with the `dispatch.wrap` - not sure what it is. I was able to fix some tests by modifying the `jax_odeint` function so that everything passed to the jax solver is either a JAX array (not an `Array`) or only returns jax arrays.
     - Edit: This statement is wrong but keeping it here because we may want to undo these changes later.

### DIFF
--- a/docs/tutorials/jax_backend_demo.ipynb
+++ b/docs/tutorials/jax_backend_demo.ipynb
@@ -3,6 +3,8 @@
   {
    "cell_type": "code",
    "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
    "source": [
     "from matplotlib import pyplot as plt\n",
     "import numpy as np\n",
@@ -10,22 +12,22 @@
     "from qiskit_dynamics import solve_lmde\n",
     "from qiskit_dynamics.signals import Signal\n",
     "from qiskit_dynamics.models import HamiltonianModel"
-   ],
-   "outputs": [],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
    "source": [
     "# In this demo\n",
     "\n",
     "We show how `qiskit_dynamics` can be used with the `jax` backend to perform transformations like just-in-time compiling and gradient computation."
-   ],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "code",
    "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
    "source": [
     "from qiskit_dynamics import dispatch\n",
     "from qiskit_dynamics.dispatch import Array\n",
@@ -42,13 +44,16 @@
     "def gaussian(amp, sig, t0, t):\n",
     "    # Note: to enforce using the jax backend for the computation of the gaussian, wrap the\n",
     "    # input to the numpy function in an Array\n",
-    "    return amp * np.exp( -Array((t - t0)**2 / (2 * sig**2)))"
-   ],
-   "outputs": [],
-   "metadata": {}
+    "    amp = Array(amp)\n",
+    "    sig = Array(sig)\n",
+    "    t0 = Array(t0)\n",
+    "    t = Array(t)\n",
+    "    return amp * np.exp( -(t - t0)**2 / (2 * sig**2))"
+   ]
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
    "source": [
     "# 1. `jit`: Just-in-time compiling a parameterized simulation\n",
     "\n",
@@ -57,12 +62,21 @@
     "- Simulating a system over a range of model parameters\n",
     "\n",
     "One way to do this is the following. First, construct a model with just operators."
-   ],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "code",
    "execution_count": 3,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)\n"
+     ]
+    }
+   ],
    "source": [
     "#####################\n",
     "# construct operators\n",
@@ -81,20 +95,11 @@
     "#####################################\n",
     "\n",
     "hamiltonian = HamiltonianModel(operators=operators)"
-   ],
-   "outputs": [
-    {
-     "output_type": "stream",
-     "name": "stderr",
-     "text": [
-      "WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)\n"
-     ]
-    }
-   ],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
    "source": [
     "Next, define a function which takes the simulation input parameters and outputs the results. In this case, we parameterize a gaussian drive pulse, set it into the model, then solve and return the results.\n",
     "\n",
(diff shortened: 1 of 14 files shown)


DanPuzzuoli:
Just fixed some tests that were failing due to the following:
```
a + b
```
will raise an error if `a` is a JAX `DeviceArray` and `b` is an `Array`, as this gets passed directly to a binary operation function in JAX which will check the types. 

On the other hand:
```
b + a
```
will work because in the above `b` will control how addition is handled, and the `Array` wrapping is undone before being passed to JAX.

DanPuzzuoli:
Another test is currently failing due to the previous comment, but where raw `DeviceArray` is appearing because, for an `Array` `c`:

```
a, b = np.linalg(c)
```

`a` and `b` will be `DeviceArray`s instead of `Array`s. Is this expected behaviour?

DanPuzzuoli:
Another thing that I _think_ is strange is: I have a function `f` that takes in and returns `Array`. If I do `jit_wrap = wrap(jit, decorator=True)`, `jit_wrap(f)` doesn't work, but if I do `jit_wrap(wrap(f))` it works.

DanPuzzuoli:
Related to the binary operations issues:

It seems that a large number of the errors were due to a few lines in `RotatingFrame` of the form:
```
np.exp(t * self.frame_diag)
```

In the above we don't wrap `t` to avoid `Array` overhead in the numpy case, in which `self.frame_diag` will be a standard `np.ndarray`. However, in the jax case, `self.frame_diag` is an `Array`, and during jax tracing `t` will be some tracer. The above multiplication fails in this case because the tracer controls the binary operation. Changing the above line to:
```
np.exp(self.frame_diag * t)
```
fixes the error. 

I can't decide if this is very sketchy or not.

DanPuzzuoli:
Fixed errors in signal algebraic operations - they were also of the form `a + b` for `a` being a jax array type and `b` being an `Array`.

DanPuzzuoli:
Just "fixed" the final error:
- The JAX update made it so that calls to `rotating_wave_approximation` cannot be included in JAX-transformable functions.
- This in itself is actually fine: the outputs of `rotating_wave_approximation` can still be used in JAX-transformable functions, and this is by far the most important thing with with this function. (It was simply a nice "bonus" initially when `rotating_wave_approximation` was being written that this could be transformed.)
- To "fix" this error, I just rewrote the erroring test to verify that the outputs of `rotating_wave_approximation` can be used in transformable functions, rather than `rotating_wave_approximation` itself.
- I've added a corresponding note in the documentation for `rotating_wave_approximation` indicating this limitation. 

Still to do:
- Add a similar note to `Solver` - depending on the arguments to the init of `Solver`, `rotating_wave_approximation` may be used. The above means that a `Solver` cannot be constructed within a function that a user wants to JAX-transform _if_ they are also specifying a cutoff frequency.

While the above isn't particularly problematic (it's a well-defined limitation that we can document) it does indicate that new subtleties have been introduced for `dispatch`/`Array` when it comes to controlling "static" arrays within JAX transformations. JAX is now _always_ turning `jax.numpy.array` types into tracers when doing transformations, even if those  arrays are the result of "static" computations (i.e. they haven't yet come into contact with anything being influenced by the inputs to the function). Hence, it means we should be even more careful with when we wrap things in an `Array`. 

For example: 
- If `dispatch.default_backend() == 'jax'`, within `jit`, the line `Array(np.array([1., 2.]))` will turn the internal array into a JAX array, and therefore it will automatically become a JAX-tracer on the spot, even though it is completely "concrete".
- I don't think this used to be the case: the error in the `rotating_wave_approximation` jit test was due to a tracer appearing in a place where a concrete type was expected based on the fact that the variable is "concrete" in the above sense.

I'm not sure what the right lesson to take from this is in terms of usage and behaviour of `dispatch`. This is somewhat tricky, but at least its unambiguous.

DanPuzzuoli:
Okay this PR is ready for review.

It seems like all sources of errors were due to the one of the following (which are indicated in the JAX release notes):
- `jax.numpy` functions are now much more strict about input types. Seems like they must be either JAX array types or numpy arrays. E.g. lists are no longer valid, but more annoyingly `Array` is no longer valid. One realization of this is in the form of binary operations:
    - `a + b` will raise an error if `a` is a jax array-type and `b` is an `Array`, however
    - `b + a` will _not_ raise an error, as in this case `b.__add__` is used (the `Array` version).
- When `jax.jit` or other transformations are being used, JAX will turn absolutely everything JAX touches into tracers. I.e. while jitting, the outputs of all `jax.numpy` functions (including `jax.numpy.array`) will be tracers, even if all of the inputs are "static" with respect to the transformation (i.e. it isn't actually affected by one of the arguments of the function being transformed). To be honest I'm not 100% sure if this was already the case and I had just misinterpreted some aspects of JAX tracing, but in any case something of this form was causing errors.
    - For example, there were some parts of the code in which essentially this was happening: `Array(Array(x), backend='numpy')` (not this line literally but, `x` is turned into an `Array` and at some point later, into an `Array` that is explicitly numpy-backed). If the default backend was JAX, the first `Array` call would turn `x` into jax array, and then the outer `Array` call would turn it into a numpy array. 
    - The tests involving jitting this code now raises the error that `nump.array` is being called on a tracer, but it didn't use to. In the tests, `x` was "static", in the sense that at the point in which the above occurred, `x` didn't come into contact with any of the arguments of the function being transformed. Hence, I _believe_ in the previous JAX version, it was no problem to call `numpy.array` on `x` after it was explicitly turned into a `jax.numpy.array`, as it didn't get turned into a tracer. However, now that everything does, this raises an error.

Changes to fix the errors resulting from above:
- There are a ton of small changes to fix errors of the first type. A lot of it is just adding or removing `Array` wrapping in various places, or even changing the order of things in certain binary operations.
- For errors resulting from the second change:
    - Removed some unnecessary early wrapping on `t_span` in `solve_ode` and `solve_lmde`. If `t_eval` is set, then `t_span` can't be compiled over and needs to be turned into a `numpy.array` (this is something we could in principle change but for now, the way our special time-argument-merging code is written is not JAX-compatible). Originally, `solve_ode` and `solve_lmde` automatically wrapped `t_span` in an `Array`, but if JAX is the default backend this was causing problems later `t_span` was converted back to a `numpy.array`.
         - In terms of functionality of dynamics this changes nothing, as the inability to compile over this arrangement of parameters was known/expected, and "intended" in the sense that an explicit choice was made to write the code in this way.
     - `rotating_wave_approximation` can no longer be compiled over. This is a "deeper error" than the previous one as it has to do with how the `Signal` classes use `Array` on their frequencies/phases, which depends a lot on my assumptions about how JAX/dispatch works, which have been made somewhat invalid by the recent changes.
          - In the end this isn't really a problem: I wasn't even expecting `rotating_wave_approximation` to be compilable when it was first written, and it was just a nice bonus that it was. This ultimately doesn't affect any "intended" modes of use of anything.
          - I've added documentation to `rotating_wave_approximation` to indicate that it can't be compiled over, however its outputs _can_ (which is the most important thing). I.e. you can transform a model, then use the model in jax-transformable functions (including updating signals), but you just can't do the whole RWA within a JAX-transformable function.
          - I've also added documentation to `Solver` to indicate that it can't be constructed within a JAX-transformable function if `rwa_cutoff_freq` is specified, but that any instance can be used as normal within a JAX-transformable function.

chriseclectic (approved):

PR #35, opened 2021-09-27
Adding JAX to setup.py as optional install

DanPuzzuoli:
### Summary

As per suggestion/instructions by @mtreinish , added JAX as optional install for dynamics in `setup.py`, so 

```
pip install "qiskit-dynamics[jax]"
``` 

should work. (For now though just did `pip install ".[jax]"` from local repo directory to test.)


setup.py
```diff
@@ -22,6 +22,9 @@
     "qiskit-terra>=0.16.0",
 ]
 
+jax_extras = ['jax>=0.2.21',
```
chriseclectic, line 25: Do we actually need the latest version of jax, or do the changes you made to fix the module for the latest version of jax still work with older versions?

DanPuzzuoli: I don't actually know - I believe CI is currently just installing the latest version though, so if we want to list older versions we should add them to the tests. What are your thoughts? I feel like there isn't much need at this stage, especially since it's the first release.

chriseclectic (approved):

PR #36, opened 2021-09-27
Update wrap function to avoid conditional checks

chriseclectic:
### Summary

This updates the `dispatch.wrap` function to avoid conditional checks within the function defs of the wrapped functions it returns when possible. It also adds an extra kwarg to `wrap` that allows disabling the recursive wrapping of function arguments (which is enabled by default). Disabling this may also give performance improvements for compatible functions.

### Details and comments

Originally the hope was to return wrapped functions as a class instance with a call method that could be inspected to return the original function, and details about the wrapped, however this caused several tests to fail that used jax jit and grad functions on model evaluation. 

I'm not sure how to fix this so just added in the refactor code for the wrapped functions to move the conditional checks outside the function definitions instead. This increases the verbosity of the source code, but should help improve performance if these wrapped functions are ever called in a performance intensive loop since the functions themselves do less conditional checking.

### DIFF
--- a/qiskit_dynamics/dispatch/wrap.py
+++ b/qiskit_dynamics/dispatch/wrap.py
@@ -18,70 +18,6 @@ from typing import Callable
 from .array import Array
 
 
-class WrappedFunction:
-    """Class for storing wrapped array backend functions to work with Arrays."""
-
-    def __init__(
-        self,
-        func: Callable,
-        wrap_return: bool = True,
-        wrap_args: bool = True,
-        decorator: bool = False,
-    ):
-        """Wrap an array backend function to work with Arrays.
-
-        Args:
-            func: a function to wrap.
-            wrap_return: If ``True`` convert results that are registered array
-                        backend types into Array objects (Default: True).
-            wrap_args: If ``True`` also wrap function type args and kwargs of the
-                   wrapped function.
-            decorator: If ``True`` the wrapped decorator function ``func`` will
-                    also wrap the decorated functions (Default: False).
-        """
-        self._func = func
-        self._wrap_return = wrap_return
-        self._wrap_args = wrap_args
-        self._decorator = decorator
-
-        if decorator:
-            self._wrapped_func = _wrap_decorator(func, wrap_return=wrap_return, wrap_args=wrap_args)
-        else:
-            self._wrapped_func = _wrap_function(func, wrap_return=wrap_return, wrap_args=wrap_args)
-
-    def __call__(self, *args, **kwargs):
-        """Evaluate the wrapped function."""
-        return self._wrapped_func(*args, **kwargs)
-
-    def __repr__(self):
-        ret = "WrappedFunction"
-        ret += f"\n decorator: {self._decorator}"
-        ret += f"\n wrap_return: {self._wrap_return}"
-        ret += f"\n wrap_args: {self._wrap_args}"
-        ret += f"\n func: {repr(self._func)}"
-        return ret
-
-    @property
-    def func(self):
-        """The original function."""
-        return self._func
-
-    @property
-    def decorator(self):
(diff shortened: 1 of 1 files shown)


qiskit_dynamics/dispatch/wrap.py
```diff
@@ -125,3 +69,150 @@ def wrapped_function(*args, **kwargs):
         return result
 
     return wrapped_function
+
+
+def _wrap_args(args):
+    """Return wrapped args"""
+    return tuple(_wrap_array_function(x) if isinstance(x, FunctionType) else x for x in args)
```
DanPuzzuoli, line 76: Should this do: `... else Array._wrap(x) ...`?

qiskit_dynamics/dispatch/wrap.py
```diff
@@ -125,3 +69,150 @@ def wrapped_function(*args, **kwargs):
...
+
+
+def _wrap_kwargs(kwargs):
+    """Return wrapped kwargs"""
+    return dict(
+        (key, _wrap_array_function(val)) if isinstance(val, FunctionType) else (key, val)
```
DanPuzzuoli, line 82: similar to above, `... else (key, Array._wrap(val)) ...`?

DanPuzzuoli (commented):
I've commented in two places about something I'm not sure about:

Should `_wrap_args` and `_wrap_kwargs` not be doing some `Array` wrapping of the arguments that are registered array types if the arg/kwarg isn't a `FunctionType`?

Maybe not something to worry about for now, but i'm wondering if you think we should have this stuff handle "pytrees" - not sure how common this terminology is but the JAX documentations uses this term to refer to nested python containers (lists, tuples, dicts) whose base data is JAX arrays. `jit`, `grad`, etc... all work on functions whose arguments are pytrees, rather than just functions whose arguments are arrays. I guess this would require doing some recursion on the args/results.

qiskit_dynamics/dispatch/wrap.py
DanPuzzuoli: I guess I have this backwards, it's supposed to remove `Array` wrapping here right?

DanPuzzuoli: Ah sorry I realize now how the `__qiskit_array__` method is being used now at the beginning after reading the docs update. Disregard this comment!

qiskit_dynamics/dispatch/wrap.py
DanPuzzuoli: Disregard this for the same reason as above.

qiskit_dynamics/dispatch/wrap.py
```diff
@@ -18,86 +18,30 @@
 from .array import Array
 
 
-def wrap(func: Callable, wrap_return: bool = True, decorator: bool = False) -> Callable:
+def wrap(
+    func: Callable, wrap_return: bool = True, wrap_args: bool = True, decorator: bool = False
+) -> Callable:
     """Wrap an array backend function to work with Arrays.
```
DanPuzzuoli, line 24: Can you expand on this to more explicitly say how it sets up the function to "work with arrays"? I.e. add a sentence or two explaining that it unwraps `Array`-type arguments (or anything with `__qiskit_array__`) and wraps the results.

DanPuzzuoli (changes requested):
Disregard my previous review - I was getting a bit mixed up by some stuff that was cleared up when I read your dispatch docs PR.

Aside from the doc and my following question looks good:

Can you explain the intended difference between `decorator` and `wrap_args` arguments (or how they interact)? It seems they are both for recursive wrapping of arguments that are functions. Could I equally well use `wrap(jit, wrap_args=True)` and `wrap(jit, decorator=True)`?

PR #49, opened 2021-10-22
Add RK4 fixed step solver and parallelized JAX fixed step solvers

DanPuzzuoli:
### Summary

The goal of this PR is to:
- Add jax and numpy-based RK4 fixed step solver
- Add a new "fixed step LMDE solver template" for JAX that uses parallel operations, i.e. all time-steps are integrated simultaneously in parallel using `vmap`.
    - This is specific to LMDEs as it assumes that structure for the parallelization to work.
    - This method works by simulating the full propagator over each time-step in parallel, then multiplying them all together (also in parallel, using `jax.lax.associative_scan`). A consequence of this is that, if the user wants to simulate only a single state, this approach will actually be more expensive than serial computations (in terms of linear algebraic operations), and may not be worth the parallelization. However, if a user wants to simulate the full propagator itself, then no additional computational cost is created.
- Using the above, add jax parallel versions of both RK4 and expm solvers.

### Details and comments

Currently implemented the jax parallel template, the jax parallel expm solver, and have added tests for the parallel expm solver derived directly from the already existing tests for expm solvers through inheritance. The template code currently assumes that `y0` is square, so it needs to be expanded, but the tests are passing so this is a good start!

To do:
- ~Generalize the parallel template to accept non-square `y0`.~
    - ~Test!~
- ~Add RK4 solvers~
    - ~Test! (Can probably abstract out content of expm tests.)~
- Add proper interface access to new solvers through `solve_lmde` and `solve_ode`.
    - Add some higher level tests for these.
    - Update documentation.
- Check if update necessary in module docs.


qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -84,8 +95,15 @@ def solve_ode(
     - ``scipy.integrate.solve_ivp`` - supports methods
       ``['RK45', 'RK23', 'BDF', 'DOP853', 'Radau', 'LSODA']`` or by passing a valid
       ``scipy`` :class:`OdeSolver` instance.
-    - ``jax.experimental.ode.odeint`` - accessed via passing
-      ``method='jax_odeint'``.
+    - ``'RK4'``: A fixed-step 4th order Runge-Kutta solver.
```
chriseclectic, line 98: What is the difference between RK4 and RK45? Is the scipy RK45 not fixed step?
I wonder if we should use some naming convention so you can tell which methods are fixed step and which aren't.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -84,8 +95,15 @@ def solve_ode(
...
+      Requires additional kwarg ``max_dt``, indicating the maximum step
+      size to take. This solver will break integration periods into even
+      sub-intervals no larger than ``max_dt``, and step over each sub-interval
+      using the standard 4th order Runge-Kutta integration rule.
+    - ``'jax_RK4'``: JAX-version of ``'RK4'``, implemented using JAX
```
chriseclectic, line 103: "JAX looping logic" is pretty vague for a method description. Probably either just something like "JAX backend implementation of RK4 method." or could specifically say if it automatically uses jit/autograd in the method or not.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -174,14 +196,22 @@ def solve_lmde(
...
+    - ``'scipy_expm'``: A fixed-step matrix-exponential solver using ``scipy.linalg.expm``.
+      Requires additional kwarg ``max_dt`` indicating the maximum step
       size to take. This solver will break integration periods into even
-      sub-intervals no larger than ``max_dt``, and solve over each sub-intervals via
+      sub-intervals no larger than ``max_dt``, and solve over each sub-interval via
       matrix exponentiation of the generator sampled at the midpoint.
     - ``'jax_expm'``: JAX-implemented version of ``'scipy_expm'``, with the same arguments and
       logic.
-
+    - ``'jax_expm_parallel'``: Same as ``'jax_expm'``, however all loops are implemented using
```
chriseclectic, line 206: Is there any reason to have a separate method for these, rather than just making "parallel" a kwarg option for these jax methods that you can set to True or False?

chriseclectic (commented):
I can't remember if we looked into this in the past, but I'm wondering if the scipy fixed step solvers should be implemented as custom `OdeSolver` for `solve_ivp` or if that didn't work or was too difficult.

qiskit_dynamics/solvers/solver_functions.py
DanPuzzuoli: I think this is a good idea. One slightly awkward thing though is that `jax_RK4_parallel` is an LMDE method, whereas `jax_RK4` is an ODE method, in the sense that, even though RK4 is an ODE method, the approach to parallelization is based on LMDE structure.

This will require a change in how `solve_lmde` delegates to `solve_ode`; right now it assumes that the list of methods available in each has no overlap, and now we'll need to do some checking based on additional kwargs. I think this is fine, but will just increase the complexity of that step a little bit.

qiskit_dynamics/solvers/solver_functions.py
DanPuzzuoli: Changed it to your suggestion. We don't jit anything automatically in dynamics so I won't say anything about this.

qiskit_dynamics/solvers/solver_functions.py
DanPuzzuoli: Yeah RK4 is a fixed step method, but RK45 is a variable step method that I believe uses 4th order runge kutta to take steps, but also computes 5th order runge kutta and compares to 4th order as part of it's routine for adjusting step size.

Turns out it's possible to actually "hack" the scipy RK45 with kwargs to behave like RK4, by just setting the tolerances insanely large and setting the `max_step` to whatever step size you want (described in comments [here](https://stackoverflow.com/questions/54494770/how-to-set-fixed-step-size-with-scipy-integrate)). I initially thought we could maybe drop the `RK4` method because of this, though `solve_ivp` will still be performing error checkign computations, which serve no purpose in this case, so maybe it's still worth keeping it.

DanPuzzuoli:
Before proceeding with this PR we need to do some speed testing of the parallelized JAX solvers on both CPU and GPU.

I've just tried some test cases with CPU on my laptop, and what I'm seeing is:
1. `jax_expm` is consistently faster than `jax_expm_parallel`. Not by too much, e.g. 10s v.s. 13s for a large system with a ton of time-steps, but it does seem to be consistent (even when varying system size and number of time steps). Watching the CPU usage, it seems like `jax_expm` keeps usage consistently high, whereas `jax_expm_parallel` has periods during which the CPU usage drops. 
2. `jax_rk4_parallel` is insanely slower than `jax_rk4`.

I've even tried writing some smaller toy functions to strip out all of the other stuff going on in dynamics, and the above behaviour is the same.

@brosand  is going to run the same tests on a GPU, to see if this is an XLA CPU v.s. GPU utilization efficiency issue, or if the usage of `vmap` and `associative_scan` (as opposed to serially doing everything with `scan`) is really worth it or not.

brosand:
Following up with the results of the GPU testing:

For `jax_expm` we do observe a speedup with `jax_expm_parallel`, when run on a GPU, specifically a `v100`. With dimension 70, we see jitted times of 2 s for the serial function and 130-150 ms for the parallel solver. 

For the `rk4` solver, on the same gpus with d=40, we see a 10 second run time for the jitted serial solver, and then 700ms run time for the jitted parallel solver.

It seems like both the parallel and serial solvers are highly accelerated by the gpus. However the parallel solver becomes even faster than the serial solver once it has access to gpus.

qiskit_dynamics/solvers/solver_functions.py
DanPuzzuoli: I'm now of the mind that this should remain as is - just tried changing it and due to RK4 being an ODE method, but the parallel version is an LMDE method, it makes the logic to check when `solve_lmde` delegates to `solve_ode` very weird. (E.g. we will need to add `jax_RK4` to the list of both `solve_ode` and `solve_lmde` methods.) The current way feels a little less elegant but a lot simpler/obvious.

DanPuzzuoli:
Based on the CPU v.s. GPU speed issues with these solvers, added a sentence to the description of each parallel solver that we recommend only using it with GPU.

DanPuzzuoli:
One more thing to add to this PR, as suggested by @brosand : 

While the addition to the doc string that the parallel solvers are only intended for GPU is good, it doesn't seem like quite enough to discourage their use on CPU. The `jax.default_backend()` command can be used to determine if JAX is going to cpu or gpu to execute, so we can add a line in the parallel solver template that raises a warning if `jax.default_backend() == 'cpu'`.

chriseclectic (approved):
LGTM

PR #51, opened 2021-11-02
Introduce JAX sparse evaluation mode

### LINKED ISSUE #40: Sparse evaluation mode for GeneratorModel raises error if default array backend is JAX

The following raises an error (note that this code depends on syntax from #30):
```
dispatch.set_default_backend('jax')
static_op = np.array([[1, 0.], [0., -1.]])
operators = [np.array([[0., 1.], [1., 0.]])]

ham_model = HamiltonianModel(static_operator=static_op, 
                             operators=operators,
                             signals=[1.],
                             evaluation_mode='sparse')
ham_model(1.)
```

This is not a particularly major or urgent bug: sparse operations aren't even supported in JAX anyway (in this package - and in JAX itself there are just experimental implementations of sparse matrices), so nothing is actually being blocked by this. 

Nevertheless: I didn't expect this to result in an error. The error is due to the signals being evaluated as an `Array` if the default backend is JAX. Evaluation of sparse models uses object arrays and `np.tensordot`, and the `Array(..., backend='jax')` for the coefficient array causes this to get dispatched to JAX (which does not supported object arrays and therefore raises an error).

### Possible solutions

It may be natural to just raise an error if a sparse mode is set while the default backend is JAX. This makes it apparent that there is no JAX sparse support in this package.

DanPuzzuoli:
This PR is ready for review. See latest comment for up to date summary.

### Summary

Closes #40 (incidentally)

JAX now has experimental support for a sparse array type called `BCOO`. While the features are currently limited, there is enough to implement a jax sparse model evaluation mode.

This PR will add `BCOO`-based evaluation modes to all model classes. This involves:
- Adding `BCOO`-based versions of the relevant operator collections.
- Adding handling for this in the model classes.
- Adding a `to_BCOO` to `type_utils`, similar to `to_array` and `to_csr`. Also, add `BCOO` as a type that "gets a pass" in `to_numeric_matrix_type`.
- Adding handling for BCOO in hamiltonian validation.
- Modifying some core functions in `RotatingFrame` that have special handling based on type (currently just `array` v.s. `csr_matrix`, but will need to now handle `BCOO`).
- Figuring out how to incorporate this into `rotating_wave_approximation`. 

### Details and comments

Some questions/comments:
- Should we have the jax sparse evaluation mode triggered by setting `model.evaluation_mode = 'sparse'` when `dispatch.default_backend() == 'jax'`, or should it have an alternate name like `model.evaluation_mode = 'jax_sparse'`? I think the former option is a good idea given that we want to avoid mixing JAX and numpy/scipy types. This would also automatically solve #40 (problems when the default backend is `jax` but the csr sparse mode is chosen).
- Due to limited functionality, the method for computing a linear combination of sparse matrices is a bit awkward. Something better will be available with the next release of JAX, so need to keep an eye out for this.
- Related to the above: I'm not sure when BCOO arrays were introduced, but this PR is likely to lead to a large increase in the minimum jax version number.
- Jax expm doesn't work on BCOO, so maybe we should detect this and raise an error? This should be documented as well


DanPuzzuoli:
Current status:
- Made `JAXSparseOperatorCollection`
   - Added test cases
- Updated some `RotatingFrame` functions to work directly with the `BCOO` type
   - No tests yet
- Added `to_BCOO` to `type_utils` and updated `to_numeric_matrix_type` to leave `BCOO` alone
   - No tests yet
- Added temporary `evaluation_mode='jax_sparse'` to `GeneratorModel` to set up/work with the `JAXSparseOperatorCollection`. Seems to be working, though need to add proper testing.

DanPuzzuoli:
(Hopefully) Final touches:

- Updated evaluation mode documentation in model classes. Noted that reverse mode autodiff doesn't currently work for sparse jax mode.

DanPuzzuoli:
This PR is ready for review.

## Summary

It adds jax-based sparse evaluation for all models, including a `'sparse_vectorized'` for `LindbladModel`. This is implemented using `jax.experimental.BCOO` arrays. Some high level comments:
- Accessing the jax-based sparse implementations requires setting the default backend to `'jax'` and then setting evaluation mode to `'sparse'` or `'sparse_vectorized'` as before. (If the default backend is `'jax'` it'll set up the jax sparse stuff, and otherwise it'll set up the `csr_matrix` sparse types.) This is now explained in the `evaluation_mode` property doc strings.
- `evaluation_mode='sparse'` for `LindbladModel` requires some operations on `BCOO` arrays that are not currently reverse-mode differentiable.  I was told in this [discussion](https://github.com/google/jax/discussions/8454) that this is not a fundamental limitation, it just hasn't been implemented yet, and in the meantime it is forward mode differentiable. I've explained this in the `LindbladModel.evaluation_mode` doc string.
- Related to the above, generally, the implementation of a `BCOO`-based sparse evaluation modes required some work arounds that will need to be "fixed" as jax sparse functionality advances. Based on some speed testing it seems to be well worth it anyway, hence this PR.
- The jax/jaxlib minimum versions needed to be bumped up to be very recent.
- As the `jax.scipy.linalg.expm` doesn't work on BCOO arrays (or `csr_matrix` inputs for that matter), I've set `solve_lmde` to raise an error if `method='jax_expm'` is selected for a model in sparse mode.

## Details

- Added operator collections implemented with `jax.experimental.sparse.BCOO` arrays:
    - `JAXSparseOperatorCollection`, `JAXSparseLindbladCollection`, and `JAXSparseVectorizedLindbladCollection`
    - Refactored operator collections tests to make it easier to add tests for new operator collections, and added test classes for the above. Currently `JAXSparseLindbladCollection` overrides the autodiff test of the class it inherits from to test forward-mode autodiff only.
- Added jax-based sparse evaluation mode to `GeneratorModel`/`HamiltonianModel` and to `LindbladModel`. Enabled by passing `evaluation_mode='sparse'` when `dispatch.default_backend() == 'jax'`.
    - Refactored tests for all of these classes to make it easier to add tests for new modes, and added tests for the all jax-based sparse modes.
    - Updated `evaluation_mode` doc string to describe this new option and how to access it.
- Added a `to_BCOO` function to `type_utils`, which has the same formatting as `to_array` and `to_csr` and corresponding tests. Also updated `to_numeric_matrix_type` to pass `BCOO` arrays through.
    - For now it just returns the input if already of type `BCOO`, or calls `BCOO.fromdense(to_array(x).data)`. Not sure how else to instantiate this other than from a dense jax array.
-  Updated several `RotatingFrame` functions to work directly on `BCOO` arrays. These are mainly used when switching frames in a model class.
    - Added test cases.
- Added a test case for `rotating_wave_approximation` for a `GeneratorModel` in jax sparse evaluation mode.
    - This function just converts to dense arrays and does the computation on them, then converts back to whatever the input was - so no need to add functionality here.
- Added error raising to `solve_lmde` if `method='jax_expm'` is selected for a model in sparse mode.
    - Added test case for this.
- Bumped jax/jaxlib versions in setup.py to `jax>=0.2.19` and `jaxlib>=0.1.70` as they appear to be the minimum versions compatible with all sparse operations we use.

DanPuzzuoli:
As a speed gain example to using sparse: on my local CPU, simulating two coupled transmons, each with dimension 30 (so total dimension 900), compiled time to simulate a particular CR sequence with dense takes about 250s, while for sparse it takes about 12s. (With the former in the frame of the whole drift, and the latter in the frame of the diagonal of the drift.) This of course is an incredibly sparse system, so we expect a huge benefit to sparse arrays in this case, and it's nice to see that realized here.

brosand:
Initial testing on GPUs is showing that jax sparse evaluation is much slower on gpus than dense evaluation, and even slower than sparse evaluation on cpus. In some cases this slow down is a factor of 50x. So far I have only examined small dimensions, d<100. We are currently testing higher dimensionality on GPUs to see if the performance improves. This may be because while XLA is optimized well on GPUs, the sparse format is not setup for the same optimization.

DanPuzzuoli:
One more thing that needs to be done with this PR is to update the JAX userguide entry, which describes how there is no JAX sparse support. Could either completely remove this, or update it to say that sparse support is only recommended for use on CPUs.

DanPuzzuoli:
Some updates to the PR:

- Rebased on main after a recent merge.
- Added text to the `evaluation_mode` property indicating that JAX sparse mode is only recommended to be used on CPU.
- Added a warning to both `GeneratorModel` and `LindbladModel` (and `HamiltonianModel` via inheritance) when JAX sparse evaluation mode is set: If `jax.default_backend()` is not CPU, it will warn that it is only recommended for use on CPU.
- Updated the "how to use jax" userguide:
    - At the end it had a section on how sparse evaluation is not supported with JAX. I've deleted this section and replaced it with a section listing JAX features that are only recommended on one of CPU or GPU.

brosand (commented):
Left a number of comments, nothing that I think needs to be changed, mostly suggestions. Will be discussed, along with some of the longer scripts which I want to talk through.

qiskit_dynamics/models/operator_collections.py
```diff
@@ -247,6 +268,61 @@ def evaluate_rhs(self, signal_values: Union[Array, None], y: Array) -> Array:
...
+                            operators cannot be evaluated."""
+        )
+
+    def evaluate_rhs(self, signal_values: Union[Array, None], y: Array) -> Array:
+        if y.ndim < 3:
+            if isinstance(y, Array):
+                y = y.data
+            return Array(jsparse_matmul(self.evaluate(signal_values), y))
+
+        raise QiskitError(self.__class__.__name__ + """  cannot evaluate RHS for y.ndim > 3.""")
```
brosand, line 322:
```suggestion
        raise QiskitError(self.__class__.__name__ + """  cannot evaluate RHS for y.ndim >= 3.""")
```

brosand (approved):
LGTM

PR #58, opened 2021-11-25
Simulation customization userguide (rotating frames, RWA, sparse)

### LINKED ISSUE #57: Add some documentation entry for jax sparse evaluation mode
PR #51 introduces JAX sparse evaluation mode. As suggested by @chriseclectic , it would be nice to have some additional documentation for this, e.g. a userguide or tutorial.

I think probably a userguide entry is the right format, something answering the question(s): "When and how should I use sparse representations?", and in particular how to choose the simulation frame when simulating a sparse model.

One thing to consider is whether or not we should gear this tutorial towards JAX sparse explicitly. I think yes as the scipy sparse mode still needs to be performance optimized, though the lessons/interface points hold true in this case as well.

DanPuzzuoli:
### Summary

Closes #57 
Depends on #51 

Introduces a new userguide entry centered around "customizing" simulations via model transformations and evaluation modes. Phrases everything in terms of "how-to", e.g. "how-to use rotating frames, and their benefits", and demonstrates in a simple example the impact of making these choices. Also shows how to use sparse evaluation mode within JAX and emphasizes that when doing this, it is recommended to use diagonal frames only.


docs/userguide/how_to_configure_simulations.rst
```diff
@@ -0,0 +1,331 @@
+How-to customize simulations using model transformations and evaluation modes
+=============================================================================
+
+Qiskit Dynamics provides model transformations and different numerical methods
+for evaluating models that can be used to customize how a simulation is done,
```
brosand, line 5: Line 5 is a bit confusing, when I think it through it makes sense. But if you could split it up a bit that might be nice

docs/userguide/how_to_configure_simulations.rst
```diff
@@ -0,0 +1,331 @@
...
+    # transform the results of the solver in the rotating frame out of the rotating frame
+    U_rf = rf_solver.model.rotating_frame.state_out_of_frame(T, rf_results.y[-1])
+
+    fidelity(U, U_rf)
+
+Based on the fidelity, we see that the two simulations are numerically equivalent
+with reasonable accuracy based on our specified tolerances.
+
+The discrepancy in solving times can be understood by examining the number of
+RHS evaluations when solving in each instance. The number of RHS evaluations for
```
brosand, line 131: Maybe define RHS here in case a new user isn't familiar with the terminology. Although maybe everyone who is using this package already has a strong understanding?

docs/userguide/how_to_configure_simulations.rst
```diff
@@ -0,0 +1,331 @@
...
+    jax.config.update('jax_platform_name', 'cpu')
+
+    # import dispatch and set default backend
+    from qiskit_dynamics import dispatch
+    dispatch.set_default_backend('jax')
+
+Reconstruct the model pieces at a much larger dimension, to observe the
+benefits of using sparse arrays. Furthermore, set up the initial state to
+be a single column vector, to
+further highlights the benefits of the sparse representation.
```
brosand, line 230:
```suggestion
further highlight the benefits of the sparse representation.
```

brosand (approved):
A couple small edits, @chriseclectic may want to take a closer look as well, I would be interested to see other notes to see what I should look at in the future.

docs/userguide/how_to_configure_simulations.rst
DanPuzzuoli: Agreed, just expanded this into what I think is a better paragraph.

docs/userguide/how_to_configure_simulations.rst
DanPuzzuoli: Changed this to: `right-hand side (RHS)`, and added: "when solving the differential equation". I think this is a good addition, and _should_ be sufficient, all the key words are there. Also yeah, if a user is at the point of wanting to configure simulations they should know what these words mean at this point anyway.

DanPuzzuoli:
@brosand I tried to address your comments by modifying the sections you suggested. 

I also needed to make a couple of other fixes:
- Switch to using the new `Array.set_default_backend` over the now-deprecated `dispatch.set_default_backend` in the userguide entry
- This change also still needed to be made in the `GeneratorModel` class, which was raising a further deprecation warning when setting the evaluation mode to `'sparse'`, which I've fixed now.
- Related to the previous point, I've added the word "Array" when referring to the default backend in descriptions of jax sparse evaluation mode.

docs/userguide/how_to_configure_simulations.rst
brosand: Agreed, great

brosand (approved):
LGTM

Issue #54, opened 2021-11-11
Add channel filtering/ordering to pulse Schedule -> Signal converter


DanPuzzuoli:
### What is the expected behavior?

Connected to #53.

Update the converter to include optional argument `channels`, specifying a list of channels as strings e.g. `['D0', 'D2', 'U0']`, which modifies the behaviour of the function to:
- Filter out channels not included in the list.
- Return signals in a list given by the ordering in `channels`.

DanPuzzuoli:
We also need to add test cases for this function.

PR #59, opened 2021-11-26
Added channel filtering to pulse converter

eggerdj:
### Summary

This PR solves #54 

### Details and comments

The channel filtering is done using an extra argument to the converter.


qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -65,14 +92,17 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
             qiskit.QiskitError: if not enough frequencies supplied
         """
 
-        if self._carriers and len(self._carriers) < len(schedule.channels):
-            raise QiskitError("Not enough carrier frequencies supplied.")
+        if self._channels is None:
+            if self._carriers and len(self._carriers) < len(schedule.channels):
+                raise QiskitError("Not enough carrier frequencies supplied.")
```
DanPuzzuoli, line 98: Can you add a comment here saying what data is being initialized - these dicts and the subsequent loop are nearly self explanatory but would be helpful to have a brief comment on how these dicts will be used.

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -65,14 +92,17 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
...
 
         signals, phases, frequency_shifts = {}, {}, {}
 
         for idx, chan in enumerate(schedule.channels):
```
DanPuzzuoli, line 101: If `self._channels is not None`, can we already start doing some "filtering" in these loops? E.g. only adding `chan.name` as a key to these `dict`s if `self._channels is not None` and `chan.name in self._channels` (similarly for the subsequent loop, if an instruction isn't associated to one of these channels, just skip to the next instruction.

I guess this isn't an option if there's some interdependency between channels/instructions? I know control channels have their frequency specified as linear combinations of frequencies of drive channels (unless this has changed since I last looked at it). If you set or shift frequency of a drive channel mid-schedule, does that impact the frequency of associated control channels?

DanPuzzuoli (changes requested):
It's looking generally good.

I think with the new addition of the `channels` initialization kwarg this class needs more documentation as to how it behaves. Specifically:
- The behaviour of the object depending on what combination of the `carriers` and `channels` kwargs are set. The behaviour of the object is non-trivial impacted by whether or not one is `None`, or both are specified.
- An explanation of the format required by the `channels` argument. I guess the strings are either `'dxx'`, `'uxx'`, or `'mxx'`, for `'xx'` being some integer?

To encourage consistency in documentation style, the detailed description of these initialization arguments and their impact on the behaviour of the class could be explained in the class doc string, similar to the `qiskit_dynamics.solvers.solver_classes.Solver` class. It gives a more detailed description of the initialization args and their impact on the class in the class doc string, and then is a bit more concise with the `__init__` doc string. Let me know what you think of this - I'm sure not all classes in dynamics follow this, but `Solver` is one of the more recent ones.

Similar to the above, for the `get_signals` method:
- Can you document what format of `schedule` or instructions it can handle? This request is partially a function of my ignorance of the details of `pulse` - maybe not much needs to be said here, but I noticed you're calling `block_to_schedule` in the tests. Is `get_signals`  assumed to work on the outputs of this function? What is it about the format output by this function that is required?

For testing:
- This `ddt` module looks super useful! Is the license for this okay? Is it being used in other qiskit packages?

One general thing, and this isn't a comment on this PR so much as a question related to the string parser that I'm in the process of adapting to dynamics:
- I notice that `channel.name` is lower case. The model strings that I've seen have historically used `'D0'` instead of `'d0'` to indicate `DriveChannel(0)`, and as I've been adapting the string parser I've documented that it should be specified as upper case `'D0'`. I think based on what I'm seeing in this PR we should actually prefer lower case - the string parser can of course still accept upper case _or_ lower case, but when coming up with canonical lists internally for pulse simulation it seems like lower case is the way to go. Thoughts? Is the lower case standard throughout `pulse`?

eggerdj:
Looking at this again I'm wondering if it would not make more sense to have `carriers` be a dict in the converter. This would make the code more robust.

qiskit_dynamics/pulse/pulse_to_signals.py
eggerdj: The changes here c37d941 simplify the interface quite a bit. This also reduces the documentation need.

eggerdj:
Replying to a few questions:
* ddt is used in other qiskit packages. See e.g. here: https://github.com/Qiskit/qiskit-terra/blob/main/requirements-dev.txt
* Regarding upper vs lower case: we should use lower case since this is how the pulse channel prefixes are defined. See here for example: https://github.com/Qiskit/qiskit-terra/blob/d662fbbc7cfc8a98181b28ba6d87b7f0b26d88f3/qiskit/pulse/channels.py#L168-L171

DanPuzzuoli (changes requested):
The interface/documentation looks good! My only remaining substantial comment is:

- If `self._channels is not None` can you pre-filter the schedule before line 100 via `schedule.filter(channels=self._channels)` (not sure what the exact input/output format of `filter` are) so that only the instructions on channels of relevance are parsed?

Minor formatting consistency comments:
- In the `__init__` and `get_signals` method docs make sure all sentences start with capital letters.
- Can you indent multiline `Args` descriptions so that all text appears after the colon, e.g.:
```
carriers: a dict of carrier frequencies. The keys are the names of the channels
          and the values are the corresponding carrier frequency.
```
v.s.
```
carriers: a dict of carrier frequencies. The keys are the names of the channels
    and the values are the corresponding carrier frequency.
```
Not sure what's generally standard here - and I was just looking and other parts of qiskit are formatted similarly to what you've written - but the majority of docs in dynamics use the former.

DanPuzzuoli (approved):
LGTM

PR #60, opened 2021-11-29
[WIP] Dispatch 2.0

chriseclectic:
### Summary

The PR rewrites the dispatch system to have better performance and be easier to use for registering new array backends and dispatch function rules, and for using it directly as a common interface when working with different array libraries directly.

The `Array` class has been refactored to use this new dispatcher internally, and has been moved to its own `array` module outside of the `dispatch` module since it is no longer the core interface for dispatching array functions. The eventual goal is to remove this class entirely and just work with raw array objects

### Details and comments

Depends on PR #61 which moves Array to its own module.


chriseclectic:
Closing this since the dispatcher is now planned to be moved to its own module

PR #61, opened 2021-11-29
Move Array to its own module

chriseclectic:
### Summary

This moves the `Array` class and `wrap` function  from `qiskit_dynamics.dispatch` to `qiskit_dynamics.array` to separate these classes from the underlying dispatch system.

### Details and comments

This is to separate the Array from the rewrite of dispatch system in #60 with the goal of eventually removing the need for the `Array` class at all.


DanPuzzuoli (approved):
LGTM

PR #69, opened 2021-12-08
Fixing errors introduced by recent JAX release

DanPuzzuoli:
### Summary

Needed to add additional `Array` wrapping in one spot in `rotating_wave_approximation` in which a raw JAX array was controlling a binary operation.

Also, in `final_state_converter`, JAX arrays were being fed into quantum info types, e.g. `DensityMatrix`, raising an error. I've changed this function to always convert the array to a `numpy.array` before wrapping in `cls`, as `cls` will always be a quantum info type, which can only accept `numpy.array`s in any case. The doc string for this function has been updated to explain this.

### DIFF
--- a/qiskit_dynamics/models/lindblad_model.py
+++ b/qiskit_dynamics/models/lindblad_model.py
@@ -419,9 +419,7 @@ class LindbladModel(BaseGeneratorModel):
             density matrix by left-multiplication. Allows for direct evaluate generator.
          * 'sparse': Like dense, but matrices stored in sparse format. If the default Array backend
             is JAX, uses JAX BCOO sparse arrays, otherwise uses scipy
-            :class:`csr_matrix` sparse type. Note that reverse mode automatic differentiation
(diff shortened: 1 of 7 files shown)


chriseclectic (commented):
If this is going into a patch release there should be a release note for `fixes` saying what the bug that was fixed is. I think in this case it is something like 

> Fixes a bug where simulations could raise an error using JAX solvers for Statevector and DensityMatrix final states.

DanPuzzuoli:
Updates:
- Added release note, as suggested.
- Discovered that new JAX release advanced sparse support enough that LindbladModel.evaluate_rhs can now be reverse mode differentiated when in sparse mode. Removed this caveat from the documentation, and changed the test case for this function to test reverse-mode autodiff instead of forward mode (reverse is more restrictive so it's enough to just test this one). This test change was done by deleting the forward mode test case, which overrode an inherited reverse-mode test case.
- Due to the above, bumped the minimum JAX version, and explained this in the release note.

chriseclectic (approved):

PR #72, opened 2022-01-05
Upper bound numpy version based on qutip import error

DanPuzzuoli:
### Summary

One of the tests is currently failing on the latest version of numpy (1.22.0) due to a bug it introduced in qutip. We don't really use qutip, but our tests do, so this is causing problems.


setup.py
```diff
@@ -17,6 +17,7 @@
 
 requirements = [
     "numpy>=1.17",
+    "numpy<1.22",
```
mtreinish, line 20: This isn't the best way to do this. This is basically equivalent to doing `pip install numpy>=1.17 numpy<1.22`, you can just use a comma to put this on the same line 

```suggestion
    "numpy>=1.17,<1.22",
```

mtreinish (commented):
Left an inline comment, but I'm wondering if it'd be better to do this via a constraints file. Since this isn't a general dynamics thing but just a testing issue when using qutip right because of qutip's incompatibility with the latest numpy release? Doing this in the setup.py like this is basically advertising that qiskit-dynamics doesn't work with the latest version of numpy.

DanPuzzuoli:
> Left an inline comment, but I'm wondering if it'd be better to do this via a constraints file. Since this isn't a general dynamics thing but just a testing issue when using qutip right because of qutip's incompatibility with the latest numpy release? Doing this in the setup.py like this is basically advertising that qiskit-dynamics doesn't work with the latest version of numpy.

That's correct, it's purely a testing issue as dynamics doesn't use qutip outside of tests. I agree that we shouldn't advertise that it doesn't work with the latest version of numpy. 

Can the numpy version be temporarily constrained in the CI configuration somehow?

DanPuzzuoli:
Just created a `constraints.txt` file specifying `numpy<1.22`, and removed the initial change.

mtreinish:
You don't need a requirements file here, because that's what the list in the setup.py is for. To use constraints you need to modify the tox.ini here: https://github.com/Qiskit/qiskit-dynamics/blob/main/tox.ini#L8 to be:
```
install_command = pip install -c{toxinidir}/constraints.txt -U {opts} {packages}
```

mtreinish (approved):

PR #73, opened 2022-01-12
Introduce perturbation module

DanPuzzuoli:
## Summary

This PR introduces a new perturbation theory sub module for qiskit dynamics. It establishes a layer of low level functions and objects for computing and utilizing time-dependent perturbation theory terms for linear matrix differential equations. 

### User-facing elements

The user-facing elements of this PR that explicitly do perturbation calculations are described in the API docs (included with this PR), but to summarize:
- Some classes for describing/representing multi-variable polynomials (which arise as truncated power series)
    - `ArrayPolynomial`: A class for representing a multi-variable array-valued polynomial. This can be evaluated on an array of variable-values, and transformed through either algebraic operations with other `ArrayPolynomial` objects, or by calling methods which reproduce some `np.ndarray` functionality (e.g. `conj()`, `trace()`, ...).
    - `Multiset`: A class representing a mathematical multi-set of integers. This class is used for representing *labels* of power series terms. The convention used here is to label a multi-variable power series term by the collection of variables appearing in the term (which can be repeated, hence multiset rather than set).  As described in the API docs, there are convenient correspondences between certain operations on power series and multiset operations on their labels that make this a good choice.
- Dyson and Magnus series computation (time-dependent perturbation theory)
   - `solve_lmde_perturbation`: A function which performs 3 general classes of time-dependent perturbation theory problems: "Dyson", "Magnus", and "Dyson-like" for a class of integrals encountered when working with the Dyson series.
   - `PerturbationResults`: A data container class similar to scipy's `OdeResults` for containing the results of `solve_lmde_perturbation`.
- Schrieffer-Wolff perturbation theory
   - `schrieffer_wolff`: A function for computing a truncated expansion for the generator of the Schrieffer-Wolff transformation in the multivariable setting. Currently specialized to the case of diagonalization, though could be generalized to block-diagonalization at a later time.
- Specialized perturbation-theory based LMDE solvers
   - `PerturbativeSolver`: A class that exposes two fixed-step solvers built from the Dyson series and Magnus expansions. Quite specialized and technical to use, but can result in huge speed ups for solving once constructed.

### Documentation

Due to the advanced and non-standard nature of this module, I've attempted to give a more detailed description of it in the API docs than other sub-modules. I.e. users may not immediately know what these perturbative expansions are, and the notation/representation is definitely not used in typical literature on the topics. I think this is mainly owing to the fact that typical presentations don't explicitly cover the multi-variable case, which we do here.

This PR also includes two new tutorials, which have been put under an "advanced" heading on the tutorial page:
- One walking through how to accelerate simulation using `PerturbativeSolver`, and
- One showing how to generate perturbative approximations to unitary evolution over a range of perturbative parameter values. This tutorial also includes a section investigating the quality of the approximation at various orders in the Magnus expansion.

### Non-user facing elements

The other, non-user facing, elements of the PR: 
- `dyson_magnus.py` 
    - Contains the actual computational pieces and helper functions for `solve_lmde_perturbation`
- `custom_dot.py`
    - A small module for defining, constructing, and computing a class of "custom multiplication rules" between lists of matrices (specified as 3d arrays).
    - May be useful at some point to make this a usable user-facing component in its own right, but for now it is mainly just a helper function for `dyson_magnus.py`.


DanPuzzuoli:
Just FYI to anyone seeing this PR - the tests are currently failing due to a bug in the latest version of JAX. This was discussed and identified [here](https://github.com/google/jax/discussions/9556), and was resolved in [this PR](https://github.com/google/jax/pull/9563). It looks like this PR was added to the 0.3.0 JAX release branch, but not sure when these changes will be reflected in pypi.

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+    # envelope function
+    envelope_func = lambda t: t * (T - t) / (T**2 / 4)
+
+3. How-to construct and simulate using the Dyson-based perturbative solver
+--------------------------------------------------------------------------
+
+Constructing the Dyson-based perturbative solver requires specifying several
+configuration parameters, as well as specifying the structure of the
+differential equation more explicitly than using the standard
+:class:`.Solver` object in qiskit-dynamics, which automatically builds
```
chriseclectic, line 104: Suggested rewording/ordering
```
Setting up a :class:.`DysonSolver` requires more setup han the standard
:class:`.Solver` as the user must specify several configuration parameters,
along with the structure of the differential equation:

- A :class:`.DysonSolver` requires direct specification of the LMDE to the
  solver. If we are simulating the Schrodinger equation, we need to
  multiply the Hamiltonian terms by ``-1j`` when passing in as LMDE
  operators.
- The :class:`.DysonSolver` is a fixed step solver, with the step size
  being fixed at instantiation. This step size must be chosen in conjunction
  with the ``expansion_order``, to ensure that a suitable accuracy is attained.
- Over each fixed time-step the :class:`.DysonSolver` solves by computing a
   truncated perturbative expansion.
  - To compute the truncated perturbative expansion, the signal envelopes are
    approximated as a linear combination of Chebyshev polynomials.
  - The order of the Chebyshev approximations, along with central carrier frequencies
    for defining the “envelope” of each ``Signal``, must be provided at instantiation.

See the :ref:`API docs <.DysonSolver>` for a more details.
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+
+  - :class:`.DysonSolver` solves by computing a truncated perturbative
+    expansion.
+  - To compute the truncated perturbative expansion, the signal envelopes are
+    approximated as a linear combination of Chebyshev polynomials.
+  - The order of the Chebyshev approximations, along with central carrier frequencies
+    for defining the “envelope” of each ``Signal``, must be provided at instantiation.
+
+
+.. jupyter-execute::
```
chriseclectic, line 126: If going with above ordering, can add a leading sentance to next block

```suggestion
For our example Hamiltonian we configure the :class:`.DysonSolver` as follows:

.. jupyter-execute::
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+        rotating_frame=-1j * static_hamiltonian,
+        dt=dt,
+        carrier_freqs=[v],
+        chebyshev_orders=[1],
+        expansion_order=7,
+        integration_method='jax_odeint',
+        atol=1e-12,
+        rtol=1e-12
+    )
+
```
chriseclectic, line 144: Can you add a sentence explaining the chosen values for expansion_oder and chebyshev orders?

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+
+Construct a function that simulates the system for the pulse sequence
```
chriseclectic, line 146: Things get confusing here. I think there should be longer paragraph explaining in a little more detail how to use the dyson solver to simulate something.

eg something like: 
> To use the Dyson solver we wrap it in a function which evaluates the perturbation expansion over a fixed length parameterized drive envelope and returns the final system states. This can then be used to compare the accuracy and performance with the regular solver over the same drive pulse. We also use JAX to just-in-time compile this function to improve performance for repeated evaluations for different amplitude parameter values.

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+with a given amplitude.
+
+.. jupyter-execute::
+
+    from qiskit_dynamics import Signal
```
chriseclectic, line 151: All imports at the top
```suggestion
    from qiskit_dynamics import Signal
    from jax import jit
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+
+    def dyson_sim(amp):
```
chriseclectic, line 153: Add doc string

```suggestion
    def dyson_sim(amp):
        """Evaluate dyson solver over a fixed drive envelope with given amplitude"""
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+        return dyson_solver.solve(
+            signals=[drive_signal],
+            y0=np.eye(dim, dtype=complex),
+            t0=0.,
+            n_steps=int(T // dt)
+        ).y[-1]
+
+    from jax import jit
+
+    jit_dyson_sim = jit(dyson_sim)
```
chriseclectic, line 164:
```suggestion
    # Jit the evaluation function to improve performance for repeated calls
    jit_dyson_sim = jit(dyson_sim)
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+
+First run includes compile time.
```
chriseclectic, line 166: Needs some explanantion
```suggestion
Because we are using a just-in-time compiled function, the first call will take longer as it compiles the functions. We can time this compilation using the Juypter magic method:
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+
+.. jupyter-execute::
+
+    %time yf_dyson = jit_dyson_sim(1.).block_until_ready()
```
chriseclectic, line 170: Maybe add a note on `block_until_ready` to previous sentence for people not familar with jax?
> Note that the `.block_until_ready` is required for timing the jax evaluation since the returned jax array is evaluated lazily.

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+
+
+Second run demonstrates the speed of the solver.
```
chriseclectic, line 173:
```suggestion
Once JIT compilation has been performance we can benchmark the performance of the jit-compiled solver:
```

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+    np.linalg.norm(yf_low_tol - yf_ode)
+
+Here we see that, once compiled, the Dyson-based solver has a
+significant speed advantage over the traditional solver, at the expense
+of the initial compilation time and the technical aspect of using the solver.
+
+5. How-to construct and simulate using the Magnus-based perturbation solver
+---------------------------------------------------------------------------
+
+Next, build the Magnus-based perturbative solver. The :class:`.MagnusSolver` uses the
```
chriseclectic, line 245:
```suggestion
We repeat our example using the Magnus-based perturbative solver. Setup of the :class:`.MagnusSolver`
is the same to the :class:`.DysonSolver`, but uses the Magnus expansion and
matrix exponentiation to simulate over each fixed time step.
```
Im not sure if it should be *the same as * or *similar to*?

chriseclectic (commented):
First part of review: This is just of the rst docs and mostly includes some formatting suggestions and areas where some extra details/clarification could be good.

qiskit_dynamics/perturbation/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+(i.e. a truncated power series), and provides functionality for
+both evaluating and transforming array-valued polynomials.
+
+This module makes use of the `multiset package <https://pypi.org/project/multiset/>`_ for
+indexing multi-variable power series. See the
+:ref:`multiset and power series notation section <multiset power series>`
+of the perturbation review for an explanation of this convention.
+
+
+Perturbation module API
```
chriseclectic, line 55: Maybe split up into two lists `Classes` and `Functions` ?

qiskit_dynamics/perturbation/perturbation_results.py
```diff
@@ -0,0 +1,76 @@
...
+from typing import List, Union
+from dataclasses import dataclass
+
+from multiset import Multiset
+
+from qiskit import QiskitError
+from qiskit_dynamics.array import Array
+
+
+@dataclass
```
chriseclectic, line 26: Maybe this should be a `frozen=True` dataclass so user cant change the fields once its returned?

qiskit_dynamics/perturbation/perturbation_results.py
```diff
@@ -0,0 +1,76 @@
...
+
+    .. code:: python
+
+        perturbation_results[[0, 1, 2]]
+
+    .. automethod:: __getitem__
+    """
+
+    expansion_method: str
+    expansion_labels: Union[List[List[int]], List[Multiset]]
```
chriseclectic, line 50: This is a bit of a confusing type for the dataclass. When should it be a list instead of a multiset?

releasenotes/notes/perturbation-module-b819dfac35c60551.yaml
```diff
@@ -0,0 +1,24 @@
...
+    differential equations.
+  - |
+    Adds data container class :class:`qiskit_dynamics.perturbation.PerturbationResults`
+    for storing and accessing the results of
+    :func:`~qiskit_dynamics.perturbation.solve_lmde_perturbation`.
+  - |
+    Adds class :class:`~qiskit_dynamics.perturbation.PerturbativeSolver`, which implements
+    specialized fixed-step solvers for linear matrix differential equations,
+    based on the Dyson series and Magnus expansion.
+other:
```
chriseclectic, line 22: This probably is an upgrade note

qiskit_dynamics/solvers/dyson_magnus_solvers.py
```diff
@@ -0,0 +1,995 @@
...
+            single_step = lambda x: self.model.Udt @ jexpm(self.model.evaluate(x).data)
+            ys = [y0, perturbative_solve_jax(single_step, self.model, signals, y0, t0, n_steps)]
+        else:
+            single_step = lambda coeffs, y: self.model.Udt @ expm(self.model.evaluate(coeffs)) @ y
+            ys = [y0, perturbative_solve(single_step, self.model, signals, y0, t0, n_steps)]
+
+        return OdeResult(t=[t0, t0 + n_steps * self.model.dt], y=ys)
+
+
+class ExpansionModel:
```
chriseclectic, line 380: Is there a base Model class this and other models can inherit from? Maybe something to think about later

qiskit_dynamics/solvers/dyson_magnus_solvers.py
```diff
@@ -0,0 +1,995 @@
...
+            The first dimension indexes the signals, and the second the time-step.
+        """
+        return self._approximate_signals(signals, t0, n_steps)
+
+    def evaluate(self, coeffs: Array) -> Array:
+        """Evaluate the expansion polynomial for a given list of coefficients."""
+        return self.expansion_polynomial(coeffs)
+
+
+def perturbative_solve(
```
chriseclectic, line 554: If these are all intended as private methods, might be good to add an underscore just to make this clear

qiskit_dynamics/solvers/dyson_magnus_solvers.py
```diff
@@ -0,0 +1,995 @@
...
+try:
+    import jax.numpy as jnp
+    from jax import vmap
+    from jax.scipy.linalg import expm as jexpm
+    from jax.lax import scan, associative_scan
+except ImportError:
+    pass
+
+
+class PerturbativeSolver(ABC):
```
chriseclectic, line 51: Can you break this up into 4 separate files for the 4 main classes: PertubativeSolver, MagnusSolver and DysonSolver, ExpansionModel? Might be good to put them in a subfolder for organization too.

chriseclectic (changes requested):
Looks good, few more comments on organization and api docs

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
+How-to use Dyson and Magnus based solvers
+=========================================
+
+In this tutorial we walk through how to use perturbation-theory based solvers.
+
+.. note::
```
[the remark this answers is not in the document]
DanPuzzuoli: Can you elaborate? Not sure what you mean by this.

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
+How-to use Dyson and Magnus based solvers
+=========================================
+
+In this tutorial we walk through how to use perturbation-theory based solvers.
```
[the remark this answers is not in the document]
DanPuzzuoli: Moved the sentence to after the notes. Also added an additional sentence referencing the documentation of the solvers for information about how they work, along with the discussion for background info.

Linking to the class docs is a little bit redundant with the first note, though they are linking to the docs for different things (API documentation v.s. description of what computations the class is actually doing).

docs/userguide/perturbative_solvers.rst
```diff
@@ -0,0 +1,308 @@
...
+    jax.config.update('jax_platform_name', 'cpu')
+
+    # set default backend
+    Array.set_default_backend('jax')
+
+2. Construct the model
+----------------------
+
+First, construct the model described in the introduction. Here we use a
+higher dimension to observe a difference between the solvers.
```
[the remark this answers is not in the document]
DanPuzzuoli: Expanded on this point in the doc.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Implemented this suggestion with a few edits.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Added this sentence.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Add a few sentences giving some context to the parameters.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Moved the JAX import.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Added.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: I think this might be getting too verbose/explain-y for a how-to guide. I know this one has to be more verbose/explain-y than the others when it comes to details/usage of the solvers, but I don't think we need to explain details of JAX compilation/magic methods.

I've added an explicit reference to the userguide entry on using JAX in the "Configure to use JAX" section of this doc. Let me know if this alleviates your concerns.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Same as above.

docs/userguide/perturbative_solvers.rst
DanPuzzuoli: Put this in - I think "similar to" is good here.

qiskit_dynamics/perturbation/__init__.py
DanPuzzuoli: I've split it into two sections.

qiskit_dynamics/perturbation/perturbation_results.py
DanPuzzuoli: Yeah good idea - changing it to this.

qiskit_dynamics/perturbation/perturbation_results.py
DanPuzzuoli: When `solve_lmde_perturbation` has been used to compute "dyson-like" terms (rather than the standard "Dyson" or "Magnus"), the terms are labelled by lists of integers, and not by multisets. The different types of labels are explained in detail in the `solve_lmde_perturbation` docs.

qiskit_dynamics/perturbation/perturbation_results.py
```diff
@@ -0,0 +1,76 @@
...
+    expansion_terms: Array
+
+    def __getitem__(self, label: Union[list, Multiset]) -> Array:
```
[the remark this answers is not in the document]
DanPuzzuoli: Just changed it to `get_item`.

releasenotes/notes/perturbation-module-b819dfac35c60551.yaml
DanPuzzuoli: Do you mean here that everything under "features" should be under "upgrade"?

docs/userguide/perturbative_solvers.rst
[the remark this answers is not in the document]
DanPuzzuoli: Changed to a warning - I like the colour difference between warning/note for this.

releasenotes/notes/perturbation-module-b819dfac35c60551.yaml
DanPuzzuoli: Changed to upgrade - after talking to @mtreinish  I've realized I misunderstood the typical qiskit usage of `upgrade` section.

qiskit_dynamics/solvers/dyson_magnus_solvers.py
DanPuzzuoli: Ended up deciding to go with the underscores - changed every class/method that is internal to have an underscore at the beginning of the name.

qiskit_dynamics/perturbation/perturbation_results.py
DanPuzzuoli: Okay I've ended up changing this, which I think simplifies it a bit. I've made a base class called `_LabeledData` for indexing a list of data with arbitrary labels. Obviously this is dictionary-like but it's nice to not have to make labels hashable, and the class can hold metadata.

There are then two subclasses `PowerSeriesData` and `DysonLikeData` - all they do functionally is make explicit assumptions about what the labels are (`Multiset` or `list`), and what the stored data is (it should be returned as an `Array`).

I still haven't set these to frozen or anything.

I think these changes make the functioning of these data classes a bit clearer. Either way it shouldn't be too important.

DanPuzzuoli:
Okay I've made the following changes since our discussion:

- Moved the `multiset` dependency release note from the `other` section to the `upgrade` section.
- For the userguide entry on perturbative solvers:
     - Changed the initial paragraph to a warning. Definitely is more of a “warning” than a note, and also since we have 2 notes back to back it’s nice to have color differentiation between them.
     - Added a bit more of an explanation about how we need to compile to properly compare speed of different solver methods, and to do this we need to write functions wrapping the calls to the solver.
- Went through the entire PR and put underscores in front of all internal classes/methods.
- Changed the `PerturbationResults` into 3 classes: 
    - an internal class that represents “labeled data” implementing the `get_item` method, and
    - two that inherit from it: `PowerSeriesData` (labels are multisets) and `DysonLikeData` (labels are lists).
    - I think this simplifies these classes and should make it more explicit what the results objects contain.

chriseclectic (approved):

PR #82, opened 2022-02-09
Add Python 3.10 support/testing, remove numpy version constraint, update black, astroid, and pylint

DanPuzzuoli:
### Summary

Various maintenance tasks triggered by a qutip release that supports the latest numpy, as well as qiskit-terra now supporting python 3.10.


.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
```
mtreinish, line 13: You technically don't need quotes around the single digit versions since the float representation is the same as the string. But it doesn't hurt either

mtreinish (commented):
LGTM, probably worth adding a release note though to indicate the addition of Python 3.10 support since it's a new feature and definitely something worth highlighting

releasenotes/notes/python-310-support-f7040d4ff5e4022e.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Added support for Python 3.10.
```
mtreinish, line 4: A bit more verbiage here would probably be useful. Maybe something like:

```suggestion
    Added support for running with Python 3.10. Starting with this release Qiskit Dynamics now supports
    running with Python 3.7 through 3.10.
```

mtreinish (approved):
LGTM, some small suggestions in the release notes so it reads a bit clearer when ti gets published in the documentation. But it's not critical (and something we can expand at release time.

mtreinish (approved):

PR #87, opened 2022-02-14
Rearranging which test configurations test JAX

DanPuzzuoli:
### Summary

Sets test environments to use the JAX version for ubuntu and macOS, and only run the non-jax test environments on windows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,12 +29,12 @@ jobs:
             ${{ runner.os }}-${{ matrix.python-version }}
       - name: Install Deps
         run: python -m pip install -U tox setuptools virtualenv wheel
-      - name: Install and Run Tests
-        if: ${{ matrix.os }} == "windows-latest"
-        run: tox -e py
       - name: Install and Run Tests with JAX
-        if: ${{ matrix.os }} != "windows-latest"
         run: tox -e jax
+        if: runner.os != 'Windows'
+      - name: Install and Run Tests
+        run: tox -e py
+        if: runner.os == 'Windows'
   lint:
     name: lint
     runs-on: ubuntu-latest


chriseclectic (commented):
If we are only going to have a single workflow for each OS (Linux+mac w jax, win wo jax) we could probably edit the workflow so that you just have a conditional for the os on which tox job to run for installing dependencies (@mtreinish probably knows syntax for doing that)

mtreinish (commented):
Yeah you can do that. If you change:
```yaml
- name: Install and Run Tests
  run: tox -e py
```

to something like

```yaml
- name: Install and Run Tests
  run: tox -e jax
  if: runner.os != "Windows"
- name: Install and Run Tests
  run: tox -e py
  if: runner.os == "Windows"  
```

you can eliminate the second jax specific job and have a single job definition for the full matrix

.github/workflows/main.yml
```diff
@@ -7,6 +7,13 @@ on:
 jobs:
   tests:
     name: tests-python${{ matrix.python-version }}-${{ matrix.os }}
+    rules:
+      - if: ${{ matrix.os }} == "windows-latest"
+        variables:
+          run_tests_command: "tox -e py"
+      - if: ${{ matrix.os }} != "windows-latest"
+        variables:
+          run_tests_command: "tox -e jax"
```
mtreinish, line 16: I don't think `rules` is a valid key. Reading the env variable section of github actions's documentation I'm not sure there is a way to do conditional variables like this: https://docs.github.com/en/actions/learn-github-actions/environment-variables

There might be a way with inline expressions but that'll be a bit harder to read imo if possible. I think my earlier suggestion is the best path forward here. Where we just do an if else in the steps to run either job based on the runner

```suggestion
    env:
      run_tests_command: ${{ 
        variables:
          run_tests_command: "tox -e py"
      - if: ${{ matrix.os }} != "windows-latest"
        variables:
          run_tests_command: "tox -e jax"
```

DanPuzzuoli: Thanks, I had tried that but I think it wasn't quite right, and then got lost in some random pages through google :P

chriseclectic (approved):

PR #91, opened 2022-02-22
Changing `DiscreteSignal.envelope` to return `0.` for times outside sample range

### LINKED ISSUE #90: Change `DiscreteSignal.envelope` to return `0.` for times outside sample range
Currently `DiscreteSignal.envelope` is setup to clip the time into the time-range on which the samples are defined. The result is that for all time before start time, the envelope evaluates to the first sample, and for all times after the end time, it evaluates to the last sample. 

 E.g. For a discrete signal with `start_time=0`, `samples=[1., 2., 3.]` and `dt=1.`, calling `discrete_signal.envelope(4.)` will return `3.`, and `discrete_signal.envelope(-1.)` will return `1.`.

It will make more sense, and be more useful for coming applications, if the envelope returns `0.` when `t` is outside the sample window.

brosand:
### Summary

Closes #90, causing the `DiscreteSignal.envelope` to return `0.0` for any time input outside of the sample range.

### Details and comments

This is achieved by adding a float 0 to the start and end of the sample list, and changing the clipping accordingly. This is preferable (at first pass) to using jax conditionals due to the dependence on a `DeviceArray`.


DanPuzzuoli (commented):
Padding the samples seems like an interesting approach so that the whole thing is still done purely with array indexing (no conditionals as you say). From the perspective of constructing the envelope function this is a good approach, but it actually interferes with other functionality: the pulse to signals tutorial is failing to build, and I think its because the zero padding doesn't play nice with the `DiscreteSignal.add_samples` method. `add_samples` modifies the internal `_samples` attribute, but with the changes here this has no effect on the envelope, due to the `_widesamples` attribute being unaffected.

This is indicative of a bit of awkwardness with this approach. It doesn't necessarily invalidate it, but *if possible* we should try to avoid using something like `_widesamples`, assuming it can be done without taking a performance hit. For the record, I don't actually know if this is possible, but nevertheless it would be good to try alternate options to try to avoid any unnecessary mismatches between the user facing attribute `samples` and its internal representation, as it complicates its usage and its internal functionality.

In a jupyter notebook you can write a couple different versions of these envelope functions, and time them to see which is fastest. If the padding approach is clearly faster then we can stick with it and deal with the internal awkwardness, otherwise we can maybe do something else that preserves using only the existing `_samples` attribute unchanged. The options I can think of:
- Actually write the envelope function using a conditional, assuming that `t` is a scalar, then call `np.vectorize()` on the envelope function before passing it to the `Signal` constructor. (For JAX, use `cond` for the conditional and `jnp.vectorize`). Time this for `t` being an array or scalar, and for samples being 1d or 2d.
- The padding approach.
- I still have a feeling there may be some internal numpy function handling conditionals in a vectorized way that will work, but I could be wrong. I just tried doing it using `np.where` but ended up running into the issue that I still had to pad the samples! :P 

In any case, for the padding approach, my comments are:
- The `_samples` and `_widesamples` attributes contain duplicate information. If this approach ends up being used only the padded version should remain, and the `samples` attribute should return the truncated version.
- When constructing `_widesamples` you actually only need to add the zero on the end. You're currently calling `_widesamples[idx + 1]` so that if `idx==-1` `idx + 1 == 0`. However, if you do `a[-1]` it will actually return the last sample, so there's no need to shift the index. Fairly minor but I think it's slightly more elegant.
- I don't think there's any need for the shape checking/validation. Both cases can be handle via (assuming you only append at the end): 
```
zero_pad = np.expand_dims(np.zeros_like(Array(samples[0])), 0)
wide_samples = np.append(samples, zero_pad, axis=0)
```
- I prefer the name `_padded_samples` over `_widesamples`. "Padding" is a common terminology for something like this whereas "wide" is ambiguous.
- `add_samples` will need to be corrected to work with the padded arrays. Also, can you add a test case for this method to make sure its caught somewhere other than the docs?

brosand:
Comparing the padding method vs a conditional approach, I observed a modest speedup from the padding approach, at least for a simple test case of sampling from a very simple signal n times, with ns up to 10,000,000. For example, at n=10,000,000 the sampling time for the padding approach is 1.69 and the conditional approach is 1.93. These differences can be observed in the attached python file (in text form for GitHub).
[signal_demo.txt](https://github.com/Qiskit/qiskit-dynamics/files/8235486/signal_demo.txt)

DanPuzzuoli:
I guess lets proceed with the padding method. I'm still a little bit weirded out by how it will impact sample manipulation in the various methods, but if we're careful it should be fine I think.

DanPuzzuoli (changes requested):
I left a few minor comments, but otherwise its looking good.

There are two other things:
- Can you modify the `DiscreteSignal` class doc string to explain that the envelope will be evaluated to `0` if the time is outside the sample window. It currently gives a formula for the how it is evaluated `s[floor((t-t0) ...]`, but now it needs to be updated to say it'll be 0 outside of this. Even though it's internally implemented with padding you can just explain it via if statements.
- The only other thing is to add a reno file. You can see how to do this in the `Contributing.md` for the package. I think you can probably either phrase this change as a "fix", or just in "other". It's not a huge deal where it goes, as the renos are all consolidated/reviewed for each release.

releasenotes/notes/discrete_signal_bounds_handling-e960ae879c409233.yaml
```diff
@@ -0,0 +1,14 @@
+---
+depreciated:
```
DanPuzzuoli, line 2: As this is not a public attribute this is not a deprecation.

releasenotes/notes/discrete_signal_bounds_handling-e960ae879c409233.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - |
+    The internal attribute `DiscreteSignal._samples` has been removed,
+    all methods should now be calling the public method `DiscreteSignal.samples`.
+
+other:
```
DanPuzzuoli, line 7: Similarly here, no need to refer to `_samples`. The description also shouldn't refer to new private methods.

You could maybe say this more concisely:

```
The `DiscreteSignal.envelope` function has been changed so that if `t` is outside the interval defined by `start_time` and `start_time + dt * len(samples)` it will return `0.0`.
```

releasenotes/notes/discrete_signal_bounds_handling-e960ae879c409233.yaml
DanPuzzuoli, line 2: As this is not a public attribute this is not a deprecation. You can remove this section.

qiskit_dynamics/signals/signals.py
```diff
@@ -272,6 +272,10 @@ class DiscreteSignal(Signal):
     The envelope is specified by an array of samples ``s = [s_0, ..., s_k]``, sample width ``dt``,
     and a start time ``t_0``, with the envelope being evaluated as
     :math:`f(t) =` ``s[floor((t - t0)/dt)]``.
+    If the signal is sampled outside of its defined sample list, it will return a 0.0. So if `
```
DanPuzzuoli, line 275: The comment that the envelope is `0.0` outside of the interval `start_time` and `start_time + dt * len(samples)` should be merged into the preceding sentence. E.g.

```
    The envelope is specified by an array of samples ``s = [s_0, ..., s_k]``, sample width ``dt``,
    and a start time ``start_time``, with the envelope being evaluated as
    :math:`f(t) =` ``s[floor((t - start_time)/dt)]`` if ``t`` is in the interval with endpoints ``start_time``
   and ``start_time + dt * len(samples)``, and ``0.0`` otherwise.
```

DanPuzzuoli (changes requested):
Just about done, I've left a couple of final changes in comments.

Now the main thing is that you should avoid referring to internal details in documentation. E.g. there is no need to deprecate the usage of `DiscreteSignal._samples`, as it is internal no one should be using it. `DiscreteSignal.samples` has always been the "public version", and its usage hasn't changed.

Similarly when describing the definition of the envelope function, there is no need to describe how it is implemented, as these are internal details that we are not committing to maintain.

DanPuzzuoli (approved):
LGTM!

PR #97, opened 2022-04-22
Remove deprecated dispatch methods

chriseclectic:
### Summary

This removes the deprecated dispatch methods that were scheduled to be removed  in 0.3.0 release.


DanPuzzuoli (changes requested):
Only thing is that I think the release note should be more explicit about where to find the non-deprecated versions - some are importable functions from `qiskit_dynamics.array`, and some are class methods of `qiskit_dynamics.array.Array`.

DanPuzzuoli (approved):
LGTM

PR #98, opened 2022-04-25
Add explicit deprecation of carriers argument as list in InstructionToSignals

### LINKED ISSUE #88: Add handling and deprecation warning for old carriers formatting for InstructionToSignals.get_signals
As pointed out by @chriseclectic , the function `InstructionToSignals.get_signals` has been updated so that the `carriers` argument is supplied as a dictionary mapping channel names to frequencies, rather than a list of frequencies (whose channel associations are automatically determined by the iteration ordering of `schedule.channels`). This is a breaking change (and was discovered by running `main` code while playing around with the existing pulse to signals tutorial). 

We should handle this change with a deprecation warning:
- Re-add the ability to pass `carriers` as a list by creating the proper `dict` format using the associations the old code would have generated.
- Include a deprecation warning, that this style of formatting is deprecated as of `0.3.0`, and will be fully removed by `0.4.0`.
- Update the release note file for the `pulse-converter-filtering` to include a note that this deprecation is occurring.

DanPuzzuoli:
### Summary

Closes #88.

### Details and comments

Re-adds the handling of ``carriers`` being passed as a list to the ``InstructionToSignals`` constructor. If passed as a list, a ``DeprecationWarning`` is raised indicating that this will be removed in the subsequent release. A new test has also been added to verify this behaviour.

### DIFF
--- a/qiskit_dynamics/pulse/pulse_to_signals.py
+++ b/qiskit_dynamics/pulse/pulse_to_signals.py
@@ -14,6 +14,7 @@
 Pulse schedule to Signals converter.
 """
 
+import warnings
 from typing import Dict, List, Optional
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


eggerdj (approved):
LGTM. Thanks!

PR #100, opened 2022-05-09
Lindblad model RWA bug fix

DanPuzzuoli:
### Summary

Fixes a bug in the `rotating_wave_approximation` function for when `model` is a `LindbladModel` and `return_signal_map=True`. The signature of the returned signal mapping function was inconsistent with the documentation, and has been fixed to restore consistency.

### Details and comments

For a `LindbladModel` the rwa signal mapping function must act on both the Hamiltonian signals as well as the dissipator signals. The documentation for the `rotating_wave_approximation` implies that the returned signal mapping function, in the case of a `LindbladModel`, should have a single argument, corresponding to the structure of the `LindbladModel.signals` property, which contains both the Hamiltonian and dissipator signals in a two-entry tuple.

Currently, however, the signal mapping  function returned by `rotating_wave_approximation` accepts the Hamiltonian and dissipator signals as two separate arguments, which is inconsistent. This PR fixes this to be consistent with the documentation.

### DIFF
--- a/qiskit_dynamics/models/rotating_wave_approximation.py
+++ b/qiskit_dynamics/models/rotating_wave_approximation.py
@@ -232,7 +232,7 @@ def rotating_wave_approximation(
         )
 
         if return_signal_map:
-            signal_translator = lambda a, b: (get_rwa_signals(a), get_rwa_signals(b))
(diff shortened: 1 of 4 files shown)


brosand (approved):
LGTM

PR #102, opened 2022-05-12
Fixing single-line pylint disable of unused-import

DanPuzzuoli:
### Summary

Both `solvers/solver_functions` and `solver/solver_classes` had the lines:

```
# pylint: disable=unused-import
from scipy.integrate._ivp.ivp import OdeResult
```

`OdeResult` is being imported for type-hinting, but since it isn't being used for anything else the pylint error `unused-import` was being raised.

The intention was to disable this error for only the `OdeResult` import, but it actually accidentally disables it for the entire file, rather than just the single line. This has been corrected and the unused imports in these files have been removed.

### DIFF
--- a/qiskit_dynamics/solvers/solver_classes.py
+++ b/qiskit_dynamics/solvers/solver_classes.py
@@ -23,8 +23,7 @@ from copy import copy
 
 import numpy as np
 
-# pylint: disable=unused-import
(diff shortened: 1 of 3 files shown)


brosand (approved):
LGTM -- good to keep an eye on these pylint lines

PR #104, opened 2022-05-16
Add diffrax solvers

### LINKED ISSUE #93: Add diffrax solvers to solve_ode
[Diffrax](https://github.com/patrick-kidger/diffrax) is a JAX ODE solving library providing access to many different solvers, and more control over how computations (like back propagation) are done. It also allegedly offers speed improvements over JAX's `odeint` function for the same method and tolerances.

Add access to solvers from this library through `solve_ode`.

brosand:
### Summary

Add the option to pass a Diffrax solver in for the solver method. [Diffrax](https://github.com/patrick-kidger/diffrax) is a library built on Jax to enable flexibility and speed using many more differential equation solvers than the built in Jax option. 

### Details and comments
This PR enables a user to pass in an `AbstractSolver` class from `Diffrax`, previously solvers were determined through a string, but this PR enables the Diffrax modality of defining solvers as objects. The user can initialize the solver with any options they prefer, or pass in the initialized class to have Dynamics initialize for them. In addition, a Stepsize Controller can also be passed into the qiskit-dynamics solve functions, in order to utilize more features of Diffrax.

In addition, testing is added for this PR, as well as testing to confirm that solvers can perform differentiation both forwards and backwards through time.

Closes #93


qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
+# -*- coding: utf-8 -*-
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2020.
```
DanPuzzuoli, line 5: Set the copyrights to 2022

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+@requires_backend("jax")
+def diffrax_solver(
+    rhs: Callable,
+    t_span: Array,
+    y0: Array,
+    method: "AbstractSolver",
+    t_eval: Optional[Union[Tuple, List, Array]] = None,
+    **kwargs,
+):
+    """Routine for calling `diffrax.diffeqsolve`
```
DanPuzzuoli, line 47: Make sure to put two ticks here:
```
``diffrax.diffeqsolve``
```
this doesn't get rendered in the docs anyway but might as well use correct sphinx syntax

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Args:
+        rhs: Callable of the form :math:`f(t, y)`
```
DanPuzzuoli, line 50: Add period to rhs sentence.

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+        method: Which diffeq solving method to use.
+        t_eval: Optional list of time points at which to return the solution.
+        **kwargs: Optional arguments to be passed to ``diffeqsolve``.
+
+    Returns:
+        OdeResult: Results object.
+    """
+    if isinstance(method, type) and issubclass(method, AbstractSolver):
+        solver = method()
+    else:
```
DanPuzzuoli, line 62: What is the `else` block doing here? Is it to catch if it's an already instantiated instance? Does diffrax allow you to pass either the class or an instance?

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+        solver = method
+
+    t_list = merge_t_args(t_span, t_eval)
+
+    # convert rhs and y0 to real
+    rhs = real_rhs(rhs)
+    y0 = c2r(y0)
+
+    # TODO: do we want to allow kwargs for this part as well?
```
DanPuzzuoli, line 71: How is this controlled in diffrax? To pass things like `atol` and `rtol`, does the user need to define the step-size controller?

If so, I'm partial to just requiring the user to do this, rather than allowing them to specify a stepsize controller, and also in some cases handling `rtol` and `atol` manually.

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+
+    term = ODETerm(lambda t, y, _: Array(rhs(t.real, y), dtype=float).data)
+
+    # Ensure that atol and rtol are not present to be passed into the diffrax solver
+    kwargs.pop("rtol")
+    kwargs.pop("atol")
+
+    diffeqsolve = wrap(_diffeqsolve)
+
+    saveat = SaveAt(ts=t_list)
```
DanPuzzuoli, line 88: How does `SaveAt` work? It may not actually be necessary here to merge `t_span` and `t_eval`, should this just be `SaveAt(ts=t_eval)`? (Whether it should or shouldn't depends on how both control the output values.)

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+        **kwargs,
+    )
+
+    ys = jnp.array([r2c(y) for y in results.ys])
+    results = OdeResult(t=t_list, y=Array(ys, backend="jax", dtype=complex))
+
+    return trim_t_results(results, t_span, t_eval)
+
+
+def real_rhs(rhs):
```
DanPuzzuoli, line 108: Is it necessary to redefine `real_rhs`, `c2r`, and `r2c`? They exist in `scipy_solve_ivp.py`, so can we not just import those?

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -105,8 +113,12 @@ def solve_ode(
         QiskitError: If specified method does not exist.
     """
 
-    if method not in ODE_METHODS and not (
-        isinstance(method, type) and issubclass(method, OdeSolver)
+    if (
```
DanPuzzuoli, line 116: Can we merge this into:
```
if (
    method not in ODE_METHODS
    and not (isinstance(method, type) and (
        issubclass(method, OdeSolver) or (diffrax_installed and issubclass(method, AbstractSolver))
    )
):
```

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,125 @@
...
+        t1=t_list[-1],
+        dt0=None,
+        y0=Array(y0, dtype=float),
+        stepsize_controller=stepsize_controller,
+        saveat=saveat,
+        **kwargs,
+    )
+
+    ys = jnp.array([r2c(y) for y in results.ys])
+    results = OdeResult(t=t_list, y=Array(ys, backend="jax", dtype=complex))
```
DanPuzzuoli, line 103: What is the return type of `diffeqsolve`? Is there any other information in that object that we might want to include? E.g. `odeint` doesn't return any additional information, so we just pass the times and the `ys`, but if there is more info (like number of steps taken, etc) it'd be cool to include.

releasenotes/notes/add-diffrax-solvers-946869d5a304318a.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - |
+    Added support for solvers from the diffrax package:
+    https://github.com/patrick-kidger/diffrax. A new option 
+    is enabled to pass in an object -- a solver from diffrax
+    instead of a string for a jax or scipy solver, for example::
+    
+      from diffrax import Dopri5
+      from qiskit-dynamics.solvers.solver_functions import solve_ode
+
+      sol = solve_ode(
```
DanPuzzuoli, line 12: Is the `example::` valid sphinx?

For this it looks like a code example but you have used a type-hinting syntax, which seems odd. Would it be better to just do:
```
sol = solve_ode(
    rhs,
    t_span,
    y0,
    method=Dopri5()
)
```

test/dynamics/common.py
```diff
@@ -101,6 +106,32 @@ def jit_grad_wrap(self, func_to_test: Callable) -> Callable:
...
+    Test cases that inherit from this class will automatically work with diffrax solvers
+    backend.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        try:
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                # pylint: disable=import-outside-toplevel
```
DanPuzzuoli, line 121: should the `import-outside-toplevel` also be moved to the next line, like:
```
import diffrax  # pylint: disable=unused-import,import-outside-toplevel
```
Not sure if this will disable it for the rest of the file.

test/dynamics/common.py
```diff
@@ -101,6 +106,32 @@ def jit_grad_wrap(self, func_to_test: Callable) -> Callable:
...
+                import diffrax  # pylint: disable=unused-import
+        except Exception as err:
+            raise unittest.SkipTest("Skipping diffrax tests.") from err
+
+    def Dopri5(self) -> Callable:
```
DanPuzzuoli, line 126: Is this necessary? Can you not just import it in whatever file you're using this in?

Edit: It seems like you're importing them anyway in other files.

tox.ini
```diff
@@ -37,6 +39,7 @@ deps =
     -r{toxinidir}/requirements-dev.txt
     jax
     jaxlib
+    git+https://github.com/patrick-kidger/diffrax.git@main#egg=diffrax
```
DanPuzzuoli: Just a reminder to update this.

DanPuzzuoli (changes requested):
Overall looks good.

I left a bunch of minor questions/comments. I think the main questions are:
- What is the return type of the diffrax solver, and is there anything else being returned that we'll want to include in the `OdeResult` object? Alternatively, could we just return whatever the diffrax solver returns, if the main `results.y` and `results.t` property names are the same? 
- How does `SaveAt` work? Here I mean things like: does diffrax only return the t and y values specified in `SaveAt`? I *think* that's how `t_eval` works in scipy, e.g. if `t_eval` doesn't contain the `t_span` endpoints it won't actually return those. If `SaveAt` works this way we don't even need to call `merge_t_args`, could just pass `t_eval` directly to `SaveAt`. (Confirm scipy's behaviour when looking into this.)
- For any `kwargs` to the diffrax solver that we haven't committed to support in the `solve_ode`/`solve_lmde` interfaces, there's no need to handle them. E.g. if `atol` and `rtol` aren't kwarg options to the diffrax solver then we don't need to support them, the user should pass the step size controller.

qiskit_dynamics/solvers/diffrax_solver.py
brosand: No that was me allowing both, should I just stick to instance?

qiskit_dynamics/solvers/diffrax_solver.py
brosand: yes there are other things it returns, can the OdeResult add other optional kwargs? hard to tell from docs

releasenotes/notes/add-diffrax-solvers-946869d5a304318a.yaml
brosand: Are you saying to remove the imports as well?

brosand: should I remove the import lines as well?

DanPuzzuoli: Maybe keep the imports as it shows where things are coming from, but you need to fix the `solve_ode` one:
```
from qiskit_dynamics import solve_ode
```
This part is pretty minor as all release notes from other changes get merged into a single file for a release, and at this stage they typically get re-edited anyway.

qiskit_dynamics/solvers/diffrax_solver.py
DanPuzzuoli: `OdeResult` is basically just a data container class, so it's within standard usage to just add attributes to these things as necessary. You can compare the standard attributes of the `OdeResult` (return type described [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html)) and `RESULTS` object [here](https://github.com/patrick-kidger/diffrax/blob/main/diffrax/solution.py) and try to map them onto eachother.

qiskit_dynamics/solvers/diffrax_solver.py
DanPuzzuoli: Yeah stick to instance if that's what diffrax typically requires.

qiskit_dynamics/solvers/diffrax_solver.py
brosand: They are in the numpy format for the scipy version, so this is the jaxified version. I guess r2c could be imported though

test/dynamics/common.py
brosand: Yes, I had set this up for an issue with importing, which was solved later.

brosand: Removed this

qiskit_dynamics/solvers/diffrax_solver.py
brosand: I tried to map some important features, let me know if this seems reasonable

DanPuzzuoli: I think here we should try to write something that doesn't explicitly need to reference the fields of the diffrax `Solution` object, as things may be added or changed there. 

Here is an approach which is based on the fact that the `OdeResult` object is really just a data container with no extra information or requirements (Note that this is a detailed sketch and may need to be refined to fit into what you want):

```
solution = diffeqsolve(...)

# convert to dictionary
sol_dict = vars(solution) 

# pop and store the attributes we need to handle specially
y = sol_dict.pop('ys')
t = sol_dict.pop('ts')

# do any extra computations
...

# construct OdeResult object from all non-popped attributes
ode_results = OdeResult(**sol_dict)

# set popped attributes
ode_results.y = y
ode_results.t = t

return ode_results
```

qiskit_dynamics/solvers/diffrax_solver.py
DanPuzzuoli: yeah okay I think it's fine to leave this - it does create some logic reproduction, but it is probably best to let each solver do what is more efficient, as opposed to merging them by using `dispatch`

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,135 @@
...
+    # convert rhs and y0 to real
+    rhs = real_rhs(rhs)
+    y0 = c2r(y0)
+
+    term = ODETerm(lambda t, y, _: Array(rhs(t.real, y), dtype=float).data)
+
+    diffeqsolve = wrap(_diffeqsolve)
+
+    if "saveat" in kwargs and t_eval is not None:
+        raise QiskitError(
```
DanPuzzuoli, line 77: I think here I'd prefer a phrasing like:
```
Only one of t_eval or saveat can be passed when using a diffrax solver, but both were specified.
```

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -0,0 +1,135 @@
...
+            term,
+            solver=method,
+            t0=t_list[0],
+            t1=t_list[-1],
+            dt0=None,
+            y0=Array(y0, dtype=float),
+            **kwargs,
+        )
+
+    ys = jnp.array([r2c(y) for y in results.ys])
```
DanPuzzuoli, line 105: I think the scipy solver does exactly this, but since this solver uses JAX it might be good to avoid python loops by utilizing array operations:

```
ys = r2c(y).transpose((1, 0))
```

Fairly certain the above is correct, but you should double check. Actually it may be good to make this change in the scipy solver too as I believe the array transpose should be faster.

DanPuzzuoli (commented):
I commented directly on the open issues, and added a few more comments. I think with this last few changes it'll be good to go!

DanPuzzuoli (approved):
LGTM!

PR #105, opened 2022-05-18
Updating Solver.solve to take signals as an argument and to accept multiple simulations

DanPuzzuoli:
### Summary

The main purposes of this PR are:
- Update `Solver.solve` to take signals as an argument, rather than passing them to the constructor of `Solver` or by setting the `Solver.signals` property.
- Update `Solver.solve` to perform lists of simulations by specifying `t_span`, `y0`, and/or `signals` as lists of valid single inputs. This can be controlled by setting all as lists of the same length, or by setting only some as lists of the same length, and the others as a single input to be used repeatedly.

In connection with the above, secondary changes are:
- A new `rwa_carrier_freqs` argument has been added to the `Solver` constructor for specifying the carrier frequencies relative to which the rotating wave approximation should be performed. This information used to be extracted from the signal arguments to the constructor. During the deprecation period, if both signals and `rwa_carrier_freqs` are passed to the constructor, `rwa_carrier_freqs` will take precedence.
- The `hamiltonian_signals` and `dissipator_signals` arguments to the `Solver` constructor are deprecated.
- The `Solver.signals` property has been deprecated.
- The `Solver.copy` method has been deprecated. It is no longer needed now that the object no longer internally stores signals.

All deprecation warnings indicate that the deprecated behaviour will be removed in the next major release.

### Details and comments

- The old behaviour is preserved - if a user specifies signals at instantiation or via the `signals` property, and doesn't pass `signals` to `Solver.solve`, the solver will fall back on the signals stored in the `signals` property.
- When calling `Solver.solve` with a single simulation, the results will be returned as before in an `OdeResult` instance, however if a list of simulations is specified, the results will be returned in a list of `OdeResult` instances.
- Tests updates:
      - Tests have been added checking that deprecation warnings are raised, and that deprecated behaviour still works.
      - Tests have been added verifying that running a list of simulations using a single call to `solve` is consistent with running only a single simulation.
      - Existing tests have been updated to pass `signals` to `Solver.solve` rather than at instantiation or through the property.
- Documentation updates:
      - `(Deprecated)` has been added to the beginning of the doc string for every property, method, and argument that has been deprecated.
      - The `Solver.solve` method doc has been updated to include a description + examples of specifying a list of simulations.
      - All tutorials have been updated to pass `signals` to the `solve` method. 

Note that most lines added are due to tests.


releasenotes/notes/solver-generalization-ad588a54c830c66f.yaml
```diff
@@ -0,0 +1,27 @@
+---
+upgrade:
+  - |
+    The :class:`~qiskit_dynamics.Solver` class has been modified to no longer
```
chriseclectic, line 4: You can make these sphinx links work without the full path doing `.Solver` or `~.Solver` (assuming the class/function names are unique in the documentation scope)

```suggestion
    The :class:`~.Solver` API has been updated so that ``signals`` are passed
    as a kwarg to the :meth:`.Solver.solve` method rather than being set as a
    class attribute. A new instantiation argument, ``rwa_carrier_freqs``, has been
    added to control behaviour of the rotating wave approximation in the absence
    of signals being specified.
```

Probably dont need to mention the deprecation twice

releasenotes/notes/solver-generalization-ad588a54c830c66f.yaml
```diff
@@ -0,0 +1,27 @@
...
+    internally store signals: rather than passing them at instantiation, or setting them
+    via the ``signals`` property, ``signals`` must be passed as an argument to the
+    ``solve`` method when solving. Setting ``signals`` at instantiation or through the
+    property has been deprecated (see deprecations). A new instantiation argument,
+    ``rwa_carrier_freqs``, has been added to control behaviour of the rotating wave
+    approximation in the absence of signals being specified.
+  - |
+    The ``solve`` method of :class:`~qiskit_dynamics.Solver` can now be made to perform
```
chriseclectic, line 12: This bullet should go under `features` since its adding a new feature not a breaking API change

```suggestion
    The  :meth:`.Solver.solve` method can now run multiple simulations in a single call by specifying...
```

Might be worth noting that these simulations are not currently parallelized are are run one after the other?

releasenotes/notes/solver-generalization-ad588a54c830c66f.yaml
```diff
@@ -0,0 +1,27 @@
...
+    lists of simulations by specifying one or more of the ``t_span``, ``y0``, and ``signals``
+    arguments as lists of valid single inputs. This supports specifying all as a list of the
+    same length (with a different value for each to be used in each simulation), or a subset
+    can be specified with lists of the same length, with the other arguments giving only a single
+    value (to be used repeatedly while varying the other parameters). Results are returned as a
+    list of the results of the individual simulations.
+deprecations:
+  - |
+    Storing signals in the :class:`~qiskit_dynamics.Solver` class has been deprecated:
```
chriseclectic, line 21: The storing doesnt really need to be mentioned:

```suggestion
    - |
       The ``signals`` property and ``hamiltonian_signals`` and ``dissipator_signals`` init kwargs of :class:`.Solver` have been deprecated. Signals are now specified as a kwarg to the :meth:`.Solver.solve` method. The deprecated functionality is scheduled to be removed in the <actual version> release.

File: qiskit_dynamics/solvers/solver_classes.py [hunk 4]
```diff
@@ -180,40 +229,69 @@ def model(self) -> Union[HamiltonianModel, LindbladModel]:
...
 
         Args:
             t_span: Time interval to integrate over.
             y0: Initial state.
+            signals: Specification of time-dependent coefficients to simulate.
+                     If ``dissipator_operators is None``, specify as a list of signals for the
+                     Hamiltonian component, otherwise specify as a tuple of two lists, one
+                     for Hamiltonian components, and one for the ``dissipator_operators``
+                     coefficients.
+            wrap_results: Whether or not to wrap the result arrays in the same class as y0.
```
chriseclectic, line 294: Is this something we want to add if Array is planned to be deprecated?

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -224,97 +302,151 @@ def solve(
 
         Additional Information:
 
-            The behaviour of this method is impacted by the input type of ``y0``:
+        The behaviour of this method is impacted by the input type of ``y0``:
```
chriseclectic, line 305: General thought not for this specific PR: should there be more detailed documentation on how solvers/rotating frames etc work in the `solvers.__init__` documentation?

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -224,97 +302,151 @@ def solve(
...
 
-             * If ``y0`` is an ``Array``, it is passed directly to
-                :func:`~qiskit_dynamics.solve_lmde` as is. Acceptable array shapes are
-                determined by the model type and evaluation mode.
-             * If ``y0`` is a subclass of :class:`qiskit.quantum_info.QuantumState`:
+         * If ``y0`` is an ``Array``, it is passed directly to
+            :func:`~qiskit_dynamics.solve_lmde` as is. Acceptable array shapes are
+            determined by the model type and evaluation mode.
+         * If ``y0`` is a subclass of :class:`qiskit.quantum_info.QuantumState`:
```
chriseclectic, line 310: An idea for this would be to put a table somewhere to make the solver modes easier to see eg along lines of

```
| y0 | Model | yf | Method / Details / Description |
| --- | --- | --- |
| Array, np.ndarray | Any | Same as y0 | solve_lmde ... |
| Statevector | HamiltonianModel | Statevector | ... |
| Statevector | LindbladModel | DensityMatrix | ... |
| DensityMatrix | Hamiltonian, Lindblad | DensityMatrix  | ... |
| Operator | Hamiltonian | Operator | .... |
| Operator | Lindblad | Choi/SuperOp? | ... |
| QuantumChannel | Hamiltonian, Lindlbad | Choi/SuperOp? | ... |
```

chriseclectic (commented):
High level review. Looks good, couple suggestions for docs etc

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -348,3 +480,124 @@ def initial_state_converter(obj: Any) -> Tuple[Array, Type, Callable]:
...
+                y.shape[0], y.shape[1] ** 2, y.shape[1] ** 2
+            )
+            @ y_input
+        )
+    elif (y0_cls is DensityMatrix) and is_lindblad_model_vectorized(model):
+        new_y = y.reshape((len(y),) + y_input.shape, order="F")
+    else:
+        new_y = y
+
+    return new_y
```
brosand, line 508: First of all, don't need to initialize `new_y` as None (unless that's what lint prefers). Also could replace setting `new_y` with a bunch of return statements, not sure which method is preferred.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -348,3 +480,124 @@ def initial_state_converter(obj: Any) -> Tuple[Array, Type, Callable]:
...
+        t_span: Time interval specification.
+        y0: Initial state specification.
+        signals: Signal specification.
+
+    Returns:
+        Tuple: tuple of lists of arguments of the same length, along with a bool specifying whether
+        the arguments specified multiple simulations or not.
+
+    Raises:
+        QiskitError: If the length of each argument is incompatible, or if any singleton
```
brosand, line 536:
```suggestion
        QiskitError: If the length of any arguments are incompatible, or if any singleton
```
Consider ^^

test/dynamics/solvers/test_solver_classes.py
```diff
@@ -67,80 +141,267 @@ def test_validation_override_lindblad(self):
 
         self.assertAllClose(solver.model.hamiltonian_operators, operators)
 
+    def test_incompatible_input_lengths(self):
```
brosand, line 144: Consider testing single length args? make sure they convert to the right size

brosand (commented):
Some small notes left, but overall looks good

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -166,11 +187,39 @@ def __init__(
 
         self._rwa_signal_map = None
         if rwa_cutoff_freq is not None:
```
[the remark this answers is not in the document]
DanPuzzuoli: Correct, will change.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -166,11 +187,39 @@ def __init__(
...
+
+            original_signals = model.signals
+
+            if rwa_carrier_freqs:
```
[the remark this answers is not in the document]
DanPuzzuoli: good catch - changing it.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: True - changing this to not define `new_y` at all, do return statements in the `if` statements, and the final case `new_y = y` is removed and replaced with just a `return y` statement at the end.

test/dynamics/solvers/test_solver_classes.py
DanPuzzuoli: This is actually tested in the `TestSolverListSimulation` class further down in the file. I put a bunch of tests there for making sure various configurations of lists/singletons work.

docs/tutorials/qiskit_pulse.rst
```diff
@@ -119,9 +119,9 @@ rotating frame and perform the rotating wave approximation.
     hamiltonian_solver = Solver(
         static_hamiltonian=drift,
         hamiltonian_operators=operators,
-        hamiltonian_signals=signals,
         rotating_frame=drift,
-        rwa_cutoff_freq=2*w
+        rwa_cutoff_freq=2*w,
+        rwa_carrier_freqs=[w]
```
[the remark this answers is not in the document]
brosand: yeah great my thought as well

test/dynamics/solvers/test_solver_classes.py
brosand: Perfect!

brosand (approved):
LGTM

brosand (dismissed):

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: This is actually for controlling whether the results are wrapped in quantum info types, not whether they're wrapped in `Array`. Since JAX arrays can't be put into quantum info types the final wrapping can't be performed if using JAX. Setting `wrap_results=False` allows you to use JAX to solve while taking advantage of the automatic input state handling by specifying the initial state as a quantum info object.

releasenotes/notes/solver-generalization-ad588a54c830c66f.yaml
DanPuzzuoli: Moved this to `features`, and also added an extra sentence to specify that the simulations are done in a serial loop.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: yeah I agree and was thinking about this as well while making this PR. All of this is actually explained in the `models.__init__` documentation, but given that `Solver` has become the main interface it definitely makes sense to include some version of it there (or even transfer it from the `models` section). I'll make an issue for this.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -224,97 +302,151 @@ def solve(
...
+        t_span_list, y0_list, signals_list, multiple_sims = setup_simulation_lists(
+            t_span, y0, signals
+        )
+
+        all_results = []
+
+        for current_t_span, current_y0, current_signals in zip(t_span_list, y0_list, signals_list):
+
+            # convert types
+            if isinstance(current_y0, QuantumState) and isinstance(self.model, LindbladModel):
```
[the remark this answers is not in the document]
DanPuzzuoli: Added a `_solve` and moved it in there.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -224,97 +302,151 @@ def solve(
...
-            raise QiskitError("""Shape mismatch for initial state y0 and HamiltonianModel.""")
-        if is_lindblad_model_vectorized(self.model) and (
-            y0.shape[0] != self.model.dim**2 or y0.ndim > 2
-        ):
-            raise QiskitError(
-                """Shape mismatch for initial state y0 and LindbladModel
-                                 in vectorized evaluation mode."""
+            results = solver.solve(t_span=t_span, y0=y0, signals=signals)
+
+        The following code block runs three simulations, for different sets of signals,
```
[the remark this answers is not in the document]
DanPuzzuoli: Yeah possibly. I was worried that the description I wrote for doing multiple simulations may not be self-explanatory enough. Maybe this makes more sense at a module level?

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: Yeah agreed - just made this change.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: Is there a better name we can use for this?

chriseclectic: Maybe something like this
```
convert_results: If True, convert returned solver state results to the same class as y0. If False they will be returned in as the native array type used by the specified solver.
```

Might also be good to mention in "Additional information" that for jax solvers it can be beneficial to set this to False

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -180,40 +229,69 @@ def model(self) -> Union[HamiltonianModel, LindbladModel]:
...
-    ) -> OdeResult:
-        r"""Solve the dynamical problem.
-
-        Calls :func:`~qiskit_dynamics.solvers.solve_lmde`, and returns an `OdeResult`
-        object in the style of `scipy.integrate.solve_ivp`, with results
+        self,
+        t_span: Array,
+        y0: Union[Array, QuantumState, BaseOperator],
+        signals: Optional[Union[List[Signal], Tuple[List[Signal], List[Signal]]]] = None,
+        wrap_results: Optional[bool] = True,
```
chriseclectic, line 275:
```suggestion
        wrap_results: bool = True,
```
You dont need optional if its always a bool

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: Made this change (with minor editing to your suggestion).

Also added a paragraph to the Additional Information section after the table indicating that the wrapping can be avoided with `convert_results`, and that using JAX is a time when this will want to be done.

I also added a test cases to `test_solver_classes` to check that this option is operating correctly.

chriseclectic (approved):
LGTM

PR #109, opened 2022-06-06
Implementing Lanczos algorithm as a new solver method

rupeshknn:
### Summary
Lanczos algorithm is an approximate diagonalisation method. It is implemented as an LMDE method and is a considerable speedup compared to `scipy.expm` 

### Details and comments
This PR adds a new fixed-step solver method `lanczos_diag`. This method only works with hermitian generators and works best in sparse evaluation mode. A follow up with a Jax implementation of the same is in progress.


qiskit_dynamics/solvers/lanczos.py
```diff
@@ -0,0 +1,130 @@
...
+    using Lanczos algorithm.
+    Args:
+        array : Array to diagonalize.
+        v_0 : Initial state.
+        k_dim : Dimension of the krylov subspace.
+
+    Returns:
+        q_basis : Basis of the krylov subspace.
+        eigen_values : lowest ``k_dim`` Eigenvalues.
+        eigen_vectors_t : Eigenvectors in both krylov-space.
```
DanPuzzuoli, line 98: The descriptions of `eigen_vectors_t` and `eigen_vectors_a` are unclear.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -273,6 +282,17 @@ def rhs(t, y):
 
     if method == "scipy_expm":
         results = scipy_expm_solver(solver_generator, t_span, y0, t_eval=t_eval, **kwargs)
+    if method == "lanczos_diag":
+        if "sparse" not in generator.evaluation_mode:
```
DanPuzzuoli, line 286: Note that `generator` here could be an arbitrary function, so you can't rely on it being one of the model classes. To raise this warning you'll need to first verify that `generator` is an instance of a `BaseGeneratorModel`. If it isn't you unfortunately won't be able to warn the user.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -273,6 +282,17 @@ def rhs(t, y):
...
+            warn(
+                "lanczos_diag must be used with a generator in sparse mode for better performance.",
+                category=Warning,
+                stacklevel=5,
+            )
+            # raise QiskitError("lanczos_diag must be used with a generator in sparse mode.")
+        if not is_hermitian(1j * solver_generator(1.12)):
```
DanPuzzuoli, line 293: We discussed this before but right now I'm leaning towards not checking this condition since it can't be checked for all time anyway. 

I think the best that can be done is: 
- to check if its a `LindbladModel` or a `GeneratorModel`, and raise an error in those cases, something like "Lanczos solver can only be used for HamiltonianModel or function-based anti-Hermitian generators.".
- Be very clear in the documentation of the solver that it only works for anti-Hermitian generators.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -194,6 +197,12 @@ def solve_lmde(
       size to take. This solver will break integration periods into even
       sub-intervals no larger than ``max_dt``, and solve over each sub-interval via
       matrix exponentiation of the generator sampled at the midpoint.
+    - ``'lanczos_diag'``: A fixed-step matrix-exponential solver similar to ``'scipy_expm'``
+      but uses using Lanczos algorithm. Requires additional kwargs ``max_dt`` and ``k_dim``
+      indicating the maximum step size to take and Krylov subspace dimension, respectively.
+      ``k_dim`` acts an adjustable accuracy parameter and ``k_dim`` < ``model.dim``. Note
+      that the generator must be necessarily hermitian and preferably in sparse evaluation
```
DanPuzzuoli, line 204: the generator must actually be anti-Hermitian.

DanPuzzuoli (commented):
I've left a bunch of individual comments at various points for this initial review. 

Some higher level discussion points: 
- Given that this approach is pretty non-standard I think it's important to document clearly what algorithms are being used and what the function inputs/outputs do so that it will be more intelligible to maintainers of the code. As a result I will nitpick the documentation of the internal functions a little more than I would otherwise.
- You should start to add some test cases - the file `test/dynamics/solvers/test_solver_functions.py` contains tests for methods in `solve_ode` and `solve_lmde`. It's set up so you can test a new method through inheritance. Take a look and let me know if you have questions.
- I think I am going to have to dig in on the point that this solver should work for `y0` being a matrix, not just for vectors. I don't understand the details of the method at this point, and I remember you saying that there are some computational steps you can skip if you assume `y0` is a vector, but my understanding is that some version of this is still doable with `y0` being a matrix. So, I think there are two options:
    - You could generalize whatever cost-saving step you have used for `y0` being a vector to `y0` being a matrix. (It seems like you seed the lanczos diagonalization algorithm with an initial vector, could you instead seed it with an initial subspace by doing gram schmidt on `v_0`? I'm asking this completely naively as I don't understand how the method works.)
    - If there isn't a way to generalize the cost-saving measure, can you make it possible to run a version the algorithm that doesn't utilize this cost saving measure? You could check if `y0` is a vector or a matrix, then run one or the other version.

The last thing is: can you share a notebook in this PR thread demonstrating the observed speed ups, so we can discuss/refine them?

Thanks!

qiskit_dynamics/solvers/lanczos.py
```diff
@@ -0,0 +1,130 @@
...
+        + np.diag(beta[: k_dim - 1], k=1)
+    )
+    q_basis = q_basis[:k_dim]
+    q_basis = q_basis.T
+    return tridiagonal, q_basis
+
+
+def lanczos_eig(array: Union[csr_matrix, np.ndarray], v_0: np.ndarray, k_dim: int):
+    """
+    Finds the lowest k_dim eigenvalues and corresponding eigenvectors of a hermitian array
```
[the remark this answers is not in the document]
rupeshknn: lowest means the smallest algebraic value, which may or may not be largest absolute. Although it typically is.

qiskit_dynamics/solvers/lanczos.py
```diff
@@ -0,0 +1,130 @@
...
+from scipy.sparse import csr_matrix
+
+
+def lanczos_basis(array: Union[csr_matrix, np.ndarray], v_0: np.ndarray, k_dim: int):
+    """Tridiagonalises a hermitian array in a krylov subspace of dimension k_dim
+    using Lanczos algorithm.
+
+    Args:
+        array : Array to tridiagonalise.
+        v_0 : Initial state.
```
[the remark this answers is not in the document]
rupeshknn: v_0 is the inital state given by the user (y0). Although a random vector is usually prefered, for the case of time-evolution as implemented in this PR, this has to be the same as the state that we want to evolve. The cost saving steps that I had mentioned, requries this.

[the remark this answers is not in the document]
rupeshknn: should i rename v_0 to y0?

qiskit_dynamics/solvers/fixed_step_solvers.py
```diff
@@ -100,6 +105,49 @@ def take_step(generator, t0, y, h):
...
+    size no larger than ``max_dt``.
+
+    Args:
+        generator: Generator for the LMDE.
+        t_span: Interval to solve over.
+        y0: Initial state.
+        max_dt: Maximum step size.
+        t_eval: Optional list of time points at which to return the solution.
+        k_dim: Integer which specifies the dimension of Krylov subspace used for
+               lanczos iteration. Acts as an accuracy parameter. ``k_dim < dim(generator)``.
```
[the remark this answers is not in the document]
rupeshknn: I've modified it to raise an error.

qiskit_dynamics/solvers/solver_functions.py
rupeshknn: fixed.

rupeshknn:
Regarding `y0` being a matrix:
When Lanczos is initialised using `y0` as the seed, we can skip computing the matrix exponential since the equation simplifies to something in terms of the Lanczos basis and tridiagonal matrix. 
However, when this is not possible, I could seed using a random vector -> compute the matrix exponential, -> find action on expm on the `y0` matrix. 

This change would affect the speed only a little bit, **but** this also takes a toll on the accuracy. This is because the rows of Q (the basis vectors) are not perfectly orthogonal. 

To mitigate this, one has to perform a QR decomposition which adds to the runtime. However, since we QR decompose, one could make do with a smaller `k_dim` value.
@miamico what do you think?

miamico:
> Regarding `y0` being a matrix: When Lanczos is initialised using `y0` as the seed, we can skip computing the matrix exponential since the equation simplifies to something in terms of the Lanczos basis and tridiagonal matrix. However, when this is not possible, I could seed using a random vector -> compute the matrix exponential, -> find action on expm on the `y0` matrix.
> 
> This change would affect the speed only a little bit, **but** this also takes a toll on the accuracy. This is because the rows of Q (the basis vectors) are not perfectly orthogonal.
> 
> To mitigate this, one has to perform a QR decomposition which adds to the runtime. However, since we QR decompose, one could make do with a smaller `k_dim` value. @miamico what do you think?

I'm still thinking how to deal with an initial matrix instead of a vector. If it's just about batching many initial vectors which are put together into a matrix, couldn't we just do Lanczos on each of the columns (or rows) of the matrix which represent a vector?

qiskit_dynamics/solvers/lanczos.py
```diff
@@ -0,0 +1,130 @@
...
+Module containing Lanczos diagonalization algorithm
+"""
+
+from typing import Union
+import numpy as np
+from scipy.sparse import csr_matrix
+
+
+def lanczos_basis(array: Union[csr_matrix, np.ndarray], v_0: np.ndarray, k_dim: int):
+    """Tridiagonalises a hermitian array in a krylov subspace of dimension k_dim
```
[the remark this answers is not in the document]
rupeshknn: added https://tensornetwork.org/mps/algorithms/timeevo/global-krylov.html as a reference.

rupeshknn:
@DanPuzzuoli, I've made the requested changes. As for rectangular arrays, I've modified the code so that we run a Lanczos iteration for each row, as @miamico suggested. Let me know what you think.

DanPuzzuoli:
> @DanPuzzuoli, I've made the requested changes. As for rectangular arrays, I've modified the code so that we run a Lanczos iteration for each row, as @miamico suggested. Let me know what you think.

Okay awesome, TY. I'm ending early today, but will take a look next week.

qiskit_dynamics/solvers/lanczos.py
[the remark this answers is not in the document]
DanPuzzuoli: Ah cool, thanks. I was imagining this might be in terms of an ordering from "most significant" to "least significant" or something like that.

qiskit_dynamics/solvers/lanczos.py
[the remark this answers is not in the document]
DanPuzzuoli: yeah I think it'll be clearer if you rename it to `y0`.

DanPuzzuoli (commented):
It's really taking shape! I've left some further detailed comments amongst the code for some concrete changes, but I think that it's getting close interface-wise.

I think the next stage is to add more detailed testing: 

- I mentioned this in direct code comments, but for the inherited test cases that you overrode with `pass`, you should instead override with actual tests that are in the spirit of the original test methods (but for anti-hermitian generators, etc). For testing correctness, it's really important to have pseudo-random tests, to avoid baking accidental assumptions into your test cases. Having a test case explicitly testing a `HamiltonianModel` is also important for verifying the functionality in that instance (which is how it will be used at a higher level through the `Solver` class).
- Add a test that checks consistency of `y0` being 1d v.s. 2d.
- Add tests to the `test_solver_function_interface` files that verify the correct errors get thrown in validation checks for this solver. You can use the `assertRaisesRegex` method of the test class to verify that a type of error is raised with a given message. You can see examples of this in that file (though you'll need to merge `main` into your branch to get the correct version of the file, which has been updated since you started your PR).
- Add tests that directly test the functions in `lanczos.py` to verify in various cases that they're performing as expected. Since you're writing the actually core computations from scratch we can't assume their correctness.

Now that you're at this stage, I'm going to set the CI to automatically start running tests on your PR. This can also help to reveal any environment differences you have (e.g. if tests are passing locally but not on CI).

Lastly, with regards to tests, I've started playing around with the solver and have found problems:

### Example 1

```
from qiskit_dynamics import solve_lmde
import numpy as np
from scipy.linalg import expm

iX = -1j * np.array([[0., 1.], [1., 0.]])
gen = lambda t: iX

t_span = [0., np.pi/4]
y0 = np.array([0., 1.])

results = solve_lmde(
    generator=gen, 
    t_span=t_span, 
    y0=y0,
    method='lanczos_diag',
    max_dt=0.1,
    k_dim=1,
)

results.y[-1]
```
yields
```
array([0.+0.j, 1.+0.j])
```
however, the solution is:
```
expm(gen(0) * t_span[-1]) @ y0
```
which gives
```
array([0.        -0.70710678j, 0.70710678+0.j        ])
```
This holds whether I set `k_dim = 1` or `k_dim = 2`.

### Example 2

```
iZ = -1j * np.array([[1., 0., 0.], [0., 0.5, 0.], [0., 0., -1]])

test_gen2 = lambda t: iZ

t_span = [0., np.pi/4]
y0 = np.array([0., 0., 1.])

results2 = solve_lmde(
    generator=test_gen2, 
    t_span=t_span, 
    y0=y0,
    method='lanczos_diag',
    max_dt=0.1,
    k_dim=2,
)
```
ends up raising a the error:
```
/Users/dpuzzuoli/Documents/projects/qiskit-dynamics/qiskit_dynamics/solvers/lanczos.py:62: RuntimeWarning: invalid value encountered in true_divide
  q_basis[[i], :] = projection.T / beta[i - 1]
```
so some kind of division by `0` is happening that probably needs to be handled in a special case.

test/dynamics/solvers/test_solver_functions.py
```diff
@@ -274,6 +274,32 @@ def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
...
+        """Override GeneratorModel test since lanczos only works on anti-hermitan generators."""
+        pass
+
+    def test_pseudo_random_model(self):
+        """Override pseudo_random_model test since lanczos only works on
+        anti-hermitan generators."""
+        pass
+
+    def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
+        print(rhs, y0)
```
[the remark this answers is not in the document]
rupeshknn: sorry, fixed it.

qiskit_dynamics/solvers/fixed_step_solvers.py
```diff
@@ -100,6 +106,55 @@ def take_step(generator, t0, y, h):
...
+    dim = generator(0).shape[0]
+    if k_dim is None:
+        k_dim = dim // 4
+    if k_dim > dim:
+        raise QiskitError(f"k_dim value {k_dim} is greater than dimension of generator {dim}")
+
+    def take_step(generator, t0, y, h):
+        eval_time = t0 + (h / 2)
+        # since qiskt-dynamics generator is -1j*H but lanczos only works on hermitian arrays
+        return lanczos_expm(1j * generator(eval_time), y, k_dim, h)
```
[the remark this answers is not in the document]
rupeshknn: There is another case for keeping it. If `h` is complex for instance, the method will still work. however, if it is multiplied with the generator, then it is no longer anti-hermitian.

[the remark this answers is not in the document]
rupeshknn: As for having a default value, setting a default value at the level of `lanczos_expm` would not have any effect since the high level solver would still need it to be passed. 
On the other hand, setting a default value in `fixed_step_solvers.py`  is probably not a good idea since that would make the signature of  `lanczos_diag`  different from `scipy_expm`

test/dynamics/solvers/test_solver_functions.py
```diff
@@ -274,6 +274,32 @@ def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
         )
 
 
+class Testlanczos_diag(TestSolverMethod):
+    """Test class for lanczos_diag_solver."""
+
+    def test_w_GeneratorModel(self):
```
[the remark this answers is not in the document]
rupeshknn: done

test/dynamics/solvers/test_solver_functions.py
```diff
@@ -274,6 +274,32 @@ def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
...
+        """Override GeneratorModel test since lanczos only works on anti-hermitan generators."""
+        pass
+
+    def test_pseudo_random_model(self):
```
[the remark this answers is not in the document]
rupeshknn: done

rupeshknn:
@DanPuzzuoli 
Thanks for those examples. The `if: break` condition in the loop was supposed to handle these div by zero errors. However, it comes only after one run of the loop. In those examples you mentioned, It hit zero before the loop starts. I've fixed it now by moving the if condition to the top. 
The JAX version (not part of this PR) does not have that `if: break` condition. I might text you on slack when I get to that. 

I believe the CI is not automatically running the tests yet. Can you check that?

I've made all the fixes as suggested. Although I'm not sure about the names I've given to the test functions, let me know if they're okay.

qiskit_dynamics/solvers/fixed_step_solvers.py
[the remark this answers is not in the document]
DanPuzzuoli: yeah makes sense

test/dynamics/solvers/test_fixed_step_solvers.py
```diff
@@ -322,6 +324,75 @@ def solve(self, rhs, t_span, y0, max_dt, t_eval=None):
...
+                    gen, t_span=t_span, y0=self.rand_y0_2d[:, 0], max_dt=0.1, k_dim=8
+                ).y,
+                lanczos_diag_solver(
+                    gen, t_span=t_span, y0=self.rand_y0_2d[:, 1], max_dt=0.1, k_dim=8
+                ).y,
+                lanczos_diag_solver(
+                    gen, t_span=t_span, y0=self.rand_y0_2d[:, 2], max_dt=0.1, k_dim=8
+                ).y,
+            ]
+        ).transpose(1, 2, 0)
```
DanPuzzuoli, line 387: Why are there 3 axes here?

DanPuzzuoli (changes requested):
Good changes. 

The `Testlanczos_diag` class is good. The tests in `TestLanczosDiagSolver` are also good but need to be changed up a bit:
- `TestLanczosDiagSolver` should inherit from `TestFixedStepBase` so that it runs through all of the tests contained there.
    - You'll need to do something similar to `Testlanczos_diag` where you redefine some of the internal matrices/functions constructed in the `setUp` so that the generator is always anti-hermitian.
    - `TestLanczosDiagSolver` doesn't need to inherit from `TestJaxBase` as it isn't doing any JAX testing, and won't need to inherit from `QiskitDynamicsTestCase` once it inherits from `TestFixedStepBase`.
    - This is maybe abit redundant with `Testlanczos_diag` (I'm forgetting what is contained in each test file), but nevertheless more tests are good.
- `TestLanczosDiagSolver` also contains tests for the helper methods in `lanczos.py`. You should move these test cases to a separate file, like `test_lanczos.py`. This will keep the `test_fixed_step_solvers.py` file strictly about the solver functions. (None of the other fixed step solvers have as much complexity, so don't need their own test files for internal machinery.)

I've confirmed that the failure cases in my last review have been fixed - have you added these as test cases? If not, add them as standalone cases in `TestLanczosDiagSolver`. I've found another issue though, which is a variation on one of my last failure cases:

```
from qiskit_dynamics import solve_lmde
import numpy as np
from scipy.linalg import expm

iZ = -1j * np.array([[1., 0., 0.], [0., 0.5, 0.], [0., 0., -1]])

test_gen2 = lambda t: iZ

t_span = [0., np.pi/4]
y0 = np.array([0., 1., 1.])

results2 = solve_lmde(
    generator=test_gen2, 
    t_span=t_span, 
    y0=y0,
    method='lanczos_diag',
    max_dt=0.1,
    k_dim=3,
)
results.y[-1]
```
gives
```
array([0.        +0.j        , 0.65328148-0.27059805j, 0.5       +0.5j       ])
```
but
```
expm(test_gen2(0) * t_span[-1]) @ y0
```
gives
```
array([0.        +0.j        , 0.92387953-0.38268343j, 0.70710678+0.70710678j])
```
Both the same up to a factor of `sqrt(2)`, which is the norm of `y0`. 

It appears that `y0` is being normalized in line 49 of `lanczos.py`. I guess the basis construction algorithm assumes `y0` is normalized? You are going to want to keep track of this normalization somewhere and re-multiply the outputs by these factors. Not sure what the best level to do this at is - maybe change `lanczos_diag` to assume `y0` is normalied, and put the normalization handling in `lanczos_expm`?

For the automatic tests: I think I may just have to manually make the tests run every time. I thought once I set them to run once they'd run automatically, but it may not become automatic until you've had a PR merged.

test/dynamics/solvers/test_fixed_step_solvers.py
DanPuzzuoli: Ah right the whole thing will be 3d, nm.

rupeshknn:
I've made the changes and changed `lanczos_expm` to store the norm and multiply it later. Thanks, @DanPuzzuoli. 
Now I'll add the Jax code to this branch

test/dynamics/solvers/test_lanczos.py
```diff
@@ -0,0 +1,90 @@
...
+    jax_lanczos_eigh,
+    jax_lanczos_expm,
+)
+from ..common import QiskitDynamicsTestCase, TestJaxBase
+
+
+class TestLanczos(QiskitDynamicsTestCase):
+    """Tests for lanczos.py"""
+
+    def lanczos_functions(self):
```
DanPuzzuoli, line 34: This method doesn't need to be here - it can just be part of the `setUp` function. For the JAX version of the class, you can make the setup like:
```
def setUp(self):
     super().setUp()
     self.basis = jax_lanczos_basis
     self.eigh = jax_lanczos_eigh
     self.expm = jax_lanczos_expm
```

I've done this in other places in Dynamics.

qiskit_dynamics/solvers/lanczos.py
```diff
@@ -0,0 +1,291 @@
...
+        y0_norm = jnp.linalg.norm(y0)
+        q_basis, eigen_values, eigen_vectors_t = jax_lanczos_eigh(A, y0 / y0_norm, k_dim)
+        y_dt = (
+            q_basis
+            @ eigen_vectors_t
+            @ (jnp.exp(-1j * scale_factor * eigen_values) * eigen_vectors_t[0, :])
+        ) * y0_norm
+
+    elif y0.ndim == 2:
+        y_dt = [jax_lanczos_expm(A, yi, k_dim, scale_factor) for yi in y0.T]
```
DanPuzzuoli, line 285: I think you either want to turn this loop into a `scan` or use `vmap`, to avoid long compile time issues coming from for loops with JAX.

DanPuzzuoli (changes requested):
The structure of the files/tests is looking great! Aside from the inline comments, one minor required change is:

- Add `'jax_lanczos_diag'` to the list of JAX methods checked in the `is_jax_method` method function in `solver_functions.py`.

For the JAX version there are two main issues:
- Getting `grad` through the lanczos expm to work.
     - One basic question I have here is how will this work given that the exponentiation routine uses `eigh`. I'm almost certain you can differentiate through `eigh`, but I'm just unclear on the actual behaviour guarantees of this (the phases of eigenvectors can change discontinuously, for example).
     - The other problem here you pointed out is that the `while_loop` function does not support gradient computation. There are two ways around this: (1) write a custom rule for differentiating `jax_lanczos_basis` (you can manually supply JAX with the differentiation rules for a given function), or (2) rewrite the `while_loop` as a `scan` of fixed length equal to `k_dim`. I think that while (1) always has the potential to be more efficient, (2) would be a lot easier. The way to make the while-loop into a loop with a fixed number of iterations, is to make the body of the loop first check if the exiting condition has been met: if it hasn't then run the normal body of the loop, but if it has, then do nothing. You can see an example of this in `fixed_step_solver_template_jax`. (Basically it has two nested loops that are implemented with `scan`s. The inner loop could have a different number of iterations dependent on the inputs to the outer loop, which is problematic for JAX compilation. To get around this I set the inner loop to always run for as many iterations as the longest possible inner loop, and did the trick I just described to "do nothing" for the extra iterations.) This trick is used in JAX itself, and I learned about it based on a suggestion from one of the developers.
- Does this solver work if the generator uses the JAX sparse type? There is a sparse type called BCOO in JAX - since this method seems to be best when used with sparse matrices, it'd be good to support this. Take a look at the [JAX sparse documentation](https://jax.readthedocs.io/en/latest/jax.experimental.sparse.html?highlight=sparse) - its possible that if the only thing you're doing with sparse arrays is sparse-dense multiplication then your code will already work for this type, but you'll have to check.

Related to the last point: it may be good to make sure that in your tests you have both sparse and dense tests, to make sure the correct behaviour is locked in for both.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -284,10 +303,39 @@ def rhs(t, y):
 
     if method == "scipy_expm":
         results = scipy_expm_solver(solver_generator, t_span, y0, t_eval=t_eval, **kwargs)
+    if method == "lanczos_diag":
```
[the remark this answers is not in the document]
rupeshknn: I've done it. However, since `is_hermitian` is not jax jitable, that is a seperate test only for the numpy version

qiskit_dynamics/solvers/lanczos.py
rupeshknn: used scan instead of for-loop

rupeshknn:
Hi @DanPuzzuoli, I've got the changes done. However, I implemented the `scan` based looping in a slightly different way than you had suggested with some inspiration from  `fixed_step_solver_template_jax`. This method uses only one `scan`, I guess that's a good thing. The cost is a bunch of loops that return zeros.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -284,6 +310,28 @@ def rhs(t, y):
 
     if method == "scipy_expm":
         results = scipy_expm_solver(solver_generator, t_span, y0, t_eval=t_eval, **kwargs)
+    elif "lanczos_diag" in method:
```
DanPuzzuoli, line 313: This block should be cleaned up/compartmentalized to keep this block of `if/else` statements in `solve_lmde` readable.
- First, I think the check using `is_hermitian` can be removed. I know we've discussed this before but I've generally tried to avoid evaluating the generator outside of solving (I don't believe it's done anywhere else). So long as the fact that this method is restricted to anti-Hermitian matrices is properly documented (it is) I think it's fine to just remove this.
- The other warnings/errors should be moved into a separate function, called something like `_lanczos_validation(generator)`, just to minimize the number of lines appearing in this block of code. This block can then be shortened to:
```
_lanczos_validation(generator)

if method == "lanczos_diag":
    results = ...
elif method == "jax_lanczos_diag":
    results = ...
```
- I also think the logic of the warnings/errors can be simplified as follows. Not a major change, but the error should be raised first if its going to be raised, and I prefer checking if its `not` a HamiltonianModel when raising the error, rather than checking if it's any of the others (in case more are added in the future).
```
if isinstance(generator, BaseGeneratorModel):
    if not isinstance(generator, HamiltonianModel):
          raise QiskitError(...)
    
    if "sparse" not in generator.evaluation_mode:
         warn(...)
```
Also I'd change the warning to say they *should* be used in sparse mode, rather than that they *must*.
- The last question I have related to the warning is: why `stacklevel=5` ? For similar warnings I've used `stacklevel=2` in the past.

rupeshknn: I was just trying out different stacklevels as i wasn't sure which was appropriate. No particular reason for `stacklevel=5` changed it to 2.

rupeshknn:
Hi @DanPuzzuoli 
I fixed all the inline comments.

Regarding using sparse mode for regular tests,
I ran into a bunch of errors for the **NumPy backend** (aka scipy.sparse.csr_matrix).
`sparse` mode doesn't seem compatible with `rotating_frame`.

Here's an example (on the main branch)

```
import numpy as np
from qiskit_dynamics import Solver, Signal

sparse_solver = Solver(static_hamiltonian=np.array([[0,1],[1,0]]),
                       hamiltonian_operators=[np.array([[0,1],[1,0]])],
                       rotating_frame=np.array([[0,1],[1,0]]),
                       evaluation_mode='sparse'
                      )

res = sparse_solver.solve(
    t_span=[0., 50],
    y0=np.array([0,1]),
    signals = [Signal(Array(1.), carrier_freq=5)],
    method="odeint"
)

# TypeError: operand type(s) all returned NotImplemented 
# from __array_ufunc__(<ufunc 'matmul'>, '__call__', 
#                      [<2x2 sparse matrix of type '<class 'numpy.complex128'>'with 4 stored elements in Compressed Sparse Row format>],
#                      Array([[-0.70710678-0.j,  0.70710678-0.j],
#                             [ 0.70710678-0.j,  0.70710678-0.j]], backend='numpy')
#                     ): 'list', 'Array'
```

Things work fine with JAX sparse + `rotating_frame` for the same example.

DanPuzzuoli:
Nice bug find. This is a bit of a weird example that I'll need to think about a little to figure out the proper fix. 

The reason it hasn't been found before is due to the expected usage that if a user sets `evaluation_mode='sparse'`, then their frame operator is already diagonal (and specified as a 1d array). 

For a bit of background: A user can specify the frame operator as a 1d array as a way of explicitly telling Dynamics that it's already diagonal. The main reason this is "expected usage" for sparse evaluation mode is that internally Dynamics always represents all operators in a basis in which the frame operator is diagonal - if it isn't diagonal, it has to change the basis of all other operators. If the operators are sparse, you don't want to do a change of basis (which could ruin sparsity), so specifying the frame operator as a 1d array is something we put in to allow the user a pathway to prevent Dynamics from trying to do any basis changes.

That all being said, everything _should_ still work if the frame operator is not given as a 1d array, but it seems that we've really baked the "expected usage" into our tests, so we've missed the bug so far. I _think_ this will be a quick fix, and will get back to you.

DanPuzzuoli:
Okay #118 should solve this issue. The example you provided is working (though you need to change the `method` to something like `'RK4'` rather than `'odeint'`.

DanPuzzuoli:
Didn't look at the tests again but just scanned through everything else and looks good. Let me know if you get the tests working, and I'll do a final holistic review.

rupeshknn:
Thanks for the quick resolution. The tests are working.

test/dynamics/solvers/test_solver_functions_interface.py
```diff
@@ -42,7 +42,7 @@ def test_method_does_not_exist(self):
             solve_ode(lambda t, y: y, t_span=[0.0, 1.0], y0=np.array([1.0]), method="notamethod")
 
 
-class Testsolve_lmde_exceptions(QiskitDynamicsTestCase):
+class Testsolve_lmde_exceptions(QiskitDynamicsTestCase, TestJaxBase):
```
DanPuzzuoli, line 45: I must have missed this before but this test class shouldn't be importing `TestJaxBase`. As JAX is an optional import, we run tests with and without JAX installed. The procedure for testing things without JAX installed will skip any test classes that import `TestJaxBase`, so updating this class to inherit from it is actually _removing_ tests from Dynamics.

Will make a suggestion in the top level review message.

qiskit_dynamics/solvers/fixed_step_solvers.py
```diff
@@ -100,7 +109,83 @@ def take_step(generator, t0, y, h):
...
+               always be less than or equal to dimension of generator.
+
+    Returns:
+        OdeResult: Results object.
+
+    Raises:
+        QiskitError: If ``k_dim`` is greater than dimension of generator.
+    """
+
+    dim = generator(0).shape[0]
```
DanPuzzuoli, line 150: This should be `generator(t_span[0])`, not `generator(0)`.

qiskit_dynamics/solvers/lanczos.py
```diff
@@ -0,0 +1,250 @@
...
+        A : Array to exponentiate. Must be anti-hermitian.
+        y0 : Initial state.
+        k_dim : Dimension of the krylov subspace.
+        scale_factor : Maximum step size.
+
+    Returns:
+        y_dt : Action of matrix exponential on state.
+
+    Raises:
+        ValueError : If ``y0`` is not 1d or 2d
```
DanPuzzuoli, line 133:
```suggestion
        ValueError : If ``y0`` is not 1d or 2d.
```

DanPuzzuoli (changes requested):
Made some small documentation suggestions. I tried to add periods to the end of every sentence in the documentation: that's the style we're trying to aim for in Dynamics.

I made a comment on this, but I just realized `TestJaxBase` was added to the inheritance list for `Testsolve_lmde_exceptions`. I have a suggestion that will remove the need to test a JAX pathway for these exceptions separately:
- Remove `TestJaxBase` from the inheritance list for `Testsolve_lmde_exceptions`.
    - Also remove associated JAX tests.
- Move the validation checks for generator dimension and `y0` having `ndim >= 3` out of both `lanczos_diag_solver` and `jax_lanczos_diag_solver` and into `_lanczos_validation`. This will bring all of the validation checks for both solvers into a single place, removing the need to separately test the JAX versions of these tests. (It will also reduce the number of lines of code.)
    - You'll need to add `y0` and `t_span[0]` to the arguments of `_lanczos_validation`.
    - You'll need to make `lanczos_diag_solver` and `jax_lanczos_diag_solver` explicitly assume `y0` is 1d or 2d, i.e. do `if y0.ndim == 1:`, then `else:` (assuming `y0.ndim==2`. Also add this assumption to the `lanczos_diag_solver` documentation for `y0`.
    - One more minor thing: add a period in the `y0` dimension warning, i.e. change it to `"y0 must be 1d or 2d."`.

qiskit_dynamics/solvers/fixed_step_solvers.py
rupeshknn: Good catch! Thanks.

DanPuzzuoli (approved):
LGTM!

PR #112, opened 2022-06-07
Small diffrax changes

### LINKED ISSUE #110: Slight fixes for diffrax
Several small changes needed:

- Change diffrax solver handling of `t_eval` so that it is not used in the final call to `odeResult`, causing an issue if t_eval is `None`
- Add handling for the diffrax solver in `solve_lmde`
- Add tests for `solve_lmde`'s delegation to `solve_ode`, basic cases for each solver to confirm the pipeline works
- Add diffrax solver to the `solve_ode` docs

brosand:
### Summary

Closes #110 by adding several small diffrax changes

### Details and comments

- Remove `t_eval` from `diffrax_solver` output, using `result.t` instead
- adding an option in `solve_lmde` to use diffrax solvers by passing the instance of the solver into `solve_ode`
- adding tests for calling `solve_ode` through `solve_lmde` for all solvers that are not lmde solvers
- adding diffrax solver note to the `solve_ode` docs


brosand:
Two notes to discuss:

- the RK4 tests do not have the property `is_ode_method`, but they call `solve_ode`. Thus they are running the `test_ode_method`, should I change this?
- In order to add an lmde test for the ode solvers (to test the pipeline), I added in separate `solve_lmde` methods to the ode solver tests. Just leaving a note to discuss other potential implementations.

qiskit_dynamics/solvers/diffrax_solver.py
```diff
@@ -81,6 +81,8 @@ def diffrax_solver(
 
     if t_eval is not None:
         kwargs["saveat"] = SaveAt(ts=t_eval)
+    # else:
```
DanPuzzuoli, line 84: Need to remove these comments.

qiskit_dynamics/solvers/solver_functions.py
```diff
@@ -94,6 +94,7 @@ def solve_ode(
     - ``'jax_RK4'``: JAX backend implementation of ``'RK4'`` method.
     - ``'jax_odeint'``: Calls ``jax.experimental.ode.odeint`` variable step
       solver.
+    - :class:`AbstractSolver` instance of any Diffrax solver.
```
DanPuzzuoli, line 97: To make this more consistent with other lines, maybe something like:

```
- ``diffrax.diffeqsolve`` - a JAX solver function, called by passing ``method`` as a valid ``diffrax.solver.AbstractSolver`` instance. Requires the ``diffrax`` library.
```

test/dynamics/solvers/test_diffrax_DOP5.py
```diff
@@ -41,6 +41,23 @@ def simple_rhs(t, y):
...
+        results = diffrax_solver(
+            self.simple_rhs,
+            t_span,
+            y0,
+            method=Dopri5(),
+            t_eval=None,
+            stepsize_controller=stepsize_controller,
+        )
+
+        self.assertAllClose(results.t, t_span[-1])
```
DanPuzzuoli, line 59: diffrax only returns the state at the final time if `t_eval` isn't specified?

DanPuzzuoli (changes requested):
Looks mostly good.

I think instead of defining a new method `solve_lmde` for the subclasses of `TestSolverMethod`, you could modify the `solve` method to take in an additional optional argument that determines whether it calls `solve_ode` or `solve_lmde`, and then use that.

e.g.

```
def solve(..., solver_func=solve_ode, ...):
     return solver_func(...)
```

test/dynamics/solvers/test_diffrax_DOP5.py
brosand: Yes, the default for diffrax is to just return the final time

test/dynamics/solvers/test_solver_functions.py
```diff
@@ -234,25 +234,71 @@ def func(a):
 class TestRK4(TestSolverMethod):
     """Test class for RK4_solver."""
 
-    def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
-        return solve_ode(
-            rhs=rhs, t_span=t_span, y0=y0, method="RK4", t_eval=t_eval, max_dt=0.001, **kwargs
-        )
+    def solve(self, rhs, t_span, y0, t_eval=None, solver_func="ode", **kwargs):
```
DanPuzzuoli, line 237: For this `solver_func` should just straight up by the callable `solve_lmde` or `solve_ode`, then for the body of the function you can do:

```
     return solver_func(
         rhs,
         t_span=t_span,
         y0=y0,
         method="RK4",
         t_eval=t_eval,
         max_dt=0.001,
         **kwargs,
     )
```

test/dynamics/solvers/test_solver_functions.py
```diff
@@ -234,25 +234,71 @@ def func(a):
...
 
 
 class Testjax_RK4(TestSolverMethodJax):
     """Test class for jax_RK4_solver."""
 
-    def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
-        return solve_ode(
-            rhs=rhs, t_span=t_span, y0=y0, method="jax_RK4", t_eval=t_eval, max_dt=0.001, **kwargs
-        )
+    # def solve(self, rhs, t_span, y0, t_eval=None, **kwargs):
```
DanPuzzuoli, line 267: Delete this commented out code.

DanPuzzuoli (changes requested):
I made one comment about one of the `solve` functions, but it applies to all of them.

DanPuzzuoli (changes requested):
Almost there, still needs to be simplified though.

DanPuzzuoli (approved):
LGTM

PR #114, opened 2022-06-16
Adding schedule solving functionality to Solver

### LINKED ISSUE #106: Incorporate model decomposition, rotating frames, and RWA description in solvers documentation
`Solver` is emerging as the main higher level interface we are promoting, but the module API docs describing it in `solvers.__init__` are pretty sparse. They don't describe how the internal models are being constructed, or how they're being transformed by rotating frames or the RWA. 

All of this information is currently contained in the `models.__init__` file. The original idea was that a user could go there to read about this, but with the `Solver` API becoming the dominant mode of interaction with the package it makes sense to either reproduce this in `solvers.__init__`, move it there, or something else.

DanPuzzuoli:
### Summary

This PR adds the ability to pass a `schedule`, or a list of `schedule`s as the `signals` argument in `Solver.solve`. Internally, the schedules are converted into `DiscreteSignal` objects, and the model is then simulated as usual.

From a user-perspective, the only other main change to enable this is the addition of arguments to the `Solver` constructor to enable interpretation of schedules:
- `hamiltonian_channels` - the channels for the Hamiltonian components
- `dissipator_channels` - the channels for the time-dependent dissipator terms
- `channel_carrier_freqs` - the analog carrier frequencies for the channels
- `dt` - the sample width

If any of the above are specified, a series of validation checks occur to ensure that they are consistent with the model information, and that they provide a complete description of how to interpret pulse schedules (e.g. if `dt` is missing an error will be raised).

Incidentally, closes #106 as it was natural to also update other aspects of the `Solver` documentation.

### Details and comments

Other updates:
- `Solver` class documentation 
    - Adds description of how to configure to simulate pulse schedules.
    - Adds more detail about the models/transformations directly into the `Solver` documentation.
- Updates the qiskit-pulse simulation tutorial to show how to directly simulate schedules with `Solver`.
- Adds validation that the pulse schedule simulation configuration data is complete and correct.
- Many tests added comparing pulse schedule simulation to simulation of manually constructed `DiscreteSignal` objects.

### Still to do and review questions

Questions:
- Currently the information for simulating pulse schedules is all being stored in private attributes. Should we add properties to the class for these? E.g. `Solver.hamiltonian_channels`, `Solver.dissipator_channels`, etc

Still to do:
- Add reno file
- Add automatic jitting in `Array.default_backend() == 'jax'`
- The `solve` method now has a huge block of code for managing the schedule -> signal conversion - it may be good to put this into a standalone function that can be separately tested.

### DIFF
--- a/docs/tutorials/qiskit_pulse.rst
+++ b/docs/tutorials/qiskit_pulse.rst
@@ -1,8 +1,8 @@
-Connecting Qiskit Pulse with Qiskit Dynamics
-============================================
+Simulating Qiskit Pulse Schedules with Qiskit Dynamics
+======================================================
 
-This tutorial shows a simple model of a qubit used to demonstrate how
-Qiskit Dynamics can simulate the time evolution of a pulse schedule. The
+This tutorial shows how to use Qiskit Dynamics to simulate a Pulse schedule
+with a simple model of a qubit. The
 qubit is modeled by the drift hamiltonian
 
 .. math::
@@ -25,8 +25,8 @@ ground state we expect that this second pulse will not have any effect
 on the qubit. This situation is simulated with the following steps:
 
 1. Create the pulse schedule
-2. Convert the pulse schedule to a Signal
-3. Create the system model
+2. Converting pulse schedules to Signals
+3. Create the system model, configured to simulate pulse schedules
 4. Simulate the pulse schedule using the model
 
 1. Create the pulse schedule
@@ -69,7 +69,7 @@ First, we use the pulse module in Qiskit to create a pulse schedule.
 2. Convert the pulse schedule to a ``Signal``
 ---------------------------------------------
 
-To use Qiskit Dynamics we must convert the pulse schedule to instances
+Qiskit Dynamics has functionality for converting pulse schedule to instances
 of ``Signal``. This is done using the pulse instruction to signal
 converter ``InstructionToSignals``. This converter needs to know the
 sample rate of the arbitrary waveform generators creating the signals,
@@ -100,8 +100,11 @@ virtual ``Z`` gate is applied.
 3. Create the system model
 --------------------------
 
-We now setup the Hamiltonian with the signals as drives, enter the
-rotating frame and perform the rotating wave approximation.
+We now setup a ``Solver`` instance with the desired Hamiltonian information,
+and configure it to simulate pulse schedules with. This requires specifying
+which channels act on which operators, channel carrier frequencies, and sample width ``dt``.
+Additionally, we setup this solver in the rotating frame and perform the
+rotating wave approximation.
 
 .. jupyter-execute::
 
@@ -120,8 +123,10 @@ rotating frame and perform the rotating wave approximation.
         static_hamiltonian=drift,
         hamiltonian_operators=operators,
         rotating_frame=drift,
-        rwa_cutoff_freq=2*w,
-        rwa_carrier_freqs=[w]
+        rwa_cutoff_freq=2 * 5.0,
+        hamiltonian_channels=['d0'],
+        channel_carrier_freqs={'d0': w},
+        dt=dt
     )
 
 4. Simulate the pulse schedule using the model
@@ -136,7 +141,7 @@ In the last step we perform the simulation and plot the results.
     # Start the qubit in its ground state.
     y0 = Statevector([1., 0.])
 
-    %time sol = hamiltonian_solver.solve(t_span=[0., 2*T], y0=y0, signals=signals, atol=1e-8, rtol=1e-8)
+    %time sol = hamiltonian_solver.solve(t_span=[0., 2*T], y0=y0, signals=xp, atol=1e-8, rtol=1e-8)
 
 
 .. jupyter-execute::
--- a/qiskit_dynamics/solvers/solver_classes.py
+++ b/qiskit_dynamics/solvers/solver_classes.py
@@ -52,16 +52,60 @@ from .solver_utils import is_lindblad_model_vectorized, is_lindblad_model_not_ve
 
 
 class Solver:
-    """Solver class for simulating both Hamiltonian and Lindblad dynamics, with high
+    r"""Solver class for simulating both Hamiltonian and Lindblad dynamics, with high
     level type-handling of input states.
 
-    Given the components of a Hamiltonian and optional dissipators, this class will
-    internally construct either a :class:`HamiltonianModel` or :class:`LindbladModel`
-    instance.
+    If only Hamiltonian information is provided, this class will internally construct
+    a :class:`~qiskit_dynamics.models.HamiltonianModel` instance, and simulate the model
+    using the Schrodinger equation :math:`\dot{y}(t) = -iH(t)y(t)`
+    (see the :meth:`~qiskit_dynamics.solvers.Solver.solve` method documentation for details
+    on how different initial state types are handled).
+    :class:`~qiskit_dynamics.models.HamiltonianModel` represents a
+    decomposition of the Hamiltonian of the form:
+
+    .. math::
+
+        H(t) = H_0 + \sum_i s_i(t) H_i,
+
+    where :math:`H_0` is the static component, the :math:`H_i` are the operator part
+    of the time-dependent part of the Hamiltonian, and the :math:`s_i(t)` are the
+    time-dependent signals, specifiable as either :class:`~qiskit_dynamics.signals.Signal`
+    objects, or constructed from Qiskit Pulse schedules if :class:`~qiskit_dynamics.solvers.Solver`
+    is configured for Pulse simulation (see below).
+
+    If dissipators are specified as part of the model, then a
+    :class:`~qiskit_dynamics.models.LindbladModel` is constructed, and simulations are performed
+    by solving the Lindblad equation:
+
+    .. math::
+
+        \dot{y}(t) = -i[H(t), y(t)] + \mathcal{D}_0(y(t)) + \mathcal{D}(t)(y(t)),
+
+    where :math:`H(t)` is the Hamiltonian part, specified as above, and :math:`\mathcal{D}_0`
+    and :math:`\mathcal{D}(t)` are the static and time-dependent portions of the dissipator,
+    given by:
+
+    .. math::
+
+        \mathcal{D}_0(y(t)) = \sum_j N_j y(t) N_j^\dagger
+                                      - \frac{1}{2} \{N_j^\dagger N_j, y(t)\},
+
+    and
+
+    .. math::
+
+        \mathcal{D}(t)(y(t)) = \sum_j \gamma_j(t) L_j y(t) L_j^\dagger
+                                  - \frac{1}{2} \{L_j^\dagger L_j, y(t)\},
+
+    with :math:`N_j` the static dissipators, :math:`L_j` the time-dependent dissipator
+    operators, and :math:`\gamma_j(t)` the time-dependent signals
+    specifiable as either :class:`~qiskit_dynamics.signals.Signal`
+    objects, or constructed from Qiskit Pulse schedules if :class:`~qiskit_dynamics.solvers.Solver`
+    is configured for Pulse simulation (see below).
 
     Transformations on the model can be specified via the optional arguments:
 
-    * ``rotating_frame``: Transforms the model into a rotating frame. Note that
+    * ``rotating_frame``: Transforms the model into a rotating frame. Note that the
       operator specifying the frame will be substracted from the ``static_hamiltonian``.
       If supplied as a 1d array, ``rotating_frame`` is interpreted as the diagonal
       elements of a diagonal matrix. See :class:`~qiskit_dynamics.models.RotatingFrame` for details.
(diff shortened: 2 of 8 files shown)


qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -40,28 +42,84 @@
...
 
-    Given the components of a Hamiltonian and optional dissipators, this class will
-    internally construct either a :class:`HamiltonianModel` or :class:`LindbladModel`
-    instance.
+    If only Hamiltonian information is provided, this class will internally construct
+    a :class:`~qiskit_dynamics.models.HamiltonianModel` instance, and simulate the model
+    using the Schrodinger equation :math:`\dot{y}(t) = -iH(t)y(t)`
+    (see the :meth:`~qiskit_dynamics.solvers.Solver.solve` method documentation for details
+    on how different initial state types are handled).
+    :class:`~qiskit_dynamics.models.HamiltonianModel` represents a
```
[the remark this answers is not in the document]
eggerdj: Same goes for other occurrences below.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -147,8 +235,79 @@ def __init__(
...
+                hamiltonian_channels = [chan.lower() for chan in hamiltonian_channels]
+                for chan in hamiltonian_channels:
+                    if chan not in all_channels:
+                        all_channels.append(chan)
+                if hamiltonian_operators is None or len(hamiltonian_operators) != len(
+                    hamiltonian_channels
+                ):
+                    raise QiskitError(
+                        """hamiltonian_channels must have same length as hamiltonian_operators"""
+                    )
```
eggerdj, line 265: Nitpick but you could check first and then do the for loop.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -147,8 +235,79 @@ def __init__(
...
+        # set pulse specific information if specified
+        self._hamiltonian_channels = None
+        self._dissipator_channels = None
+        self._all_channels = None
+        self._channel_carrier_freqs = None
+        self._dt = None
+        self._schedule_converter = None
+
+        if any([dt, channel_carrier_freqs, hamiltonian_channels, dissipator_channels]):
+            all_channels = []
```
eggerdj, line 253: Would a set be better here?

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -271,7 +445,9 @@ def solve(
         self,
         t_span: Array,
         y0: Union[Array, QuantumState, BaseOperator],
-        signals: Optional[Union[List[Signal], Tuple[List[Signal], List[Signal]]]] = None,
+        signals: Optional[
+            Union[List[Schedule], List[Signal], Tuple[List[Signal], List[Signal]]]
```
eggerdj, line 449: Why `List[Schedule]`? Could we not have everything in one `Schedule`?

eggerdj: For a list of simulations?

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -396,99 +575,178 @@ def solve(
...
+        all_results = []
+        for t_span, y0, signals in zip(t_span_list, y0_list, signals_list):
 
-        # validate types
-        if (y0_cls is SuperOp) and is_lindblad_model_not_vectorized(self.model):
-            raise QiskitError(
-                """Simulating SuperOp for a LindbladModel requires setting
-                vectorized evaluation. Set LindbladModel.evaluation_mode to a vectorized option.
-                """
+            if isinstance(signals, (Schedule, ScheduleBlock)):
```
eggerdj, line 620: I don't think we should be allowing `ScheduleBlock` here as it may not have the concept of pulse instructions played at an absolute time. This is why the converter only allows `Schedule`.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -396,99 +575,178 @@ def solve(
...
+        y0_list: List[Union[Array, QuantumState, BaseOperator]],
+        schedule_list: List[Schedule],
+        convert_results: bool = True,
+        **kwargs,
+    ) -> List[OdeResult]:
+        """Run a list of schedule simulations utilizing JAX compilation.
+        The jitting strategy is to define a function whose inputs are t_span, y0 as an array, the
+        samples for all channels in a single large array, and other initial state information.
+        To avoid recompilation for schedules with samples of different lengths, all are padded
+        to be the length of the schedule with the max duration.
```
eggerdj, line 652: Is this ok?

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -396,99 +575,178 @@ def solve(
...
+        all_results = []
+        for t_span, y0, sched in zip(t_span_list, y0_list, schedule_list):
+            # setup initial state
+            y0, y0_input, y0_cls, state_type_wrapper = validate_and_format_initial_state(
+                y0, self.model
+            )
+
+            # setup array of all samples
+            if isinstance(sched, ScheduleBlock):
+                sched = block_to_schedule(sched)
```
eggerdj, line 705: looks good.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -585,6 +903,12 @@ def setup_simulation_lists(
     elif isinstance(signals, list) and isinstance(signals[0], tuple):
         # multiple lindblad
         multiple_sims = True
+    elif isinstance(signals, (Schedule, ScheduleBlock)):
```
eggerdj, line 906: Also, type hint only says `Schedule`. Is that ok?

eggerdj (commented):
Here are some comments, overall this looks like a nice convenience to avoid requiring users to have to deal with the converter.

docs/tutorials/qiskit_pulse.rst
```diff
@@ -136,7 +141,7 @@ In the last step we perform the simulation and plot the results.
     # Start the qubit in its ground state.
     y0 = Statevector([1., 0.])
 
-    %time sol = hamiltonian_solver.solve(t_span=[0., 2*T], y0=y0, signals=signals, atol=1e-8, rtol=1e-8)
+    %time sol = hamiltonian_solver.solve(t_span=[0., 2*T], y0=y0, signals=xp, atol=1e-8, rtol=1e-8)
```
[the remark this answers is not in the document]
DanPuzzuoli: Added a couple sentences explaining that one can either pass `xp` directly to `solve` (as we have configured the solver with pulse information), or equivalently pass `signals` as generated by the converter. 

I think for someone who is only interested in directly simulating the schedules as given, without doing any computation on the signals themselves (e.g. signal processing), this PR makes it so that there probably isn't a need to ever directly use the signal converter anymore. I agree though we should point out the different options.

qiskit_dynamics/solvers/solver_classes.py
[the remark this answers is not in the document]
DanPuzzuoli: Awesome, didn't know you could do that. The long form way replicating the import paths was the only way I knew how to get the linking to work.

Just changed all of the links in this file to be as you suggest.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: yeah good catch - moved this check to before the loop.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: Just tried changing to using a `set` but errors popped up as some of the entries are `dict` or `list` instances, which are unhashable, and so can't be in a `set`.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: Yeah it's for specifying a list of simulations. 

This is the main reason why the "specifying a list of simulations to `Solver.solve`" functionality was introduced. In all of the prior focused-on use cases for dynamics (optimization) it was natural to only specify a single simulation at a given time, but for pulse simulation there needs to be some layer at which looping takes place (as I believe it is common to submit a list of simulations?). We decided to put it here, ultimately because there is an opportunity to control JAX compilation when simulating pulses, so it's nice to have it accessible even at this layer of the interface.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: `ScheduleBlock` instances are ultimately converted to `Schedule`s before simulation is attempted. Do you think it makes sense to do this? I got the impression that users wouldn't manually be converting `ScheduleBlock` instances into `Schedule`s, so we should do it automatically here.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: I think yes, but I understand the concern. If, in a single call, a user is simulating schedules with very different durations, it will be wasteful memory-wise for the shorter schedules. 

That being said, to utilize JAX compilation, we need some way of defining a simulation function that takes in arrays of fixed shape. To avoid dealing with messy looping that will be very non-trivial in JAX, the easiest first approach is to just convert the envelopes for the entire schedule into one large array. To mitigate the memory waste for the shorter schedules, I've set it up so that the combined sample array for each schedule is only being constructed/stored right when that schedule is about to be simulated. So:
- It only needs to store the combined sample array for a single schedule at any given time.
- The size of this sample array is determined by the schedule with the longest duration, so it's at least never requiring more memory than the worst-case schedule in the list.

Perhaps in the future we'll need to come up with something more clever with less memory redundancy, but given the difficulty of working the JAX limitations, for now I think it's good enough as an initial version, especially given that even this simplistic approach will produce major speed ups over the existing pulse simulator.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: What do you think? I figured the type hint for the `signals` argument was already getting kind of crazy, so thought using `Schedule` for both `Schedule` and `ScheduleBlock` was okay? It's definitely not the most formal choice, but also the size of this type hint is growing to the point that it's somewhat unreadable.

I'm happy to either leave it as is, or change it, depending on your thoughts on it.

DanPuzzuoli:
In addition to the above comments, due to errors introduced by a JAX update, I've also added the ability to specify the tolerance used to check `allclose` in some of the test cases, and increased the tolerance of some tests.

Long story short:
- There has been a JAX update since the last time the CI tests were run, which caused some test cases to start failing.
- Turns out the test cases were failing because:
    1. The test cases assumed *almost* exact equality of two different results of solving ODEs.
    2. For some reason the JAX update introduced slightly different rounding errors in some computations. These rounding errors are enough to cause runs of ODE solvers to diverge. While this is totally fine as the divergence is still within the tolerances of the ODE solver, it did lead to failure of the overly-stringent tests.

I've fixed this by additionally relaxing the tolerance of the checks in the affected tests. 

Lesson learned: This has made me realize that the assumptions that the two different setups of the RHS are numerically identical is unsustainable. While it can work for a particular version of this package or its dependency versions, the tests can end up breaking even if reasonable, unproblematic changes are made to functions/packages that the impact the test.

qiskit_dynamics/solvers/solver_classes.py
eggerdj: Alright, I think that is fine. If a user is providing schedules with different durations then said user user think hard about what he/she is doing. That being said I think the sentence is a bit misleading. The samples all have the same length right? I.e. `dt` is the same across schedules. It's more that the schedules have a different duration?
```suggestion
        To avoid recompilation for schedules with a different number of samples, i.e. a different duration,
        all schedules are padded to be the length of the schedule with the max duration.
```

qiskit_dynamics/solvers/solver_classes.py
eggerdj: Looking at this again I think it makes sense to do this conversion for the user. But I expect that this happens in only one place. Is that the case now?

qiskit_dynamics/solvers/solver_classes.py
eggerdj: I am assuming that the user specifies the schedules at only one point. Here they can be `Schedule` or `ScheduleBlock` and we explicitly write out the type hint. Then, we convert `ScheduleBlock` to `Schedule` and only ever deal with `Schedule` going forward. It think that's more or less how you have it now?

DanPuzzuoli: I've just:
- Added `ScheduleBlock` to the `solve` type hint. 
- Changed the `Solver.solve` method to check for `ScheduleBlock`s and convert them to `Schedule`s right at the beginning, and modified all subsequent code to only work with `Schedule`s.

This is good - makes the rest of the code cleaner. I think I had originally been trying to do the conversion as late as possible, but now that I think about it I can't think of any good reason to do that.

eggerdj (approved):
This looks good to me.

PR #115, opened 2022-06-29
Refactor helper function for setting up list simulation

DanPuzzuoli:
### Summary

Refactors some of the logic of the internal function ``setup_simulation_lists`` in ``solver_classes.py`` into a function ``setup_args_lists`` in ``solver_utils.py``, so that it can be reused elsewhere (e.g. it is needed in #73 for setting up list simulation in the new solvers introduced there).

### Details and comments

``setup_simulation_lists`` supports the functionality for allowing a user to pass one or more of ``t_span``, ``y0``, and ``signals`` as lists to ``Solver.solve``, so that multiple simulations can be run with a single call. This consists of two parts:
- Checking whether each of those arguments are "single instances" or lists of single instances (e.g. is ``t_span`` specifying one interval, or a list of intervals?), and turning them all into lists regardless.
- Checking the compatibility of the argument lengths, then setting all to the same length (repeating single instances of necessary) 

The first part of the above is naturally tied to the specifics of the arguments ``t_span``, ``y0``, and ``signals``, however the second part is not, and this is what has been refactored into ``setup_args_lists`` to work for an arbitrary number of input arguments. This function takes in a list of arguments, a list of names of those arguments, and a list of functions for validating whether or not the argument is a "valid single input" or a list of "valid single inputs". It applies these functions to all of the arguments, and performs the second part above.

This allows the user to construct more functions like ``setup_simulation_lists`` by only needing to supply the list conversion/validation functions.

### DIFF
--- a/qiskit_dynamics/solvers/solver_classes.py
+++ b/qiskit_dynamics/solvers/solver_classes.py
@@ -45,7 +45,7 @@ from qiskit_dynamics.array import Array
 from qiskit_dynamics.dispatch.dispatch import Dispatch
 
 from .solver_functions import solve_lmde
-from .solver_utils import is_lindblad_model_vectorized, is_lindblad_model_not_vectorized
+from .solver_utils import is_lindblad_model_vectorized, is_lindblad_model_not_vectorized, setup_arg_lists
 
 
 class Solver:
@@ -392,7 +392,7 @@ class Solver:
                 stacklevel=2,
             )
 
-        t_span_list, y0_list, signals_list, multiple_sims = setup_simulation_lists(
+        [t_span_list, y0_list, signals_list], multiple_sims = setup_simulation_lists(
             t_span, y0, signals
         )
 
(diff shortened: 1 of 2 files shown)


qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -547,87 +551,63 @@ def format_final_states(y, model, y_input, y0_cls):
...
+        elif isinstance(signals, list) and isinstance(signals[0], (list, SignalList)):
+            # multiple Hamiltonian signals lists
+            was_list = True
+        elif isinstance(signals, SignalList) or (
+            isinstance(signals, list) and not isinstance(signals[0], (list, SignalList))
+        ):
+            # single Hamiltonian signals list
+            signals = [signals]
+        else:
+            raise QiskitError("Signals specified in invalid format.")
```
brosand, line 581: You could combine some of these, but I like it better as is, I think it's clearer visually

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -547,87 +551,63 @@ def format_final_states(y, model, y_input, y0_cls):
...
+        elif isinstance(signals, list) and isinstance(signals[0], tuple):
+            # multiple lindblad
+            was_list = True
+        elif isinstance(signals, list) and isinstance(signals[0], (list, SignalList)):
+            # multiple Hamiltonian signals lists
+            was_list = True
+        elif isinstance(signals, SignalList) or (
+            isinstance(signals, list) and not isinstance(signals[0], (list, SignalList))
+        ):
+            # single Hamiltonian signals list
```
brosand, line 578: Is there no case where we pass in a hamiltonian signal not as a list?

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -547,87 +551,63 @@ def format_final_states(y, model, y_input, y0_cls):
...
+            raise QiskitError("Signals specified in invalid format.")
 
-    if not isinstance(y0, list):
-        y0 = [y0]
-    else:
-        multiple_sims = True
+        return signals, was_list
 
-    t_span_ndim = nested_ndim(t_span)
+    def y0_to_list(y0):
```
brosand, line 585: the three `to_list` functions can be define outside of `setup_simulation_lists`

brosand (commented):
Looks pretty good. Just a couple notes. I looked through in detail without much criticism. 
With regards to the `to_list` functions, I suggested moving them outside of the main function.

qiskit_dynamics/solvers/solver_classes.py
DanPuzzuoli: yeah I think that's a good idea, I'll do that

DanPuzzuoli: Actually with this change I'm able to get rid of the `setup_simulation_lists`, as it now contains only the call to `setup_args_lists`. I've changed the call to `setup_simulation_lists` to a call directly to `setup_args_lists`, which seems cleaner.

PR #116, opened 2022-06-30
Changing all assertRaises to assertRaisesRegex in test cases

DanPuzzuoli:
### Summary

Changes all unittest calls to `assertRaises` to `assertRaisesRegex`, which allows for more concise validation that a particular error type is raises with a particular message.

### Details

There were many instances throughout the tests where `assertRaises` was used, and a followup line would check that the message contained a certain string. These were written before I knew about `assertRaisesRegex`. Minor, but cleaner, and is the way thing should be done going forward.


brosand (approved):
LGTM

PR #117, opened 2022-08-02
Delaying diffrax import as late as possible

DanPuzzuoli:
### Summary

Currently, diffrax is raising a JAX deprecation warning on import. To avoid raising this warning when a user imports Dynamics while diffrax is installed, this PR moves all `diffrax` imports and solver type-checks to be as late as possible, so that it is only imported once a user actually tries to call a diffrax solver (which requires the user to import diffrax themselves).

### Details and comments

Moved all diffrax imports into functions, and set type-checks for diffrax solvers to always be last when iterating through the possible solver types. Diffrax will therefore only be imported if:
- A user is actually trying to call a diffrax solver.
- A user passes an invalid `method` kwarg to one of the `Solver.solve`/`solve_lmde`/`solve_ode` methods, which will eventually raise an error anyway.

To preserve type hints for `solve_lmde` and `solve_ode`, a type variable `TypeVar('AbstractSolver')` is created to use in the function signatures.

### DIFF
--- a/qiskit_dynamics/solvers/diffrax_solver.py
+++ b/qiskit_dynamics/solvers/diffrax_solver.py
@@ -26,14 +26,9 @@ from qiskit_dynamics.array import Array, wrap
 
 from .solver_utils import merge_t_args
 
-
 try:
-    from diffrax import ODETerm, SaveAt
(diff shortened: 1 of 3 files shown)


chriseclectic (approved):
LGTM

PR #118, opened 2022-08-08
Fix RotatingFrame basis transformation methods to work for a list of CSR matrices

DanPuzzuoli:
### Summary

The methods `RotatingFrame.operator_into_frame_basis` and `RotatingFrame.operator_out_of_frame_basis` are supposed to work on `Operator`, CSR matrices, lists of `Operator`, lists of CSR matrices, and JAX BCOO arrays. A bug was preventing them from working on a list of CSR matrices, which this PR fixes.

### Details and comments

Updates both functions so that, if a list of valid types are passed, the method will be recursively called on the elements of the list. To facilitate this, also added a `convert_type` optional argument to both that can be used to avoid the initial type conversion that takes places in these methods (which is unnecessary in the recursive calls).

Added tests for `RotatingFrame` to validate this behaviour on lists of CSR matrices.

### DIFF
--- a/releasenotes/notes/sparse-rotating-frame-bug-88fffd0f39d8128b.yaml
+++ b/releasenotes/notes/sparse-rotating-frame-bug-88fffd0f39d8128b.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    ``RotatingFrame.operator_into_frame_basis`` and ``RotatingFrame.operator_out_of_frame_basis``
+    was fixed to work on lists of scipy CSR matrices.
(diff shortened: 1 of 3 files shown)


releasenotes/notes/sparse-rotating-frame-bug-88fffd0f39d8128b.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    ``RotatingFrame.operator_into_frame_basis`` and ``RotatingFrame.operator_out_of_frame_basis``
+    was fixed to work on lists of scipy CSR matrices.
```
brosand, line 5: change "was fixed" to "were fixed"

brosand (commented):
LGTM, just fix the one typo

releasenotes/notes/sparse-rotating-frame-bug-88fffd0f39d8128b.yaml
DanPuzzuoli: Done.

brosand (approved):
LGTM

PR #125, opened 2022-09-01
Fixing pulse jit bug

### LINKED ISSUE #122: JAX pulse simulation compilation bug
### Informations

- **Qiskit Dynamics version**: 0.3.0
- **Python version**: 3.10
- **Operating system**: Mac OS X

The automatic `jit` compilation routine for simulating pulse schedules with JAX raises errors under certain conditions.

Specifying if using `method='jax_odeint'`, specifying the `t_eval` kwarg will cause the error to be raised. Also, if using `method='jax_expm'`, the error will be raised regardless of whether `t_eval` is included or not.

Examples here can probably be reduced further in terms of lines of code, but here is a minimal example:

```
import numpy as np
from qiskit import pulse

from qiskit_dynamics import Solver

import jax
jax.config.update("jax_enable_x64", True)

# tell JAX we are using CPU
jax.config.update('jax_platform_name', 'cpu')

Array.set_default_backend('jax')

from qiskit_dynamics import Solver

Z = np.array([[1., 0.], [0., -1.]])
X = np.array([[0., 1.], [1., 0.]])

v = 5.
anharm = -0.33
r = 0.02
dt = 0.222222222222222 # dt for pulse schedules

# static part
static_hamiltonian = 2 * np.pi * v * Z
# drive term
drive_hamiltonian = 2 * np.pi * r * X

solver = Solver(
    static_hamiltonian=static_hamiltonian,
    hamiltonian_operators=[drive_hamiltonian],
    rotating_frame=static_hamiltonian,
    hamiltonian_channels=['d0'],
    channel_carrier_freqs={'d0': v},
    dt=dt
)

with pulse.build(name='x') as schedule:
    pulse.set_frequency(v, channel=pulse.DriveChannel(0))
    pulse.play(pulse.Drag(300, 1., 100, 0.), pulse.DriveChannel(0))
```

This works:
```
solver.solve(
    signals=schedule,
    y0=np.eye(2, dtype=complex),
    t_span=[0, 300 * dt],
    method='jax_odeint',
    atol=1e-10,
    rtol=1e-10
)
```

However this does not:
```
solver.solve(
    signals=schedule,
    y0=np.eye(2, dtype=complex),
    t_span=[0, 300 * dt],
    t_eval=[0, 150 * dt, 300 * dt],
    method='jax_odeint',
    atol=1e-10,
    rtol=1e-10
)
```

and similarly this does not:

```
solver.solve(
    signals=schedule,
    y0=np.eye(2, dtype=complex),
    t_span=[0, 300 * dt],
    t_eval=[0, 150 * dt, 300 * dt],
    method='jax_expm',
    max_dt=dt
)
```

The above calls should work and produce correct results.

### Suggested solutions

Both errors are being raised in different places, but the underlying reason is the same: the function being internally compiled ([located here](https://github.com/Qiskit/qiskit-dynamics/blob/cf8063ce3436a2ce778bd41d486cde765aa2369d/qiskit_dynamics/solvers/solver_classes.py#L712)) takes `t_span` as an argument, but in both cases (if `t_eval` is specified, or if a fixed step solver is being used), `t_span` cannot be compiled over.

Some imperfect options:
- Set `t_span` as a static argument in the internally compiled function. This is an extremely easy option, but will limit the usefulness of the compilation to cases where all `t_span` values are the same. It will still apply to a lot of cases, but I think this is an option of last resort.
- Figure out how to make both of the failure cases compilable. I think the first one is possible - the only reason it wasn't written to be compilable originally was due to difficulty, but I think in principle it can be done. This is an imperfect option however as I seem to remember there being fundamental issues with making the second one compilable (will need to see).

DanPuzzuoli:
### Summary

Closes #122.

See the discussion in #122 for a summary of the bug including technical details. This PR fixes the bug by taking the following actions (described in [this comment](https://github.com/Qiskit/qiskit-dynamics/issues/122#issuecomment-1232177882)):
- The bug for `method='jax_odeint'` has been fixed by updating the internal time-value handling of `jax_odeint`.
- The bug for `method` being a diffrax solver has been fixed with minimal changes. (It was using the problematic time-handling functions, but this was actually unnecessary as `diffrax` is able to take care of these things itself, so dynamics doesn't need to deal with it.)
- The bug for `method` being a dynamics-based fixed-step solver is now avoided by making it so that the automatic jitting behaviour is only triggered in the above two cases. (There isn't a simple "fix" for making the jitting work for these solvers.)

### Details and comments

The following changes have been made:
- Tests for `Solver` have been added that trigger the errors described in #122 (which are now passing).
- For diffrax solver fix:
    - Removed unnecessary call to `merge_t_args`
- For jax_odeint:
    - Modified `merge_t_args` and `trim_t_args` to always append/remove the endpoints. The original logic was to check if any end points of `t_eval` and `t_span` were the same, and if yes it wouldn't append/remove time points where appropriate. I realized that this special handling is not actually necessary, it is fine to always add/remove the endpoints.
    - Added `merge_t_args_jax` and `trim_t_args_jax` to be the JAX-compilable versions of these functions. To do compilable validation, most validation checks in `merge_t_args_jax` are now written with JAX conditionals. If a validation check fails, the output will be full of `nan` values. Unfortunately this is the only way to "raise an error" in a JAX compilable way.
    - Modified tests for `merge_t_args` and `trim_t_args` to check for the updated behaviour.
    - Added tests for `merge_t_args_jax` and `trim_t_arg_jax`, utilizing inheritance from the non-JAX test cases to limit number of lines.
    - Updated `jax_odeint` to use the JAX versions of these functions.
    - Added test case for `jax_odeint` that validates we can compile/differentiate functions that take both `t_span` and `t_eval` as inputs.
- To avoid bug for fixed step solvers:
    - Updated `Solver.solve` to only do the automatic compilation when the method is either `jax_odeint` or a diffrax solver.

### DIFF
--- a/qiskit_dynamics/solvers/solver_utils.py
+++ b/qiskit_dynamics/solvers/solver_utils.py
@@ -26,6 +26,12 @@ from qiskit import QiskitError
 from qiskit_dynamics.array import Array
 from qiskit_dynamics.models import LindbladModel
 
+try:
+    from jax.lax import cond
+    import jax.numpy as jnp
+except ImportError:
+    pass
+
 
 def is_lindblad_model_vectorized(obj: any) -> bool:
     """Return True if obj is a vectorized LindbladModel."""
@@ -110,12 +116,64 @@ def trim_t_results(
         return results
 
     # remove endpoints
-    results.t = results.t[1:]
-    results.t = results.t[:-1]
+    results.t = results.t[1:-1]
+    results.y = Array(results.y[1:-1])
 
-    results.y = Array(results.y[1:])
-    results.y = Array(results.y[:-1])
+    return results
 
+
+def merge_t_args_jax(t_span, t_eval=None):
+    """JAX-compilable version of merge_t_args. Rather than raise errors, sets return values to
+    jnp.nan to signal errors.
+    """
+
+    if t_eval is None:
+        return Array(t_span, backend="jax")
+
+    t_span = Array(t_span, backend="jax").data
+    t_eval = Array(t_eval, backend="jax").data
+
+    out = jnp.append(jnp.append(t_span[0], t_eval), t_span[1])
+
+    # raise error if not one dimensional
+    if t_eval.ndim > 1:
+        raise ValueError("t_eval must be 1 dimensional.")
+
+    # output nan if t_eval point lies outside t_span
+    t_min = jnp.min(t_span)
+    t_max = jnp.max(t_span)
+    out = cond(
+        (jnp.min(t_eval) < t_min) | (jnp.max(t_eval) > t_max),
+        lambda s: jnp.nan * s,
+        lambda s: s,
+        out,
+    )
+
+    # output nan if t_eval and t_span have incompatible orderings
+    diff = jnp.diff(t_eval)
+    t_direction = jnp.sign(t_span[1] - t_span[0])
+    out = cond(jnp.any(t_direction * diff < 0.0), lambda s: jnp.nan * s, lambda s: s, out)
+
+    return Array(out)
+
+
+def trim_t_results_jax(results, t_eval):
+    """JAX-compilable version of trim_t_results. The cond call is necessary due to
+    a bug in odeint that only occurs if the first two time values are duplicates.
+    """
+
+    if t_eval is None:
+        return results
+
+    results.y = Array(
+        cond(
+            results.t[0] == results.t[1],
+            lambda y: jnp.append(jnp.array([y[0]]), y[2:-1], axis=0),
+            lambda y: y[1:-1],
+            Array(results.y).data,
+        )
+    )
+    results.t = Array(results.t[1:-1])
     return results
 
 
--- a/qiskit_dynamics/solvers/solver_utils.py
+++ b/qiskit_dynamics/solvers/solver_utils.py
@@ -82,7 +82,7 @@ def merge_t_args(
 
     diff = np.diff(t_eval)
 
-    if np.any(t_direction * diff <= 0.0):
+    if np.any(t_direction * diff < 0.0):
         raise ValueError("t_eval must be ordered according to the direction of integration.")
 
     # if endpoints are not included in t_span, add them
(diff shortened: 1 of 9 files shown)


qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -116,17 +115,65 @@ def trim_t_results(
...
+    t_min = jnp.min(t_span)
+    t_max = jnp.max(t_span)
+    out = cond(
+        (jnp.min(t_eval) < t_min) | (jnp.max(t_eval) > t_max),
+        lambda s: jnp.nan * s,
+        lambda s: s,
+        out,
+    )
+
+    # output nan if t_eval and t_span have incompatible orderings
```
brosand, line 152: why output `nan` and not raise an error?

test/dynamics/solvers/test_solver_classes.py
```diff
@@ -1182,6 +1182,63 @@ def setUp(self):
         super().setUp()
         self.method = "jax_odeint"
 
+    def test_t_eval_t_span_jax_odeint(self):
+        """Test internal jitting works when specifying t_eval and t_span. This catches a bug
```
brosand, line 1186: Why are these done through pulse schedules and not raw signals?

brosand (commented):
Left a couple questions,, and then good to approve.

qiskit_dynamics/solvers/solver_utils.py
DanPuzzuoli: The error checking needs to be compilable/differentiable, but the error depends on the _values_ of the arrays. There's no way of doing value-dependent error raising in a traceable way in JAX, so in cases like this they recommend outputing `nan` as a way of signaling an error took place.

test/dynamics/solvers/test_solver_classes.py
DanPuzzuoli: The purpose of this PR and this test are explicitly to fix jitting in the pulse schedule simulation context.

brosand (approved):
Lgtm

PR #128, opened 2022-09-15
To signal sum fix

### LINKED ISSUE #127: Empty samples breaks`to_SignalSum`
Passing an empty `DiscreteSignal` into `to_SignalSum` causes an array with dimensions `(1,0)` to be transpoed to `(0,1)`, which breaks the `DiscreteSignal` `init` function. The transposition should not be applied in the empty sample case.

Current behavior (raises error):
```
from qiskit_dynamics import DiscreteSignal
from qiskit_dynamics.signals.signals import to_SignalSum

to_SignalSum(DiscreteSignal(samples=[], dt=1.))
```

To fix, add an `if` statement to `to_SignalSum` which only transposes the input samples if their size is not `(1,0)`

brosand:
Fixes #127 

Adds an `if` statement to test for empty sample inputs in the signal sum case, which was breaking when empty samples passed through `to_SignalSum`, which was called for every `SignalSum`. Added a test case, but the test situation is a bit difficult, so will need to be discussed further. This is just about getting the PR up so it can be integrated with the ongoing `PulseSimulator` work.


qiskit_dynamics/signals/signals.py
```diff
@@ -1134,13 +1135,22 @@ def to_SignalSum(sig: Union[int, float, complex, Array, Signal]) -> SignalSum:
...
+                dt=sig.dt,
+                samples=Array([sig.samples.data]),
+                start_time=sig.start_time,
+                carrier_freq=Array([sig.carrier_freq.data]),
+                phase=Array([sig.phase.data]),
+            )
+        else:
+            return DiscreteSignalSum(
+                dt=sig.dt,
+                samples=Array([sig.samples.data]).transpose(1, 0),
```
antalszava, line 1149: Fly-by comment: imho this line breaks otherwise.
```suggestion
                samples=Array([sig.samples.data]).transpose((1, 0)),
```

brosand: Hi @antalszava, appreciate the review! Pretty sure that transpose in this case will take both tuples and individual ints. If you're having problems though let me know.

DanPuzzuoli (changes requested):
I think for the testing, if you're going to call `to_SignalSum` directly, you should make a new test class. Alternatively, don't ever test `to_SignalSum` directly (e.g. with your `self.signal0 + self.signal0` test).

The last thing for testing we should double check is how this will be used in the pulse -> signal converter flow. E.g. will fixing this totally fix the issue there? I think it probably will, but we can be sure by adding a test case for that to this PR. I think this would probably have to be an integration test, where you add a test to the `test_solver_classes.TestPulseSimulation` class and have it try to simulate a schedule where the waveform for one of the channels in the model is actually empty. Maybe the easiest thing here is if you make the other changes, then at the end I can make a PR to this PR adding this as I'll be able to do it pretty quickly since I know the test code well.

DanPuzzuoli (approved):

PR #130, opened 2022-09-26
Fix bugs introduced by dependency updates

DanPuzzuoli:
### Summary

The tests and docs have broken without any changes to the code, so presumably due to dependency updates.

### Details

JAX updates broke the tests, but fixed by changing 1 line: The function `jnp.ones_like` no longer works on `Array` objects. This PR fixes our single usage of them. Tests now working.

@mtreinish pointed out that docs build was failing due to an update to , so this PR pins it to a version that works, as per [this commit to Terra](https://github.com/Qiskit/qiskit-terra/commit/df599dc9b189fe6d027386d989d19c3946a1476a).

### DIFF
--- a/qiskit_dynamics/signals/signals.py
+++ b/qiskit_dynamics/signals/signals.py
@@ -101,7 +101,7 @@ class Signal:
                 self._is_constant = True
 
             if envelope.backend == "jax":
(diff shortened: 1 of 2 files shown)


brosand (approved):
LGTM

PR #131, opened 2022-09-28
Fix after latest JAX update

DanPuzzuoli:
### Summary

Weirdly, the latest JAX update broke a numerical comparison in one of the tests. The latest update caused numerically compared values to differ by slightly more than the comparison tolerance. It's somewhat disconcerting, but it's also okay as if we lower the simulation tolerance used to generate the arrays the go back to being below the comparison tolerance. It's possible that something they changed had an impact on some rounding errors.

To fix this, I slightly increased the comparison tolerance, rather than decrease the simulation tolerance. (I made this choice to avoid increasing the time the tests take to run.)


brosand (approved):
LGTM

PR #136, opened 2022-10-17
Add cache to `Solver` for JIT compiled simulation functions.

### LINKED ISSUE #135: Update pulse schedule simulation compilation to be variable size
When simulating a list of pulse schedules using JAX, the `Solver.solve` method internally takes advantage of JAX compilation to speed up simulation. To get around the limitation of compilation that require function inputs to be a fixed shape, the strategy is as follows:
- Given a list of schedules, determine the max duration.
- Build a function for simulating a single schedule with length of the max duration. 
     - The inputs of this function include a large array containing all samples for all channels in the schedule, along with the duration of the specific schedule in units of time `T`.
- Tell JAX to compile the function.
- For each schedule, extract all of the samples into a single array, padding them to be the length of the max duration schedule (so they are all the same), and call the simulation function with this array and the integration interval `[0, T]`.

As all samples are always padded out to be the length of the max duration schedule, JAX will only compile the function on the first schedule, then re-use the compiled function for all remaining schedules.

This strategy works well to speed up a single call to `Solver.solve`, but doesn't do anything to help repeated calls: the compilation cost is incurred with every call to `Solver.solve`. For small system sizes, where compilation can literally take orders of magnitude longer than simulation itself (e.g. 1 second vs 10 ms), this cost greatly impacts the speed of JAX-based pulse simulation in Dynamics.

As discussed with @ihincks , alternative strategies could fix this issue. For example: 
- Compile the simulation function to a fixed size. 
- For each encountered schedule: if it is a shorter duration, pad it and simulate with the compiled function, and if it is a longer duration, recompile the function for a longer fixed size (chosen heuristically).
- Repeat, possibly keeping track of each duration that was used for compilation, and always padding each schedule with the shortest number of samples to achieve an already-compiled-over shape.

Some version of this would be a great update.

ihincks:
### Summary

Fixes #135 by adding a cache of JIT compiled functions to the `Solver`.

### Details and comments

I am opening this PR before adding tests etc. to make sure that @DanPuzzuoli is on the same page about the implementation.


ihincks:
I kind of assume you want to take List[Schedule] away as an option for solve because I believe the cache doesn't make that ability as useful?

DanPuzzuoli (commented):
I generally like the strategy of putting in these parameters - it's a nice way to handle it and keep it tune-able.

One major thing is that I think the `_jit_cache` attribute is actually unnecessary. Note that the `sim_function` being constructed doesn't use `duration` at all - the duration only ends up getting used implicitly during the jitting process, when it gets called with an `all_samples` that has already been padded. If you do `jit_func = jit(func)` and then call `jit_func` with arrays of different shapes/types, for each call JAX itself will check if its already compiled for that shape/type, and either use the already-compiled version or compile to the new shape/type. On the dynamics side the only thing you'll need to keep track of is the durations to control the padding.

One other thing we need to be careful of is that a user could be calling with different `kwargs` (for different solver options), and we should be aware of this as well. So, technically, whatever caching mechanism is used (either JAX's own or something else) needs to return a different function if `kwargs` are different.

I think one way to fix these problem, taking advantage of JAX's functionality, is:
- `_get_jit_sim_function` should only be called once at instantiation, and it shouldn't have _any_ inputs.
- The `sim_function` internally should additionally take `**kwargs` as an argument, and when you call `jit`, `**kwargs` should be considered a static argument. (Side note: when you tell JAX that an arg is static, it will treat it as concrete and static when compiling. If you call the function again and change a static argument, it will recompile, even if all of the other arguments are the same shape/type. This should take care of caching compiled version of the function for different solver options.) (Other side note: I've never tried treating `**kwargs` as static, so I'm not 100% sure it'll work, but I'm hopeful!)
- Keep track of the compiled durations/padding as you're doing, but just keep calling the same "compiled" sim function, and let JAX take care of everything else.

With regards to your question about `solve` taking `List[schedule]`:
- This PR does eliminate the technical reason we put it in for shared compilation, which to be honest was a major motivator.
- However, I'm partial towards keeping it, as we just introduced it in the last release, and I've already found it somewhat convenient to be able to just do a single call to `solve`.

That all being said, this _does_ present an opportunity to simplify things internally. With this the JAX version of schedule simulation doesn't need to be completely aware of all sims being submitted at once, so `_solve_schedule_list_jax` and `_solve_list` could be simplified to run only a single simulation, and the looping over lists can be refactored out to `solve` itself. The condition for when to use the jitted JAX schedule can also be simplified to be checked on a case-by-case basis (currently it checks that _everything_ in the list is a schedule before deciding to do the compilation trick).

ihincks:
These comments all make sense to me. I'll make the changes and you can find out whether I've understood.

ihincks:
This is still WIP because of the caching problems we discussed. Other than that, it working with the limited testing I've tried. I think it'd be worth you taking a look anyhow to decide whether this DiscreteSignal direction I've taken is okay.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -503,187 +515,118 @@ def solve(
...
+    def _jit_shape(self, required_duration: int):
+        """If possible, return a nearby samples shape so that we don't have to recompile. Otherwise,
+        add a new one to the cache."""
+        for duration in sorted(self._jit_cache):
+            if duration * self._JIT_MIN_PADDING_FRAC < required_duration <= duration:
+                # something appropriate was found in the cache
+                break
+        else:
+            duration = round(self._JIT_DURATION_PADDING * required_duration)
+            self._jit_cache.add(duration)
```
DanPuzzuoli, line 598: Looks like these lines should be at one level of indentation deeper.

DanPuzzuoli (commented):
I really like how much this is already looking like it will simplify the internals of `solve`.

Aside from the minor comments, the two major technical comments are:
- It's cool that you've changed the internal compilation behaviour to work for the case that a user gives all signals as `DiscreteSignal`s. To make this work properly in general though you'll need to add some more args to `_jit_sim`. Each `DiscreteSignal` instance has its own `dt` (sample width) and `start_time`, so to properly handle this you'll need to pass an array of `dt` values and an array of `start_time` values to `_jit_sim`. For pulse schedule simulation, all of hte `dt` values will be `self._dt` and all of the `start_time`s will be `0`. it's possible you will be able to keep these as non-static args, but I'm not sure. I'm not 100% sure its worth the hassle of doing this - you could revert to just doing schedules - but I'll leave that up to you.
- If we include the all `DiscreteSignal` case, when testing we'll need to check how internal jitting impacts external jitting (e.g. if this internal jitting gets triggered when a user is calling `solve` as part of a bigger function they're trying to compile). My understnding is there shouldnt be issues here, but I have run into some weird bugs with JAX surrounding stuff like this in the past.

qiskit_dynamics/solvers/solver_classes.py
ihincks: this is a for-else statement so i think its correct. one python designer once quipped that he wishes it had been named for-nobreak, because everyone always gets confused about the meaning of else there

qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -288,3 +292,102 @@ def setup_args_lists(
...
+            self._hash = 0
+
+    def __hash__(self):
+        return self._hash
+
+    def __eq__(self, other):
+        return self.obj == other
+
+    def __repr__(self):
+        return f"HashWrapper({repr(self.obj)})"
```
ihincks, line 316: This implementation assumes that once a `HashWrapper` instance is put into a hash map, the contents of `obj` is not mutated until the hash map leaves scope. For example, this is a valid use because the `set` is dereferenced before any objects can be mutated:

```python
def make_unique(objects):
    return list(set(map(HashWrapper, objects)))
```

However, this is problematic:

```python
def yield_unique(objects):
    yield from set(map(HashWrapper, objects))
```
because something like
```python
for obj in yield_unique(objects):
    obj.x = y
```
can cause undefined behaviour.

Therefore, I have a lingering concern over the whole premise of this PR because we are essentially in the second case: the hashmap of mutable objects lives inside of the `lru_cache` whose scope is ultimately out of our control.

Given these concerns, do we:
1) Accept the risk "because it probably won't matter"?
2) Modify `HashWrapper` to do potentially (expensive?) copies of obj?
3) Modify `HashWrapper` in some other complicated way?
4) Somehow tighten the scope of the hashmap?
5) Give up on the idea?

DanPuzzuoli: As per our discussion earlier, these are the combinations of things that need to go into `HashWrapper`:
- `self`
    - I didn't think of this one initially when talking about this but I guess this is a bit annoying. I think we mostly think of the `Solver` as being immutable once constructed, and it doesn't explicitly have a public API for modifying it, but you _can_ change `solver.model`, e.g. by specifying a different rotating frame. It would be a shame to have to fully copy the solver. 
- `y_cls` 
    - This can be `None` (indicating its some form of array), `Statevector`, `DensityMatrix`, `SuperOp`, or `Operator` (where the last 4 are all from `qiskit.quantum_info`). Not sure if these types are hashable.
- `kwargs`
    - These are keyword arguments passed to `solve_lmde`, and this will only be triggered in the following cases.
    - If `kwargs['method'] == 'jax_odeint'`, then all of the other kwargs are going to be basic python types, like strings and numeric types. I think in this case there is no concern.
    - For the `diffrax` library, `kwargs['method']` will be an instance of a subclass of `AbstractSolver` (a `diffrax` class). An example usage is in [this test case](https://github.com/Qiskit/qiskit-dynamics/blob/85128b7fb7897be2450804dd2823215139167b02/test/dynamics/solvers/test_solver_functions.py#L617). Similarly, you can see in this example that the step size controlling arguments are set in a class called `PIDController`. There are other class-style arguments like this for diffrax solvers, like there is a `SaveAt` class for specifying time points at which the state should be saved in the results object.
    - The `diffrax` case complicates things, but I also think copying these should be fine, the ones I'm aware of are light-ish weight objects.

DanPuzzuoli:
We've decided to close this PR for now, due to its complicated nature, to be re-opened at a later date if necessary.

PR #139, opened 2022-10-27
Initial DynamicsBackend class

### LINKED ISSUE #74: Function to compute dressed states and frequencies

Inputs: 
 - Static hamiltonian - a hermitian operator that is assumed to be nearly diagonal
 - Subsystem dimensions - a list of the dimensions of the tensor-factors

Outputs:
 - The 'dressed states': eigenvectors of the static hamiltonian labelled according to which computational state they correspond to.
    - E.g. this could be a dictionary whose keys are the computational state labels (strings of 0, 1, 2, ...) with values being the corresponding eigenstate
    - Alternatively, same as the above but given as two lists.
- The 'dressed frequencies': For each subsystem, the difference between the eigenvalues of the dressed first excited state and the ground state, divided by 2 pi. This could just be returned as a 1d array.

Some code for performing these steps already exists in Aer in the `HamiltonianModel._compute_drift_data()` method, so we can draw on this if it makes sense to do so.

### LINKED ISSUE #75: Functions for computing measurement probabilities and sampling counts

_Compute probabilities function_:

Inputs:
- `np.ndarray` (1d for state vector or 2d for density matrix) or one of the quantum info QuantumState types.
- 'dressed states' (eigenvectors for drift hamiltonian that defines the measurement projectors) (output of function in #74 )
- 'dressed state labels' (the computational state strings corresponding to the dressed states) (output of function in #74 )

Outputs:
- Vector of probabilities (ordered appropriately according to 'dressed state labels')

_Sample counts_:

Inputs:
- Probability vector with corresponding outcome labels
- `n_shots` - number of samples to draw

Output:
- Dict of counts of samples, with keys being outcome labels and values being the counts

DanPuzzuoli:
### Summary

Closes #74 
Closes #75

This PR implements an initial version of the `PulseSimulator` class, with a very minimal feature set. The main goal is to establish the structure of the class with a minimal level of usability, an then features can be iteratively added.

Currently a WIP, the minimal functionality goal is to allow a user to instantiate an instance with a custom model (defined in terms of a Dynamics `Solver`), and to submit schedules to `PulseSimulator.run` and get results objects back containing counts. The current functionality is shown in the file `PulseSimulator_custom_demo.ipynb` in the root folder (to be deleted prior to merge).

### Details and comments

Establishes the `PulseSimulator` class, which is instantiate with:
- A qiskit Dynamics `Solver` instance configured to simulate pulse schedules.
- A list of integers specifying the subsystem structure of the model in `Solver`.

The class assumes that, regardless of the rotating frame the `Solver` is configured in, the non-rotating frame version of the static Hamiltonian defines the dressed state structure (i.e. the lab frame static Hamiltonian). At instantiation, the lab frame static Hamiltonian is diagonalized, and eigenvalues/eigenvectors are sorted according to dressed-state ordering.

Calling `run` will:
- Validate the input is a list of schedules with one measurement in each schedule.
- Default to simulating the ground state.
- Simulate the schedules.
- Normalize the output states and sample (normalization is necessary to ensure proper sampling regardless of tolerances used).
- Return a `DynamicsJob` object containing results.

For now, the `DynamicsJob` object is an `_async=False` subclass of `Job`.

I think the main things to discuss are:
- The constructor interface.
- Options handling - I think we should just use the Aer simulators as an example here, but want to discuss first.


qiskit_dynamics/pulse/dynamics_job.py
```diff
@@ -0,0 +1,80 @@
...
+        return self._result
+
+    def status(self):
+        """Gets the status of the job.
+
+        Returns:
+            JobStatus: The current JobStatus.
+        """
+        if self._result is None:
+            return JobStatus.INITIALIZING
```
eggerdj, line 78: We would need more granularity here, especially for large simulations. E.g. need to know if the job is running.

qiskit_dynamics/pulse/dynamics_job.py
```diff
@@ -0,0 +1,80 @@
...
+        fn: Callable,
+        fn_kwargs: Dict,
+        **kwargs
+    ) -> None:
+        """Initializes the job.
+
+        Args:
+            backend: The backend used to run the job.
+            job_id: A unique id in the context of the backend used to run the job.
+            fn: Function to run the simulation.
```
eggerdj, line 40: Should we also have some form of callback to track the progress of the simulation? That would be nice.

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class PulseSimulator(BackendV2):
+    def __init__(
+        self,
+        solver: Solver,
+        subsystem_dims,
+        subsystem_labels: Optional[List[int]] = None,
+        name: Optional[str] = 'PulseSimulator',
+    ):
```
eggerdj, line 47: In https://github.com/Qiskit/qiskit-experiments/pull/925 we went for a model centric approach. I guess that all of those end-up going into a solver anyhow but we do need to know things like how many levels we are dealing with. 

How do you plan on giving a discriminator to the backend? Not needed now but you could also think of the kernel. The question is in fact more general: How to implement the readout chain and which arguments does it need? The first version would at least need IQ data and counts, both single-shot and average.

In https://github.com/Qiskit/qiskit-experiments/pull/925 we either go from state vector to counts (Level 2) or we go from state vector to IQ data which we sample from a predetermined distribution. One thing that this is still lacking is a discriminator to go from IQ data to counts which would allow us to capture mis-classification of, e.g., the two state as the 1 state.

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Assumptions
+            - Solver is well-formed.
+            - The no-rotating frame Solver Hamiltonian operators are specified in undressed
+              basis using standard tensor-product convention (whether dense or sparse evaluation)
+
+        Design questions
+            - Simulating measurement requires subsystem dims and labels, should we allow this
+              to be constructed without them, and then measurement is just not possible?
+            - Should we add the ability to do custom measurements?
```
eggerdj, line 58: What do you mean by `custom measurements`? But given the name I think that this is needed.

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+            - Should we fill out defaults by extracting them from Solver?
+            - If no set frequency commands are set in the schedule, what should the channel
+              frequencies be?
+                - How are control channel frequencies handled these days?
```
eggerdj, line 62: Frequencies are in the pulse defaults. See `iq_pulse_backend` here: https://github.com/Qiskit/qiskit-experiments/pull/925/files

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+                        raise QiskitError(f"Attempted to measure subsystem {subsystem}, but it is not in subsystem_list.")
+                new_measurement_subsystems_list.append(new_measurement_subsystems)
+
+            measurement_subsystems_list = new_measurement_subsystems_list
+
+        solver_results = self.solver.solve(t_span=t_span, y0=y0, signals=schedules, **solver_options)
+
+        # construct counts for each experiment
+        ##############################################################################################
+        # Change to if statement depending on types of outputs, e.g. counts vs IQ data
```
eggerdj, line 228: This still needs IQ data handling as well as meas_return which can be single of average

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        return self._subsystem_dims
+
+    @property
+    def subsystem_labels(self):
+        """Subsystem labels."""
+        return self._subsystem_labels
+
+    def run(
+        self,
+        run_input: Union[Schedule, ScheduleBlock],
```
eggerdj, line 106: This should simulate circuits with schedules embedded in the calibrations.
```suggestion
        run_input: Union[QuantumCircuit, List[QuantumCircuit]],
```
IIRC backends should not run schedules directly but instead run circuits with schedules in them.

eggerdj (commented):
Take a look at https://github.com/Qiskit/qiskit-experiments/pull/925/files we managed to get that backend work in a Qiskit Experiments workflow. I added several comments on functionality that we need to have.

qiskit_dynamics/pulse/pulse_simulator.py
chriseclectic: This should be union of schedules and quantum circuits (single or lists)

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # get the dressed states
+        static_hamiltonian = _get_lab_frame_static_hamiltonian(solver.model)
+        dressed_evals, dressed_states = _get_dressed_state_decomposition(static_hamiltonian)
+        self._dressed_evals = dressed_evals
+        self._dressed_states = dressed_states
+        self._dressed_states_adjoint = self._dressed_states.conj().transpose()
+
+    @property
+    def solver(self):
```
chriseclectic, line 90: I wonder if we should have separate attributes for these, or access them all through options/set_options with `options.solver` the same way Aer handles its methods/noise models etc. I think this makes it more extensible since you dont need to keep adding more properties

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        """The internal Solver."""
+        return self._solver
+
+    @property
+    def subsystem_dims(self):
```
chriseclectic, line 95: Same with these and the labels, should probably be in options

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+from .dynamics_job import DynamicsJob
+from .pulse_utils import _get_dressed_state_decomposition, _get_lab_frame_static_hamiltonian
+
+
+class PulseSimulator(BackendV2):
+    def __init__(
+        self,
+        solver: Solver,
+        subsystem_dims,
```
chriseclectic, line 44: Can't the dims be inferred from the solver itself?

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        subsystem_labels: Optional[List[int]] = None,
+        name: Optional[str] = 'PulseSimulator',
```
chriseclectic, line 46: I dont think name is necessary here, unless its a base class. You want init to add `target=None` so you can pass in a custom target, and `**options` for passing arbitrary things to options (and solver, labels etc can just be handled by options).

For handling options I would do something like AerBackend does with configuration/properties of V1, where it can allow configuration of anything in target through `set_options` (unless you explicitly want to forbid it), or its own internal options.

qiskit_dynamics/pulse/pulse_simulator.py
chriseclectic: I think all these things are just internal configuration that would be handled through setting relevant simulator options with `set_options`

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+            description='Pulse enabled simulator backend.',
+            backend_version=0.1
+        )
+
+        self._solver = solver
+        self._subsystem_dims = subsystem_dims or [solver.model.dim]
+        self._subsystem_labels = subsystem_labels or list(range(len(subsystem_dims)))
+
+        # get the dressed states
+        static_hamiltonian = _get_lab_frame_static_hamiltonian(solver.model)
```
chriseclectic, line 83: I would move this into a `_set_solver` method for setting relevant attributes from a solver, so that if the solver option is changed it can be called again

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+    def drive_channel(self, qubit: int) -> Union[int, DriveChannel, None]:
+        pass
+
+    def control_channel(self, qubit: int) -> Union[int, ControlChannel, None]:
+        pass
+
+    def measure_channel(self, qubit: int) -> Union[int, MeasureChannel, None]:
+        pass
+
+    def _default_options(self):
```
chriseclectic, line 283: This is where all the options like solver, measurement config etc need to be defined

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        pass
+
+    def max_circuits(self):
+        pass
```
chriseclectic, line 287:
```suggestion
        return None
```

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def target(self):
+        pass
```
chriseclectic, line 290: This needs to return the target

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def drive_channel(self, qubit: int) -> Union[int, DriveChannel, None]:
+        pass
+
+    def control_channel(self, qubit: int) -> Union[int, ControlChannel, None]:
+        pass
+
+    def measure_channel(self, qubit: int) -> Union[int, MeasureChannel, None]:
+        pass
```
chriseclectic, line 281: For these keep them as NotImplementedError like in the base class until you implement them

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+                    success=True,
+                    results=output,
+                    date=datetime.datetime.now().isoformat()
+                )
+            )
+
+        return results_list
+
+    @property
+    def meas_map(self) -> List[List[int]]:
```
chriseclectic, line 268: Since we don't need to multiplex a simulator I think this should just be a list like `[[0], ..., [N-1]]`. You could also just leave it as `NotImplementedError` from the base class

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        measurement_subsystems_list,
+        shots
+    ) -> Result:
+        """Not sure here what the right delineation of arguments is to put in _run.
+        This feels somewhat hacky/arbitrary.
+        """
+        start = time.time()
+
+        # map measurement subsystems from labels to correct index
+        if self.subsystem_labels:
```
chriseclectic, line 211: This looks like something that can be run and cached once when the relevent option value is set or changed, not evaluated every `_run` call.

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        return self._subsystem_labels
+
+    def run(
+        self,
+        run_input: Union[Schedule, ScheduleBlock],
+        shots: int = 1,
+        y0 = None,
+        validate: Optional[bool] = False,
+        normalize_states: Optional[bool] = True,
+        solver_options: Optional[dict] = None,
```
chriseclectic, line 111: Either solver options shoudl be converted to simulator options, or you sohuld add a `solver_options` field to the simulator options, and then this run method should just be `**options` which allows temporarily overriding any simulator options when running.

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        """Subsystem labels."""
+        return self._subsystem_labels
+
+    def run(
+        self,
+        run_input: Union[Schedule, ScheduleBlock],
+        shots: int = 1,
+        y0 = None,
+        validate: Optional[bool] = False,
+        normalize_states: Optional[bool] = True,
```
chriseclectic, line 110: This hsould be a simulator option too

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+    @property
+    def subsystem_labels(self):
+        """Subsystem labels."""
+        return self._subsystem_labels
+
+    def run(
+        self,
+        run_input: Union[Schedule, ScheduleBlock],
+        shots: int = 1,
+        y0 = None,
```
chriseclectic, line 108: This isn't part of backend API, initial state should be set by the `Initialize` instruction in a circuit, or as a simulator option (if its part of the " noise model")

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @property
+    def subsystem_labels(self):
+        """Subsystem labels."""
+        return self._subsystem_labels
+
+    def run(
+        self,
+        run_input: Union[Schedule, ScheduleBlock],
+        shots: int = 1,
```
chriseclectic, line 107: Default shots should probably be 1024 like other backends. Its up to the simulator to be smart and know that if its a unitary circuit being run, you can run once till the end and sample all the shots from the final state like Aer does

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+        To test:
+        - Measuring a subset of qubits when more than one present
+        - If t_span=[0, 0] it seems that NaNs appear for JAX simulation - should solve this.
+        - Normalizing the state.
+        """
+
+        if validate:
+            _validate_run_input(run_input)
+
+        if solver_options is None:
```
chriseclectic, line 148: Options should be obtained from current values in self.options, and then overridden with any passed runtime options

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if y0 is None:
+            y0 = Statevector(self._dressed_states[:, 0])
+
+        # to do: add handling of circuits
+        schedules = _to_schedule_list(run_input)
+
+        # get the acquires instructions and simulation times
+        t_span = []
+        measurement_subsystems_list = []
```
chriseclectic, line 159: This should probably be moved to a method

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+                for subsystem in measurement_subsystems:
+                    if subsystem in self.subsystem_labels:
+                        new_measurement_subsystems.append(self.subsystem_labels.index(subsystem))
+                    else:
+                        raise QiskitError(f"Attempted to measure subsystem {subsystem}, but it is not in subsystem_list.")
+                new_measurement_subsystems_list.append(new_measurement_subsystems)
+
+            measurement_subsystems_list = new_measurement_subsystems_list
+
+        solver_results = self.solver.solve(t_span=t_span, y0=y0, signals=schedules, **solver_options)
```
chriseclectic, line 224: Do you plan on supporting circuits that need to be broken up into multiple solver class / schedules in the future? eg having a mid circuit measurement or something like that?

qiskit_dynamics/pulse/pulse_simulator.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # construct counts for each experiment
```
chriseclectic, line 226: I think you should wrap up measurement sampling into a separate function, so that you could pass in custom measurement functions as an option (and have some pre-defined ones you can choose from with string names)

chriseclectic (changes requested):
Good start, I think option handling can do with a bit of work to make it closer to existing backends and Aer simulator, rather than having custom properties and kwargs like is currently done. This will also be much more extensible in the future as we add more configurability for things like how measurement is handled, or rotating frames or things like that.

It also seems like the Target field is completely missing, which is the most important part of a pulse backend for working with the scheduler.

qiskit_dynamics/pulse/pulse_simulator.py
DanPuzzuoli: For this I mean a user being able to fully define the way counts, or IQ data, get computed. We will implement some standard options, but I figured it'd be nice for there to be a way for the user to completely specify this as well.

qiskit_dynamics/pulse/dynamics_job.py
DanPuzzuoli: This is related to above: for this PR no but could be a future improvement.

qiskit_dynamics/pulse/pulse_simulator.py
DanPuzzuoli: Agree with Chris. Adding IQ data generation will be left for a subsequent PR, and enabling any user configuration of this process can be handled by designing/implementing options.

qiskit_dynamics/pulse/pulse_simulator.py
DanPuzzuoli: To be added in subsequent PR.

qiskit_dynamics/pulse/pulse_simulator.py
DanPuzzuoli: This has been updated to handle circuits, schedules, and scheduleblocks. If a circuit, it will attempt to transpile it to a schedule. If the user has supplied enough information (either by attaching a pulse definition ot individual gates, or having basis gates defined in the target) it will succeed, otherwise the scheduler will raise an informative error itself.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+            provider: Provider of the backend.
+            options: Additional configuration options for the simulator.
+        Raises:
+            QiskitError: If any instantiation arguments fail validation checks.
+        """
+
+        super().__init__(
+            name="PulseSimulator",
+            description="Pulse enabled simulator backend.",
+            backend_version=0.1,
```
chriseclectic, line 113: This should probably be a string right?

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+      :func:`default_experiment_result_function`, and any other function set to this option
+      must have the same signature. Note that the default utilizes various other options that
+      control results computation, and hence changing it will impact the meaning of other options.
+    """
+
+    def __init__(
+        self,
+        solver: Solver,
+        target: Optional[Target] = None,
+        provider: Optional[Provider] = None,
```
chriseclectic, line 96: Is the provider kwarg actually needed? We dont have one in dynamics, so maybe can just leave as None for now when calling the super init

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+
+        if self.options.meas_map is None:
+            meas_map = [[idx] for idx in self.options.subsystem_labels]
+            self.set_options(meas_map=meas_map)
+
+        # self._target = target or Target() doesn't work as bool(target) can be False
+        if target is None:
+            target = Target()
+
+        # add default simulator measure instructions
```
chriseclectic, line 141: Not urgent, but if we decided to do more complicated things later with default target we could move this to an internal method for for initializing the target.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+            labels = list(range(len(self.options.subsystem_dims)))
+            self.set_options(subsystem_labels=labels)
+
+        if self.options.meas_map is None:
+            meas_map = [[idx] for idx in self.options.subsystem_labels]
+            self.set_options(meas_map=meas_map)
+
+        # self._target = target or Target() doesn't work as bool(target) can be False
+        if target is None:
+            target = Target()
```
chriseclectic, line 139: Since target is modified below, you might want to make a copy of the target if it is not None so you don't mutate the original kwarg value.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+                if key == "subsystem_dims":
+                    validate_subsystem_dims = True
+                self._options.update_options(**{key: value})
+
+        # perform additional validation if certain options were modified
+        if (
+            validate_subsystem_dims
+            and np.prod(self._options.subsystem_dims) != self._options.solver.model.dim
+        ):
+            raise QiskitError(
```
chriseclectic, line 207: Might be good to later add our own error class subclass for using in the backend.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+        ):
+            experiment_results.append(
+                self.options.experiment_result_function(
+                    experiment_name,
+                    solver_result,
+                    measurement_subsystems,
+                    memory_slot_indices,
+                    num_memory_slots,
+                    self,
+                    seed=rng.integers(low=0, high=9223372036854775807),
```
chriseclectic, line 326: Im not entirely sure whats going on here, but wanted to check that you still get deterministic measurement outcome if you set the `seed_simulator` option?

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+        experiment_name: Name of experiment.
+        solver_result: Result object from :class:`Solver.solve`.
+        measurement_subsystems: Labels of subsystems in the model being measured.
+        memory_slot_indices: Indices of memory slots to store the results in for each subsystem.
+        num_memory_slots: Total number of memory slots in the returned output. If ``None``,
+            ``max(memory_slot_indices)`` will be used.
+        backend: The backend instance that ran the simulation. Various options and properties
+            are utilized.
+        seed: Seed for any random number generation involved (e.g. when computing outcome samples).
+    Returns:
```
chriseclectic, line 380: nitpick: I would add a blank line between Args, Returns, Raises for all doc strings

qiskit_dynamics/pulse_simulator/pulse_simulator.py
```diff
@@ -0,0 +1,526 @@
...
+from .pulse_simulator_utils import (
+    _get_dressed_state_decomposition,
+    _get_lab_frame_static_hamiltonian,
+    _get_memory_slot_probabilities,
+    _sample_probability_dict,
+    _get_counts_from_samples,
+)
+
+
+class PulseSimulator(BackendV2):
```
chriseclectic, line 56: Should we call it `DynamicSimulator`, `ScheduleSimulator`, or something else, instead of `PulseSimulator`?

chriseclectic (commented):
Overall looks good, just a couple of suggestions and questions.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: Correct, changed to string.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: Removed it. Removed it from the `super().__init__` call as well as it defaults to `None` anyway.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: Sounds good - for now I'll leave it as is.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: True. Just set it up to copy a passed target.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: Yep. The `rng` is instantiated above using the `seed_simulator` option. This line then generates another valid seed using `rng` (a 64 bit positive integer, hence the weird high number) to pass to the results function to use for generating counts for the current experiment in the loop.

DanPuzzuoli: The tests already use this - they check exact dictionary equality, which only works due to the determinism from the seeds.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: Just added blank lines between these for all new doc strings.

qiskit_dynamics/pulse_simulator/pulse_simulator.py
DanPuzzuoli: Maybe `DynamicsBackend` would be good?

qiskit_dynamics/__init__.py
```diff
@@ -31,9 +31,11 @@
 from .solvers.solver_classes import Solver
 from .solvers.perturbative_solvers.dyson_solver import DysonSolver
 from .solvers.perturbative_solvers.magnus_solver import MagnusSolver
+from .dynamics_backend.dynamics_backend import DynamicsBackend
 
 from . import models
 from . import signals
 from . import pulse
+from . import dynamics_backend
```
chriseclectic, line 39: maybe just `backend`?

chriseclectic (approved):
Looks good. Suggestion to possibly rename submodule to just `backend`

qiskit_dynamics/__init__.py
DanPuzzuoli: changed to just `backend`, makes sense

chriseclectic (approved):

PR #149, opened 2022-11-11
Add get samples function to InstructionToSignals for JAX-jit usage

to24toro:
### Summary

This PR adds `get_samples ` to `InstructionToSignals` for JAX-jitting when using qiskit-pulse and removes the usage of `get_waveform` method of `SymbolicPulse`.

### Details and comments

`get_samples` function gets the envelope expression form` SymbolicPulse` and calls sympy.lambdify with numerical backend specified by Array class. The lambdified function is lru cached for performance.


DanPuzzuoli:
Okay cool thanks.

> Feel free to on hold this until above PR is merged, or you can merge this PR without JIT test and later add the the test in a separate (follow up) PR.

I think it makes sense to hold on this PR, as I don't think the required functionality can truly be verified until terra is at a point where the `static_argnums` can be removed.

nkanazawa1989:
Fair enough. Let's merge https://github.com/Qiskit/qiskit-terra/pull/9002 first.

test/dynamics/signals/test_pulse_to_signals.py
```diff
@@ -205,15 +218,59 @@ def test_multiple_channels_with_gaps(self):
...
+            gauss_get_samples, self.gauss_get_waveform_samples, atol=1e-7, rtol=1e-7
+        )
+
+    def test_jit_get_samples(self):
+        """Test compiling to get samples of Pulse."""
+
+        def jit_func(amp):
+            return get_samples(Constant(100, amp))
+
+        jit_samples = jax.jit(jit_func)(0.1)
```
to24toro, line 271: We expected this function will pass without using static_argnums after merging https://github.com/Qiskit/qiskit-terra/pull/9002

DanPuzzuoli:
Hey @to24toro , there are two more tests I think it'd be good to add:
- Can you add a test case for computing gradients as well? You can take the same `jit` test, but change it so that at the end, instead of `jit`, you do `jax.jit(jax.grad(jit_func))(0.1)`. You'll have to change the the function so that it returns a real scalar (as `grad` only works on real scalars), so you could just change what the function returns to be `get_samples(instance).sum().real`. Testing this combination of gradding/jitting is important because it can introduce a bit more complexity into what JAX is doing, and sometimes raise errors where `jit`/`grad` don't raise them on their own.
- I'm also thinking we should add a test for `Solver` that compiles/differentiates a simulation of a pulse schedule with symbolic pulses. In the `test/dynamics/solvers/test_solver_classes.py` file there is a test class called `TestPulseSimulationJAX`. Can you add a test to this class that defines a function building a schedule with symbolic pulses, simulating it with a `Solver`, and returning the final state, and attempts to compile/differentiate it? This may be a little redundant with the other tests but it will be a good integration test.

One last point: the `TestJaxBase` class has two helper functions: `jit_wrap` and `jit_grad_wrap`, which can be used to jit a function, or to `jit(grad(func))` a function `func`, where the latter takes the sum and the real part of the output. These may eb helpful/convenient.

test/dynamics/signals/test_pulse_to_signals.py
```diff
@@ -266,8 +268,22 @@ def test_jit_get_samples(self):
...
+                pulse_type="Constant",
+                duration=100,
+                parameters=parameters,
+                envelope=envelope_expr,
+                valid_amp_conditions=valid_amp_conditions_expr,
+            )
+            return get_samples(instance)
+
+        self.jit_wrap(jit_func)(0.1)
+        self.jit_grad_wrap(jit_func)(0.1)
```
to24toro, line 286: I tried to use jit_wrap and jit_grad_wrap.
Very useful.

qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -523,6 +524,8 @@ def solve(
             Array.default_backend() == "jax"
             and (method == "jax_odeint" or _is_diffrax_method(method))
             and all(isinstance(x, Schedule) for x in signals_list)
+            # check if jit transformation is already performed.
+            and not (isinstance(jnp.array(0), core.Tracer))
```
to24toro, line 528: This is related with the issue https://github.com/Qiskit/qiskit-dynamics/issues/175

to24toro:
I modified two points at https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/769e342a06f95d8232bd6475da3e7b46c80717d5 for JAX-jitting.

DanPuzzuoli:
I've removed the "on hold" label, as you pointed out that terra has been sufficiently updated. Once the errors are resolved and I re-review we can merge this!

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -119,29 +125,33 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
                 if isinstance(inst.pulse, Waveform):
                     inst_samples = inst.pulse.samples
                 else:
```
DanPuzzuoli, line 127: If `isinstance(inst.pulse, Waveform) == False`, is the only other option for `inst.pulse` to be a `SymbolicPulse`? Do we need to worry about any other cases in this code block?

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -254,3 +264,59 @@ def _get_channel(self, channel_name: str):
...
+        except KeyError as ex:
+            raise PulseError(
+                f"Pulse parameter '{symbol.name}' is not defined for this instance. "
+                "Please check your waveform expression is correct."
+            ) from ex
+    return _lru_cache_expr(envelope, Array.default_backend())(*args)
+
+
+@functools.lru_cache(maxsize=None)
+def _lru_cache_expr(expr: sym.Expr, backend):
```
DanPuzzuoli, line 307: Add type hint for `backend` as well as the return type.

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -254,3 +264,59 @@ def _get_channel(self, channel_name: str):
             raise QiskitError(
                 f"Invalid channel name {channel_name} given to {self.__class__.__name__}."
             ) from error
+
+
+def get_samples(pulse: SymbolicPulse):
```
DanPuzzuoli, line 269: Add type hint for return type.

test/dynamics/pulse/test_pulse_to_signals.py
```diff
@@ -303,15 +320,79 @@ def test_multiple_channels_with_gaps(self):
...
+        self.assertAllClose(signals[0].samples, samples0, atol=1e-7, rtol=1e-7)
+        self.assertAllClose(signals[1].samples, samples1, atol=1e-7, rtol=1e-7)
+        self.assertAllClose(signals[2].samples, samples2, atol=1e-7, rtol=1e-7)
+        self.assertAllClose(signals[3].samples, samples3, atol=1e-7, rtol=1e-7)
         self.assertTrue(signals[0].carrier_freq == 5.0)
         self.assertTrue(signals[1].carrier_freq == 3.1)
         self.assertTrue(signals[2].carrier_freq == 0.0)
         self.assertTrue(signals[3].carrier_freq == 4.0)
 
+    def test_get_samples(self):
```
DanPuzzuoli, line 332: Rewrite these tests to still call the `InstructionToSignals` converter. `get_samples` and `lru_cache_expr` are both internal functions, so ideally we won't directly test them. There are exceptions to this in Dynamics, but in this case I don't think it's necessary. The inputs/outputs of these functions are directly fed from/to the converter, so I don't think anything is gained by directly testing these functions, as opposed to verifying the behaviour of the converter on a symbolic pulse.

Maybe change the name of this to `test_SymbolicPulse`.

test/dynamics/pulse/test_pulse_to_signals.py
```diff
@@ -303,15 +320,79 @@ def test_multiple_channels_with_gaps(self):
...
+    def test_lru_cache_expr(self):
+        """Test lru_cache of lru_cache_expr function."""
+        gauss_envelop = Gaussian(duration=5, amp=0.983, sigma=2.0).envelope
+        self.assertTrue(
+            _lru_cache_expr(gauss_envelop, Array.default_backend())
+            is _lru_cache_expr(gauss_envelop, Array.default_backend())
+        )
+
+
+class TestJaxGetSamples(QiskitDynamicsTestCase, TestJaxBase):
```
DanPuzzuoli, line 350: Similar to the preceding comment, I don't think it's necessary to have a separate class for checking the behaviour of `get_samples`. These tests could be moved to the previous class, by rewriting them to verify that the converter returns the same thing, whether you pass in a `Waveform` or a `SymbolicPulse`. 

The `jit` test is great and very important, but again, it can be moved to the previous class and rewritten to work with the converter.

test/dynamics/solvers/test_solver_classes.py
```diff
@@ -1263,6 +1264,41 @@ def test_t_eval_t_span_jax_expm(self):
             max_dt=0.05,
         )
 
+    def test_jit_solve_with_internal_jit(self):
```
DanPuzzuoli, line 1267: Good test here. Can you make the schedule more complicated by adding `ShiftPhase`, `SetPhase`, `ShiftFrequency`, and `SetFrequency` instructions? I believe you updated the PR so that these could also be jitted through, but that is not being tested here.

Can you also add a derivative test here as well, not just `jit`.

DanPuzzuoli (changes requested):
Aside from current test failing issue (which seems to just be a sympy issue), I have some comments about reorganizing the new tests a little.

Looks good though.

test/dynamics/solvers/test_solver_classes.py
to24toro: add the new test with https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/3f214c574e2f611b635dabde4dc21e150a411697

test/dynamics/pulse/test_pulse_to_signals.py
to24toro: improved and changed the name at https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/e2c5148dade2da23c125e1471580e32046669d45

qiskit_dynamics/pulse/pulse_to_signals.py
to24toro: added at https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/4dab802b8e400282a022df72341da5aff353b5b0

qiskit_dynamics/pulse/pulse_to_signals.py
to24toro: added at https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/4dab802b8e400282a022df72341da5aff353b5b0

to24toro:
My author and commit name are not correct. So I modified them and force-pushed.

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -40,19 +46,17 @@ class InstructionToSignals:
...
-    The converter can be initialized with the optional arguments ``carriers`` and ``channels``. When
-    ``channels`` is given, only the signals specified by name in ``channels`` are returned. The
-    ``carriers`` dictionary specifies the analog carrier frequency of each channel. Here, the keys
-    are the channel name, e.g. ``d12`` for drive channel number ``12``, and the values are the
-    corresponding frequency. If a channel is not present in ``carriers`` it is assumed that the
-    analog carrier frequency is zero.
-
-    See the :meth:`get_signals` method documentation for a detailed description of how pulse
-    schedules are interpreted and translated into :class:`.DiscreteSignal` objects.
+    schedule. The converter applies to instances of :class:`Schedule`. Instances of
```
DanPuzzuoli, line 49: I think something's happened here - this is undoing some changes I had made in a previous PR (that I don't think you are intentionally trying to change).

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -64,14 +68,15 @@ def __init__(
         """Initialize pulse schedule to signals converter.
 
         Args:
-            dt: Length of the samples. This is required by the converter as pulse schedule are
-                specified in units of dt and typically do not carry the value of dt with them.
-            carriers: A dict of analog carrier frequencies. The keys are the names of the channels
+            dt: Length of the samples. This is required by the converter as pulse
```
DanPuzzuoli, line 71: These two code blocks are similarly changing some documentation I had changed previously. Maybe some issue with a previous attempt at merging `main` into this branch?

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -129,9 +134,8 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
 
         Args:
             schedule: The schedule to represent in terms of signals. Instances of
-                :class:`~qiskit.pulse.ScheduleBlock` must first be converted to
-                :class:`~qiskit.pulse.Schedule` using the
-                :func:`~qiskit.pulse.transforms.block_to_schedule` function in Qiskit Pulse.
+                :class:`ScheduleBlock` must first be converted to :class:`Schedule` using the
```
DanPuzzuoli, line 137: Similarly here, this is undoing some changes I had made.

test/dynamics/pulse/test_pulse_to_signals.py
DanPuzzuoli: For this I mean to fully get rid of this test class, and to move the cases it's testing into the preceding test class. by calling `InstructionToSignals` directly.

test/dynamics/pulse/test_pulse_to_signals.py
```diff
@@ -303,15 +317,101 @@ def test_multiple_channels_with_gaps(self):
...
+            # we can use only SymbolicPulse when jax-jitting
+            # bacause jax-jitting doesn't correspond to validate_parameters in qiskit.pulse.
+            instance = SymbolicPulse(
+                pulse_type="Constant",
+                duration=5,
+                parameters=parameters,
+                envelope=envelope_expr,
+                valid_amp_conditions=valid_amp_conditions_expr,
+            )
+            return get_samples(instance)
```
DanPuzzuoli, line 373: E.g. this test should use `InstructionToSignals` instead of `get_samples` directly.

test/dynamics/pulse/test_pulse_to_signals.py
```diff
@@ -303,15 +317,101 @@ def test_multiple_channels_with_gaps(self):
...
+
+        self.jit_wrap(jit_func_get_samples)(0.1)
+        self.jit_grad_wrap(jit_func_get_samples)(0.1)
+        jit_samples = jax.jit(jit_func_get_samples)(0.1)
+        self.assertAllClose(jit_samples, self.constant_get_waveform_samples, atol=1e-7, rtol=1e-7)
+
+    def test_pulse_types_combination_with_jax(self):
```
DanPuzzuoli, line 380: This is a great test, that could be moved to the preceding class.

test/dynamics/pulse/test_pulse_to_signals.py
to24toro: The class you said moving to is `TestPulseToSignals`?
If I move to the class, we fail to pass the test because `TestPulseToSignals` is not a subclass of `TestJaxBase`.
On the other hand, if `TestPulseToSignals` inherits `TestJaxBase`, all the test in `TestPulseToSignals` will be skipped in python test and executed in only JAX test.

qiskit_dynamics/pulse/pulse_to_signals.py
to24toro: I am so sorry to confuse you. As you say, it is not my intention.  I restored to that of main at .

test/dynamics/pulse/test_pulse_to_signals.py
DanPuzzuoli: I understand now. Maybe just rename this class then to `TestPulseToSignalsJAXTransformations` or something along these lines.

to24toro: rename and fix at https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/454f1194463df8cfe5bf002343e81fa92fdc04b4 and https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/2814cf0587ceabc6354cbf628560536bcad24577

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -43,14 +49,12 @@ class InstructionToSignals:
...
     :class:`~qiskit.pulse.ScheduleBlock` must first be converted to :class:`~qiskit.pulse.Schedule`
     using the :func:`~qiskit.pulse.transforms.block_to_schedule` function in Qiskit Pulse.
-
     The converter can be initialized with the optional arguments ``carriers`` and ``channels``. When
     ``channels`` is given, only the signals specified by name in ``channels`` are returned. The
     ``carriers`` dictionary specifies the analog carrier frequency of each channel. Here, the keys
     are the channel name, e.g. ``d12`` for drive channel number ``12``, and the values are the
     corresponding frequency. If a channel is not present in ``carriers`` it is assumed that the
     analog carrier frequency is zero.
-
```
DanPuzzuoli, line 53: My last question is: did you intentionally remove these spaces? Deleting these causes this whole block of text to be a single paragraph in the docs. I prefer these as separate paragraphs, as each is highlighting something different. Unless there is a specific reason for this, can you put the spaces back in.

to24toro: Sorry. I didn't take into account the paragraph change. I returned back at https://github.com/Qiskit/qiskit-dynamics/pull/149/commits/c3338c5b6ddd56a51cdd8fc156eff25174d6e2be.

DanPuzzuoli (approved):

PR #150, opened 2022-11-15
Fix dependency bugs in stable/0.3

### LINKED ISSUE #148: Signal class handling of constant envelope broken on stable/0.3 with latest jax version
### Informations

- **Qiskit Dynamics version**: stable/0.3
- **Python version**: 3.10
- **Operating system**: Mac OS

The how-to-use JAX userguide raises an error related to `jnp.ones_like` not accepting `Array` as an input. This was resolved in PR #130, but we should do a minor or patch release to 0.3 to fix this, either putting an upper bound on the JAX version, or by including the PRs adding fixes for JAX updates.

DanPuzzuoli:
### Summary

Closes #148

Fixes dependency bugs for stable/0.3 that were fixed in #130.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,7 +3,7 @@ astroid==2.9.3
 pylint==2.12.2
 black~=22.0
 qiskit-sphinx-theme>=1.6
-sphinx-autodoc-typehints
+sphinx-autodoc-typehints~=1.18,!=1.19.3
 jupyter-sphinx
 pygments>=2.4
 reno>=3.4.0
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -3,7 +3,7 @@ astroid==2.9.3
 pylint==2.12.2
 black~=22.0
 qiskit-sphinx-theme>=1.6
-sphinx-autodoc-typehints
+sphinx-autodoc-typehints~=1.18,!=1.19.3
```
ihincks, line 6: Without knowing too much about this particular package, I'd prefer the simpler `sphinx-autodoc-typehints>=1.19.4` here simply because for dev requirements there is no harm in being picky (as environments should always be custom for building docs).

DanPuzzuoli: just implemented this - seems fine locally, hopefully things build

ihincks (approved):

PR #153, opened 2022-11-21
Fix JAX version of t_span and t_eval merging functions

### LINKED ISSUE #147: nan now being returned in how-to-use JAX user guide in gradient computation on main
### Informations

- **Qiskit Dynamics version**: main
- **Python version**: 3.10
- **Operating system**: Mac OS
- **JAX version**: 0.3.24

The gradient computation in the how-to-use JAX user guide is returning `nan`. Weirdly, the gradient computation in the optimization tutorial is still working.

### Suggested solutions

This definitely seems to be a Dynamics problem - if I checkout `stable/0.3` then it works on the latest version of JAX, with the caveat that another bug (introduced by JAX becoming more strict with type checking) needs to first be fixed to get the code to run.

DanPuzzuoli:
### Summary

Closes #147 

Fixes the `merge_t_args_jax` and `trim_t_results_jax` functions. I think this is a final comprehensive solution to the problem that led to #147. 

### Details and comments

The bug stems from the following annoying technical issue that we've dealt with for a while:
- We've adopted scipy's `solve_ivp` time argument structure for specifying ODE problems in `solve_ode` and `solve_lmde`. `t_span` specifies the integration interval, and `t_eval` is an optional argument which tells the solver to only return the state at specific points in time (otherwise it will return it at whatever times the solvers stepped to).
- JAX's `odeint` function takes a single time argument, `ts`, which is essentially like `t_eval`, except the first and last times are interpreted as the whole integration interval. Hence, interfacing `odeint` into our `solve_ode`/`solve_lmde` functions requires translating the `t_span`/`t_eval` arguments into a single list of times to pass to `odeint`.
- One annoying peculiarity of `odeint`, which is what causes this bug, is that if the first or last entry of `ts` are repeated, the solution will have `nan` appearing in some entries (I'm forgetting which exactly). This is problematic for us because the endpoints of `t_span` are very commonly the endpoints of `t_eval`. Hence, merging `t_span` and `t_eval` into a single array by prepending/appending the entries of `t_span` onto `t_eval` will commonly result in repeated entries, causing this `nan` error to occur. (Note as well that we can't simply use `if` statements to check if these endpoints are equal and choose to append or not append them: this will violate the JAX requirement that the inputs/outputs to function are fixed-shape :D.)

I thought I had fixed this problem in #125: where the `nan` appear in the solution if duplicates are present is predictable, so `trim_t_results_jax` was written to avoid these values. However, #147 was created after discovering that this change made the calculation of gradients of solutions return `nan`. I think this is because the gradient computation of `odeint` will always utilize all intermediate computed solution values, so even though we were removing the `nan`'s, the gradient computation was still hitting them, which is something we couldn't avoid. This PR fixes this by changing the "merging strategy" of `merge_t_args` and `trim_t_results_jax`. Rather than just appending `t_span` to the ends of `t_eval`, we append them, then, conditional on if the endpoints are equal, modify the resulting combined array so that there are no duplicates, then correctly choosing the right entry when trimming the results. Importantly, this will not fundamentally impact the solver behaviour.

### DIFF
--- a/qiskit_dynamics/solvers/solver_utils.py
+++ b/qiskit_dynamics/solvers/solver_utils.py
@@ -135,12 +135,12 @@ def merge_t_args_jax(t_span, t_eval=None):
     if t_eval is None:
         return Array(t_span, backend="jax")
 
+    t_span = Array(t_span, backend="jax").data
+    t_eval = Array(t_eval, backend="jax").data
+
     # raise error if not one dimensional
     if t_eval.ndim > 1:
         raise ValueError("t_eval must be 1 dimensional.")
-
-    t_span = Array(t_span, backend="jax").data
-    t_eval = Array(t_eval, backend="jax").data
 
     out = jnp.append(jnp.append(t_span[0], t_eval), t_span[1])
 
--- a/qiskit_dynamics/solvers/solver_utils.py
+++ b/qiskit_dynamics/solvers/solver_utils.py
@@ -127,8 +127,9 @@ def merge_t_args_jax(t_span, t_eval=None):
     jnp.nan to signal errors. The merging strategy differs from merge_t_args: after appending
     the endpoints of t_span to t_eval, it is checked whether the first two entries of the resulting
     array are equal. If they are, the second entry is set to the average of the first and the third.
-    A similar procedure is done for the last two entries. This is essentially a hack to to avoid 
-    adjacent entries of the output having equal values, which causes buggy behaviour in jax_odeint.
+    A similar procedure is done for the last two entries. This is essentially a hack to to avoid
+    adjacent entries of the combined t_span, t_eval array having equal values, which causes buggy
+    behaviour in jax_odeint.
     """
 
     if t_eval is None:
@@ -159,20 +160,10 @@ def merge_t_args_jax(t_span, t_eval=None):
     out = cond(jnp.any(t_direction * diff < 0.0), lambda s: jnp.nan * s, lambda s: s, out)
 
     # if out[0] == out[1], set out[1] == (out[2] + out[0])/2
-    out = cond(
-        out[0] == out[1],
-        lambda x: x.at[1].set((x[2] + x[0])/2),
-        lambda x: x,
-        out
-    )
+    out = cond(out[0] == out[1], lambda x: x.at[1].set((x[2] + x[0]) / 2), lambda x: x, out)
 
     # if out[-1] == out[-2], set out[-2] == (out[-3] + out[-1])/2
-    out = cond(
-        out[-1] == out[-2],
-        lambda x: x.at[-2].set((x[-3] + x[-1])/2),
-        lambda x: x,
-        out
-    )
+    out = cond(out[-1] == out[-2], lambda x: x.at[-2].set((x[-3] + x[-1]) / 2), lambda x: x, out)
 
     return Array(out)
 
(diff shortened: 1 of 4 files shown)


qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -124,7 +124,12 @@ def trim_t_results(
 
 def merge_t_args_jax(t_span, t_eval=None):
```
ihincks, line 125: This function looks like it requires assumptions on the input to not run into IndexError. Perhaps these requirements could be documented. The arguments and return are also undocumented.

qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -154,26 +159,44 @@ def merge_t_args_jax(t_span, t_eval=None):
...
+    out = cond(out[-1] == out[-2], lambda x: x.at[-2].set((x[-3] + x[-1]) / 2), lambda x: x, out)
+
     return Array(out)
 
 
 def trim_t_results_jax(results, t_eval):
-    """JAX-compilable version of trim_t_results. The cond call is necessary due to
-    a bug in odeint that only occurs if the first two time values are duplicates.
+    """JAX-compilable version of trim_t_results. Note the choice of which entry to remove in the
+    case of duplicate time entries is due to peculiarities in jax_odeint.
```
ihincks, line 173: wrap jax_odeint in double ticks or a `:func:` role

qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -154,26 +159,44 @@ def merge_t_args_jax(t_span, t_eval=None):
...
+    # if out[0] == out[1], set out[1] == (out[2] + out[0])/2
+    out = cond(out[0] == out[1], lambda x: x.at[1].set((x[2] + x[0]) / 2), lambda x: x, out)
+
+    # if out[-1] == out[-2], set out[-2] == (out[-3] + out[-1])/2
+    out = cond(out[-1] == out[-2], lambda x: x.at[-2].set((x[-3] + x[-1]) / 2), lambda x: x, out)
+
     return Array(out)
 
 
 def trim_t_results_jax(results, t_eval):
```
ihincks, line 171: This function is missing argument and return docs

qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -154,26 +159,44 @@ def merge_t_args_jax(t_span, t_eval=None):
...
-    """JAX-compilable version of trim_t_results. The cond call is necessary due to
-    a bug in odeint that only occurs if the first two time values are duplicates.
+    """JAX-compilable version of trim_t_results. Note the choice of which entry to remove in the
+    case of duplicate time entries is due to peculiarities in jax_odeint.
     """
 
     if t_eval is None:
         return results
 
+    # remove second entry if t_eval[0] == results.t[0], as this indicates this was a repeated time
```
ihincks, line 179: These behavioural changes appear to be untested. Does this function not have tests to begin with?

qiskit_dynamics/solvers/solver_utils.py
DanPuzzuoli: did the `:func:` role.

qiskit_dynamics/solvers/solver_utils.py
DanPuzzuoli: Added arg/return docs.

qiskit_dynamics/solvers/solver_utils.py
DanPuzzuoli: Added documentation that `t_span` must be a list/tuple/array with two entries. Added Args/Return docs.

qiskit_dynamics/solvers/solver_utils.py
DanPuzzuoli: It did already have tests through inheritance, but yeah this specific way that it's picking entries hadn't been tested. Just added one to directly test this.

ihincks (approved):

PR #155, opened 2022-11-24
Adds `meas_level=1` for `DynamicsBackend`

rupeshknn:
### Summary
Support for `DynamicsBackend` to generate IQ data using `meas_level=1` option.

### Details and comments
TODOs:
- [x] check doc strings
- [x] reno
- [x] test 2 qubit IQ
- [x] Test mixed system dims


rupeshknn:
TODOs:
- [ ] check doc strings
- [ ] reno
- [ ] test 2 qubit IQ

rupeshknn:
@DanPuzzuoli can you review this?

rupeshknn:
@DanPuzzuoli can you review this?

qiskit_dynamics/backend/backend_utils.py
```diff
@@ -167,3 +170,67 @@ def _sample_probability_dict(
...
     """Count items in list."""
     return dict(zip(*np.unique(samples, return_counts=True)))
+
+
+def _iq_data(
+    options: Options,
+    state: Union[Statevector, DensityMatrix],
+    measurement_subsystems: List[int],
+    seed: int,
+) -> List[List[List[float]]]:
```
DanPuzzuoli, line 180: I think for functions at this level you should make the specific fields of `options` they require explicit arguments. So make `subsystem_dims`, `iq_centers`, and `iq_widths` args of this function.

DanPuzzuoli: Another question for this function is: do the memory slots play any role here? Can the IQ data be put into different orderings (other than the explicit qubit ordering)?

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -73,8 +74,15 @@ class DynamicsBackend(BackendV2):
       outcome probabilities. Defaults to ``True``. Setting to ``False`` can result in errors if
       the solution tolerance results in probabilities with significant numerical deviation from
       proper probability distributions.
-    * ``meas_level``: Form of measurement return. Only supported value is ``2``, indicating that
-      counts should be returned. Defaults to ``meas_level==2``.
+    * ``meas_level``: Form of measurement output. Suported values are ``1`` and ``2``.
+      ``1`` returns IQ points and ``2`` returns counts. Defaults to ``meas_level==2``.
+    * ``meas_return``: Level of measurement data to return. ``"single"`` returns information from
```
DanPuzzuoli, line 79: This only impacts `meas_level == 1` right? If yes can you make this doc line state this?

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -73,8 +74,15 @@ class DynamicsBackend(BackendV2):
...
+      every shot. ``"avg"`` returns average over shots of measurement output. Defaults to ``"avg"``
+    * ``iq_centers``: List[List[Tuple]] containing centers for IQ distribution. provided in the
+      format ``iq_centers[subsystem][level] = (I,Q)``. Defaults to equally spaced points on a
+      unit circle for each subsystem.
+    * ``iq_width``: Standard deviation of IQ distribution around the centers.
+      Must be positivie float. Defaults to ``0.2``.
```
DanPuzzuoli, line 85:
```suggestion
      Must be a positive float. Defaults to ``0.2``.
```

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -249,7 +266,7 @@ def run(
 
         # Configure run options for simulation
         if options:
-            backend = copy.copy(self)
+            backend = copy.deepcopy(self)
```
DanPuzzuoli, line 269: What's the reason for this change? Is it because the newly added fields contain nested mutable objects?

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -438,6 +456,29 @@ def default_experiment_result_function(
...
+        measurement_data = _iq_data(
+            backend.options,
+            yf,
+            measurement_subsystems=measurement_subsystems,
+            seed=seed,
+        )
+
+        if backend.options.meas_return == MeasReturnType.AVERAGE:
+            measurement_data = np.average(measurement_data, axis=0)
+
```
DanPuzzuoli, line 470: Does some rearranging of the measurement data using `memory_slot_indices` have to be done here?

test/dynamics/backend/test_backend_utils.py
```diff
@@ -211,3 +214,27 @@ def test_basic_counting(self):
         samples = ["00", "01", "00", "20", "01", "01", "20"]
         output = _get_counts_from_samples(samples)
         self.assertDictEqual(output, {"00": 2, "01": 3, "20": 2})
+
+
+class Test_iq_data(QiskitDynamicsTestCase):
```
DanPuzzuoli, line 219: Add some multi qubit iq data generation testing (including rearranging w.r.t. `memory_slot_indices` if necessary).

qiskit_dynamics/backend/backend_utils.py
```diff
@@ -167,3 +170,67 @@ def _sample_probability_dict(
...
+        sub_i, sub_q = [], []
+        for idx, count_i in enumerate(counts_n):
+            sub_i.append(
+                rng.normal(loc=iq_centers[sub_idx][idx][0], scale=options.iq_width, size=count_i)
+            )
+            sub_q.append(
+                rng.normal(loc=iq_centers[sub_idx][idx][1], scale=options.iq_width, size=count_i)
+            )
+        # Linear stack
+        sub_i = list(chain.from_iterable(sub_i))
```
DanPuzzuoli, line 230: Is this usage of `chain` really necessary? Won't `sub_i` and `sub_q` be lists of arrays? Not sure what's more efficient, but you could build `sub_i` and `sub_q` as arrays the whole way through using `np.concatenate`.

DanPuzzuoli (commented):
Nice, looking good, very clean addition. I've added a few comments throughout. Other than those:
- I'd like to see some tests actually checking the iq data values themselves (rather than just the outcomes of the "predict" or "discriminator" functions applied to the iq data). This will test that the values themselves are being generated correctly. Some testing of higher level outcomes would be nice too.
- Delete the release note file. As the entire `DynamicsBackend` is going to be new for this upcoming release, we'll write the release note related to this right at the end, so that we can write it in a cohesive way taking into account all of the features that ended up making the cut.

Other than these comments it's looking good.

qiskit_dynamics/backend/backend_utils.py
rupeshknn: Yeah, I don't remember why I was sticking to lists previously, I'll change them to arrays.

qiskit_dynamics/backend/dynamics_backend.py
rupeshknn: So, when I do, `job1 = backend.run(circuits, meas_level=1)` followed by `job2 = backend.run(circuits)` it continues to give `meas_level=1` data. A real backend does not have this behaviour. Options given to `run` only affect that particular job.

qiskit_dynamics/backend/dynamics_backend.py
rupeshknn: yeah

qiskit_dynamics/backend/backend_utils.py
rupeshknn: I tested some variations and `np.ravel` was faster than  `np.concatenate` so I used that instead of chain. I have not changed the entire code block to use arrays since `list.append` is cleaner than `x = np.append(x, val)`. Is that alright?

qiskit_dynamics/backend/backend_utils.py
rupeshknn: So I ran some circuits on hardware, and yes, the measure order does make a difference. Also, for unused `cl_bits`, we get an array of `0+0j` with the length of shots. Should we have similar behaviour?

rupeshknn: I was trying to understand the code (here)[https://github.com/rupeshknn/qiskit-dynamics/blob/19482c5d62e918d72a551791164f3ed67bee3ff4/qiskit_dynamics/backend/backend_utils.py#L142-L145]
```
if memory_slot_result in memory_slot_probs:
        memory_slot_probs[memory_slot_result] += prob
else:
        memory_slot_probs[memory_slot_result] = prob
```
the code under `else` is clear, but what are we expecting to do with the `if` code

DanPuzzuoli: The loop this lives in is translating the probabilities of observing the state in different computational basis states (limited to some subsystems) into probabilities for memory slot outcomes. So, e.g. if the state-outcome string is `'01'`, and the first subsystem is being mapped to memory slot 2, and the second to memory slot 0, (with memory slot 0 being unused), the corresponding memory slot string is then `'100'` (where all of these strings are in reverse subsystem ordering).

The reason for the `if` statement is that if there's a max outcome level, then the mapping of state-outcome to measurement_slot-outcome is not 1 to 1. E.g. if the max outcome level is `1`, the state-outcome string `'02'` will also be mapped to the memory-slot string `'100'`. So the if statement is checking if the string already exists in the dictionary of memory slot outcome probabilities: if it doesn't then put it in, and if it does exist, add the current probability to the existing probability.

This all being said, this block could be reduced to the single line:
```
memory_slot_probs[memory_slot_result] = prob + memory_slot_probs.get(memory_slot_result, 0)
```

qiskit_dynamics/backend/backend_utils.py
DanPuzzuoli: yep. Just wanted to avoid unnecessary additional imports.

test/dynamics/backend/test_backend_utils.py
rupeshknn: done

qiskit_dynamics/backend/dynamics_backend.py
rupeshknn: done

qiskit_dynamics/backend/backend_utils.py
rupeshknn: done

rupeshknn:
changes done. @DanPuzzuoli, can you review it?

qiskit_dynamics/backend/backend_utils.py
```diff
@@ -173,44 +171,43 @@ def _get_counts_from_samples(samples: list) -> Dict:
 
 
 def _iq_data(
```
DanPuzzuoli, line 173: pretty minor but can you change this name to `_get_iq_data` to be consistent with the other data generating functions in this file

DanPuzzuoli (commented):
Just one minor comment - I've validated the changes you've made based on my previous comments. Is there anything else that needs to be done/added to this? If no I can do a final holistic review.

DanPuzzuoli (changes requested):
A few more comments, mostly to do with validation/structure.

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -182,14 +194,20 @@ def set_options(self, **fields):
...
+                raise QiskitError("meas_return must be either 'single' or 'avg'.")
             elif key == "max_outcome_level":
                 if (value is not None) and (not isinstance(value, int) or (value <= 0)):
                     raise QiskitError("max_outcome_level must be a positive integer or None.")
             elif key == "experiment_result_function" and not callable(value):
                 raise QiskitError("experiment_result_function must be callable.")
-
+            elif key == "iq_width" and (not isinstance(value, float) or (value <= 0)):
+                raise QiskitError("iq_width must be positive float.")
+            elif key == "iq_centers":
```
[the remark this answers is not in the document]
rupeshknn: Why should we have `validate_iq_centers  = True` in the `elif` block of `subsystem_dims` and then ensure it works for `iq_centers =None`? 
What use case are we trying to convert here?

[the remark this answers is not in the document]
rupeshknn: Changing it to an array might not work as the length for each subsystem will be different, creating an array of `dtype=object`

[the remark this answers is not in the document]
rupeshknn: I am thinking something like this:

```python
elif key == "iq_centers":
    validate_iq_centers = True
    self._options.update_options(**{key: value})
```
then:

```python
if validate_iq_centers:
    iq_centers = self._options.iq_centers
    if ([len(x) for x in iq_centers] != self._options.subsystem_dims) or (
        np.concatenate(iq_centers).shape[1] != 2
    ):
        raise QiskitError("iq_centers is not consistent with subsystem_dims.")
```

qiskit_dynamics/backend/backend_utils.py
```diff
@@ -167,3 +168,66 @@ def _sample_probability_dict(
...
+    Returns:
+        (I,Q) data as ndarray[shot index, qubit index] = [I,Q]
+
+    Raises:
+        QiskitError: If number of centers and levels don't match.
+    """
+    rng = np.random.default_rng(seed)
+
+    full_i, full_q = [], []
+    for sub_idx in measurement_subsystems:
```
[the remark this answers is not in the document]
rupeshknn: Using
```python
p = state.probabilities()
dim = state.dims()
ps = [Statevector._subsystem_probabilities(p, dim, [idx]) for idx in range(len(dim))]
```
Saves some time

qiskit_dynamics/backend/dynamics_backend.py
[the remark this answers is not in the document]
DanPuzzuoli: > Changing it to an array might not work as the length for each subsystem will be different, creating an array of `dtype=object`

Ah yeah true, good point.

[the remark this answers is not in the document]
DanPuzzuoli: This block:
```
elif key == "iq_centers":
    validate_iq_centers = True
    self._options.update_options(**{key: value})
```
Should need the `self._options.update_options(**{key: value})` as it will be handled by the general `else` of:
```
if key == "solver":
    self._set_solver(value)
    validate_subsystem_dims = True
else:
    if key == "subsystem_dims":
        validate_subsystem_dims = True
    self._options.update_options(**{key: value})
```

Once the `if` statement for `key == "subsystem_dims"` is moved to the upper if block. I think also we should also move the line `validate_subsystem_dims = True` into the upper block by adding a `elif key == "solver"` to that section.

Sorry I realize this is somewhat arbitrary, and this method is just validating and not doing any computation, but this nitpicking is in an attempt to clearly delineate what's happening in each code block of this method, so that it's easier to read in the future (since it's all validation/option setting, I have a feeling this could potentially get confusing in the future if we're not careful). What I'm thinking for the block of this method is: 

1. The first if block does immediate type and independent value validation (i.e. validating that the value of the option being set is well-formed), while also setting flags to do consistency validation later if the option needs to be consistent with other option values.
2. The second block sets the option - the main thing is the `else` block doing general option setting, but some options may have custom setting routines. In this case only the `solver` has a custom routine.
3. The final block does consistency validation based on which flags were raised in the first block.

Make the changes you want to make and try to follow the above - depending on how it looks, afterwards it may be more expedient for me to just open a PR to yours to make this method exactly how I want (since this really is mostly about future-me's ability to read it after forgetting the details :D ).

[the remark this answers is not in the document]
DanPuzzuoli: This is almost there:

```
if validate_iq_centers:
    iq_centers = self._options.iq_centers
    if ([len(x) for x in iq_centers] != self._options.subsystem_dims):
        raise QiskitError("iq_centers is not consistent with subsystem_dims.")
    if not all([len(level)==2 for sub_system in iq_centers for level in sub_system]):
        raise QiskitError("iq_centers must be 2-tuples.")
```

changes:

1. Make sure `iq_centers is None` is accepted, so either envelope everything in another `if` statement that only does further validation if `iq_centers is not None`, or maybe add it to the overarching if statement, e.g. `if validate_iq_centers and self._options.iq_centers is not None`.
2. I think this `if` statement: `if not all([len(level)==2 for sub_system in iq_centers for level in sub_system])` needs to be moved into the first block as described above, since it's not part of a consistency check, it's just part of the check that `iq_centers` is in the `List[List[List[float,float]]]` format. You also need to be careful here that you're assuming you can call `len` on `level` here.

The first `if` statement is fine, assuming that you've already properly validated the `List[List[List[float,float]]]` format at this point.

qiskit_dynamics/backend/backend_utils.py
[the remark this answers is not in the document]
DanPuzzuoli: Nice - yeah that looks good. You're right though that we shouldn't use a private method from another package. You can just copy this static method into a standalone function in `backend_utils.py`. Make sure you grab the latest current corrected version on terra `main` :).

qiskit_dynamics/backend/backend_utils.py
```diff
@@ -167,3 +168,92 @@ def _sample_probability_dict(
 def _get_counts_from_samples(samples: list) -> Dict:
     """Count items in list."""
     return dict(zip(*np.unique(samples, return_counts=True)))
+
+
+def _get_subsystem_probabilities(probability_tensor: np.ndarray, sub_idx: int) -> np.ndarray:
+    """Marginalize a probability vector according to subsystems.
```
DanPuzzuoli, line 174: Can you put a note here that this is a temporary replacement for `QuantumState._subsystem_probabilities`. Or have you made changes now to make this even more efficient than if you used that method?

Also, make this doc string a bit more clear, something like: `Marginalize a probability array to a single subsystem.`.

qiskit_dynamics/backend/backend_utils.py
```diff
@@ -167,3 +168,92 @@ def _sample_probability_dict(
...
+
+    Args:
+        probability_tensor: Probability full system reshaped to reversed(subsystem_dims).
```
DanPuzzuoli, line 177: This doc string is unclear:

```suggestion
        probability_tensor: Probability full system reshaped to reversed(subsystem_dims).
```

This method doesn't know about `subsystem_dims`. Maybe say somethign like: 

```
k-dimensional probability array, where the probability of outcome `(idx1, ..., idxk)` is `probability_tensor[idx1, ..., idxk]`. 
```
I guess you also need to be clear that the `sub_idx` itself is in reverse order (which is what your original comment was alluding to).

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -182,23 +196,37 @@ def set_options(self, **fields):
...
+            elif key == "meas_return" and value not in ["single", "avg"]:
+                raise QiskitError("meas_return must be either 'single' or 'avg'.")
             elif key == "max_outcome_level":
                 if (value is not None) and (not isinstance(value, int) or (value <= 0)):
                     raise QiskitError("max_outcome_level must be a positive integer or None.")
             elif key == "experiment_result_function" and not callable(value):
                 raise QiskitError("experiment_result_function must be callable.")
+            elif key == "iq_width" and (not isinstance(value, float) or (value <= 0)):
+                raise QiskitError("iq_width must be positive float.")
+            elif key == "iq_centers":
```
DanPuzzuoli, line 210: This also needs to accept `value is None` right? That is a valid value for `iq_centers`. You should also check that `isinstance(iq_centers, list)`.

The error message here should be more informative and follow proper capitalization:
```
"The iq_centers option must be either None or of type List[List[List[int, int]]], where the outer most list is of length equal to the number of subsystems, and each inner list of length equal to the corresponding subsystem dimension."
```

DanPuzzuoli (changes requested):
Some final very minor comments.

qiskit_dynamics/backend/backend_utils.py
rupeshknn: It's only a tiny bit more efficient than `QuantumState._subsystem_probabilities` but I guess the main reason for bringing it here is so we don't use a private static method.

qiskit_dynamics/backend/backend_utils.py
rupeshknn: done

qiskit_dynamics/backend/dynamics_backend.py
rupeshknn: Here we only check if the IQ points are lists of 2 elements. Not using any info about the number of subsystems or their dimensions. So maybe for this check, we can have:
```
"The iq_centers option must be either None or of type List[List[List[int, int]]], where the innermost list is the (I, Q) pair."
```` 
And have the error message you suggested above for the consistency check.

DanPuzzuoli (approved):

PR #156, opened 2022-11-24
Tidy the models folder

ihincks:
### Summary

This PR attempts to clean up documentation and style in the models folder.

### DIFF
--- a/qiskit_dynamics/models/generator_model.py
+++ b/qiskit_dynamics/models/generator_model.py
@@ -122,7 +122,7 @@ class BaseGeneratorModel(ABC):
             y: Optional state.
 
         Returns:
-            Array: Either the evaluated model, or the RHS for the given y
+            Array: Either the evaluated model, or the RHS for the given y.
         """
         return self.evaluate(time) if y is None else self.evaluate_rhs(time, y)
 
--- a/qiskit_dynamics/models/generator_model.py
+++ b/qiskit_dynamics/models/generator_model.py
@@ -216,9 +216,6 @@ class GeneratorModel(BaseGeneratorModel):
 
     @evaluation_mode.setter
     def evaluation_mode(self, new_mode: str):
-        if new_mode not in ("dense", "sparse"):
-            raise NotImplementedError("The evaluation_mode must be 'dense' or 'sparse'.")
-
         if new_mode != self._evaluation_mode:
             self._operator_collection = construct_operator_collection(
                 new_mode,
--- a/qiskit_dynamics/models/generator_model.py
+++ b/qiskit_dynamics/models/generator_model.py
@@ -42,11 +42,10 @@ except ImportError:
 
 
 class BaseGeneratorModel(ABC):
-    r"""Defines an interface for a time-dependent
-    linear differential equation of the form :math:`\dot{y}(t) = \Lambda(t, y)`,
-    where :math:`\Lambda` is linear in :math:`y`. The core functionality is
-    evaluation of :math:`\Lambda(t, y)`, as well as,
-    if possible, evaluation of the map :math:`\Lambda(t, \cdot)`.
+    r"""Defines an interface for a time-dependent linear differential equation of the form
+    :math:`\dot{y}(t) = \Lambda(t, y)`, where :math:`\Lambda` is linear in :math:`y`. The core
+    functionality is evaluation of :math:`\Lambda(t, y)`, as well as, if possible, evaluation of the
+    map :math:`\Lambda(t, \cdot)`.
 
     Additionally, the class defines interfaces for:
 
@@ -61,19 +60,17 @@ class BaseGeneratorModel(ABC):
     @property
     @abstractmethod
     def dim(self) -> int:
-        """Gets matrix dimension."""
+        """The matrix dimension."""
 
     @property
     @abstractmethod
     def rotating_frame(self) -> RotatingFrame:
-        """Get the rotating frame."""
+        """The rotating frame."""
 
     @rotating_frame.setter
     @abstractmethod
     def rotating_frame(self, rotating_frame: RotatingFrame):
-        """Set the rotating frame; either an already instantiated :class:`RotatingFrame` object,
-        or a valid argument for the constructor of :class:`RotatingFrame`.
-        """
+        pass
 
     @property
     @abstractmethod
@@ -83,7 +80,7 @@ class BaseGeneratorModel(ABC):
     @in_frame_basis.setter
     @abstractmethod
     def in_frame_basis(self, in_frame_basis: bool):
-        """Set whether to evaluate in the basis in which the frame is diagonalized."""
+        pass
 
     @property
     @abstractmethod
@@ -93,14 +90,14 @@ class BaseGeneratorModel(ABC):
     @evaluation_mode.setter
     @abstractmethod
     def evaluation_mode(self, new_mode: str):
-        """Sets evaluation mode of model."""
+        pass
 
     @abstractmethod
     def evaluate(self, time: float) -> Array:
         r"""If possible, evaluate the map :math:`\Lambda(t, \cdot)`.
 
         Args:
-            time: Time.
+            time: The time to evaluate the model at.
         """
 
     @abstractmethod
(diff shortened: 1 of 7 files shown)


qiskit_dynamics/models/operator_collections.py
```diff
@@ -33,104 +33,126 @@
...
         self.static_operator = static_operator
 
+    @property
+    def dim(self) -> int:
+        """The matrix dimension."""
+        if self.static_operator is not None:
+            return self.static_operator.shape[-1]
+        else:
+            return self.operators[0].shape[-1]
+
```
DanPuzzuoli, line 85: Nice, I like the move of this into the collections.

qiskit_dynamics/models/operator_collections.py
```diff
@@ -33,104 +33,126 @@
...
+            The evaluated function.
+        """
 
     def __call__(
         self, signal_values: Union[List[Array], Array], y: Optional[Array] = None
     ) -> Array:
-        """Call either ``self.evaluate`` or ``self.evaluate_rhs`` depending on number of
-        arguments.
-        """
+        """Call :meth:`~evaluate` or :meth:`self.evaluate_rhs` depending on the presense of ``y``.
```
DanPuzzuoli, line 128: What's the difference between using the tilde and directly referencing `self.`?

qiskit_dynamics/models/operator_collections.py
```diff
@@ -314,6 +358,17 @@ def evaluate(self, signal_values: Union[Array, None]) -> "BCOO":
...
+        r"""Evaluate the function and return :math:`\Lambda(c, y) = (G_d + \sum_jc_jG_j)  y`.
+
+        Args:
+            signal_values: The signals :math:`c` to use on the operators.
+            y: The system state.
+
+        Returns:
+            The evaluated function.
+        Raises:
+            QiskitError: If the function cannot be evaluated.
```
DanPuzzuoli, line 370:
```suggestion
        Returns:
            The evaluated function.
            
        Raises:
            QiskitError: If the function cannot be evaluated.
```

qiskit_dynamics/models/operator_collections.py
```diff
@@ -879,28 +984,30 @@ class JAXSparseLindbladCollection(BaseLindbladOperatorCollection):
     """
 
     @property
-    def static_hamiltonian(self) -> "BCOO":
+    def static_hamiltonian(self) -> BCOO:
```
DanPuzzuoli, line 987: Ah I guess this answers the question as to why you need to define `BCOO` even if JAX can't be imported.

qiskit_dynamics/models/operator_collections.py
```diff
@@ -945,15 +1053,30 @@ def dissipator_operators(self, new_dissipator_operators: Union["BCOO", None]):
...
+
+        Args:
+            ham_sig_vals: The signals :math:`c_1` to use on the Hamiltonians.
+            dis_sig_vals: The signals :math:`c_2` to use on the dissipators.
+
+        Returns:
+            The evaluated function.
+
+        Raises:
+            ValueError: Always.
```
DanPuzzuoli, line 1067: Nice.

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -307,27 +297,27 @@ def _conjugate_and_add(
...
-        be passed as (k, dim, dim) Array objects, with the :math:`i^{th}` operator
-        being stored as the [i,:,:] entry. For vectorized_operators = True, these
-        (vectorized) operators should be passed as a (dim**2, k) Array, with the
-        :math:`i^{th}` vectorized operator stored as the [:,i] entry.
+        Note that are two conventions for passing multiple operators at the same time. For
+        evaluation with vectorized_operators=False, these operators should be passed as ``(k, dim,
+        dim)`` Array objects, with the :math:`i^{th}` operator being stored as the ``[i,:,:]``
+        entry. For vectorized_operators = True, these (vectorized) operators should be passed as a
+        ``(dim**2, k)`` Array, with the :math:`i^{th}` vectorized operator stored as the ``[:,i]``
+        entry.
```
DanPuzzuoli, line 308:
```suggestion
        Note that there are two conventions for passing multiple operators at the same time. For
        evaluation with ``vectorized_operators=False``, these operators should be passed as ``(k, dim,
        dim)`` array objects, with the :math:`i^{th}` operator being stored as the ``[i,:,:]``
        entry. For ``vectorized_operators = True``, these (vectorized) operators should be passed as a
        ``(dim**2, k)`` array, with the :math:`i^{th}` vectorized operator stored as the ``[:,i]``
        entry.
```

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -507,21 +494,19 @@ def generator_out_of_frame(
         operator_in_frame_basis: Optional[bool] = False,
         return_in_frame_basis: Optional[bool] = False,
     ) -> Array:
-        r"""Take an operator out of the frame, i.e. return
-        ``exp(tF) @ operator @ exp(-tF) + F``.
+        r"""Take an operator out of the frame, i.e. return ``exp(tF) @ operator @ exp(-tF) + F``.
```
DanPuzzuoli, line 497:
```suggestion
        r"""Take an operator out of the frame using the generator transformation rule, i.e. return ``exp(tF) @ operator @ exp(-tF) + F``.
```

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -610,28 +595,28 @@ def vectorized_map_into_frame(
         return op
 
 
-def _is_herm_or_anti_herm(mat: Array, atol: Optional[float] = 1e-10, rtol: Optional[float] = 1e-10):
+def _enforce_anti_herm(mat: Array, atol: Optional[float] = 1e-10, rtol: Optional[float] = 1e-10):
```
DanPuzzuoli, line 598: Should this be ``_enforce_herm_or_anti_herm``?

DanPuzzuoli (changes requested):
A couple code suggestions and minor comments.

qiskit_dynamics/models/operator_collections.py
ihincks: The discrepancy here is indeed unwarranted, will fix. (tilde suppresses ptah of method:`qiskit_dynamics.models.OperatorCollection.evaluate` vs just `evaluate`)

qiskit_dynamics/models/generator_model.py
```diff
@@ -272,13 +261,16 @@ def rotating_frame(self, rotating_frame: Union[Operator, Array, RotatingFrame]):
 
     @property
     def signals(self) -> SignalList:
-        """Return the signals in the model."""
+        """The signals in the model.
+
+        Raises:
```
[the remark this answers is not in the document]
ihincks: I think we can define this library-wide ourselves, and I'm not aware of any strong conventions, so I'll leave the call to you. 

I think we did, however, already chat about this in more generality: I think we decided that the getter docstring can talk about the setter, as long as its done so in a clear way. Keep in mind that the setter and the getter are two methods of the same `property` instance, so the getter docstring, imo, should just be thought of as the docstring of the shared instance.

qiskit_dynamics/models/rotating_frame.py
ihincks: IMO _enforce_anti_herm is correct because this function enforces that the output be anti-Hermitian with any input. _is_herm_or_anti_herm makes it sound like it'll return True or something if the input is either. _enforce_herm_or_anti_herm sounds like it will enforce one of those two things.

qiskit_dynamics/models/lindblad_model.py
```diff
@@ -314,37 +309,37 @@ def in_frame_basis(self, in_frame_basis: bool):
...
 
     def _get_static_hamiltonian(self, in_frame_basis: Optional[bool] = False) -> Array:
-        """Get the constant hamiltonian term.
+        """Get the constant Hamiltonian term.
 
         Args:
-            in_frame_basis: Flag for whether the returned static_operator should be
-            in the basis in which the frame is diagonal.
+            in_frame_basis: Flag for whether the returned static_operator should be in the basis
+            in which the frame is diagonal.
```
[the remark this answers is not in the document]
ihincks: just barely works

qiskit_dynamics/models/rotating_frame.py
ihincks: fixed manually for proper newlines

DanPuzzuoli (approved):

DanPuzzuoli (approved):

PR #163, opened 2022-12-13
Make stable/0.3 compatible with JAX 0.4.0

DanPuzzuoli:
### Summary

Reproduction of #162 for stable 0.3.


DanPuzzuoli:
Not sure why this is failing but #162 passed, given that `stable/0.3` does work for all JAX versions < 0.4.0.

ihincks (approved):

PR #167, opened 2022-12-13
Use opflow operators in dynamics

donsano33:
### Summary

Add support and type hints, so that qiskit.opflow operators inheriting from qiskit.opflow.OperatorBase can be used to construct a HamiltonianModel and a RotatingFrame.

### Details and comments

It kind of gets messy with the type hints, not sure what to do about that.

Needs the "New Feature" Label, not sure if I can do this myself.

### DIFF
--- a/qiskit_dynamics/models/hamiltonian_model.py
+++ b/qiskit_dynamics/models/hamiltonian_model.py
@@ -22,6 +22,7 @@ from scipy.sparse.linalg import norm as spnorm
 
 from qiskit import QiskitError
 from qiskit.quantum_info.operators import Operator
+from qiskit.opflow import OperatorBase
 from qiskit_dynamics.array import Array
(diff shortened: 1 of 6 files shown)


woodsp-ibm:
I will note that the current plan is to deprecate (next release) and later remove opflow from Terra Qiskit/qiskit-terra#9176

DanPuzzuoli:
> I will note that the current plan is to deprecate (next release) and later remove opflow from Terra [Qiskit/qiskit-terra#9176](https://github.com/Qiskit/qiskit-terra/pull/9176)

Ah good to know, thanks. 

@donsano33 based on this I think I will actually prefer to not merge this. Apologies for that. Even though it will be convenient in the short term, I think it will be more inconvenient when we run into the eventual known incompatibility of an upcoming version of terra.

PR #172, opened 2022-12-19
Fixing tests failing due to numpy eigh sign convention changes

### LINKED ISSUE #171: numpy 1.24 breaks some tests
The tests:
```
dynamics.models.test_generator_model.Testtransfer_operator_functionsJax.test_array_inputs_pseudo_random
```
and
```
dynamics.solvers.test_lanczos.TestLanczos.test_ground_state
```
now fail on `main`.

Both appear to be due to the tests depending on the sign of eigenvectors computed by `numpy.linalg.eigh`.
- The second failure is simple: the first eigenvector, as computed by two separate routines, are checked to be equal. The sign of this as computed by `numpy.linalg.eigh` has now flipped, raising an error. This can easily be fixed by multiplying both by the sign of the first element (which has to be real) before comparing.
- The first one is ultimately an error coming from the same place, but it's hidden a bit deeper in the function calls in the test case, and is arising due to new sign inconsistencies between the output of `numpy.linalg.eigh` and `jax.numpy.linalg.eigh`. (Both of course output mathematically correct values, but again, some of the eigenvectors output by `numpy` now have a flipped sign.) I think this can be fixed by more explicitly controlling the `RotatingFrame` objects being constructed in this test, so that the explicit diagonalizing basis details don't impact the checks in the test.

DanPuzzuoli:
### Summary

Closes #171.

As described in #171, a numpy update cause signs of eigenvectors output by `eigh` to change in some tests. While the output is mathematically correct in both cases, our tests were dependent on the exact value of the entries of the eigenvectors, leading to errors.

Both tests have been modified to be independent of the sign of the eigenvectors output of `eigh`, avoiding this issue.

### DIFF
--- a/test/dynamics/models/test_generator_model.py
+++ b/test/dynamics/models/test_generator_model.py
@@ -762,8 +762,7 @@ class Testtransfer_operator_functions(QiskitDynamicsTestCase):
         new_frame = rng.uniform(low=-b, high=b, size=(dim, dim)) + 1j * rng.uniform(
             low=-b, high=b, size=(dim, dim)
         )
-        new_frame = new_frame - new_frame.conj().transpose()
-        new_frame = RotatingFrame(new_frame)
+        new_frame = RotatingFrame(new_frame - new_frame.conj().transpose())
 
         out_static = transfer_static_operator_between_frames(
             Array(static_operator), new_frame=new_frame, old_frame=old_frame
--- a/test/dynamics/models/test_generator_model.py
+++ b/test/dynamics/models/test_generator_model.py
@@ -756,8 +756,7 @@ class Testtransfer_operator_functions(QiskitDynamicsTestCase):
         old_frame = rng.uniform(low=-b, high=b, size=(dim, dim)) + 1j * rng.uniform(
             low=-b, high=b, size=(dim, dim)
         )
-        old_frame = old_frame - old_frame.conj().transpose()
-        old_frame = RotatingFrame(old_frame)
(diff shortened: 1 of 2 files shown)


test/dynamics/models/test_generator_model.py
```diff
@@ -758,28 +757,32 @@ def test_array_inputs_pseudo_random(self):
             low=-b, high=b, size=(dim, dim)
         )
         old_frame = old_frame - old_frame.conj().transpose()
+        old_frame = RotatingFrame(old_frame)
+
         new_frame = rng.uniform(low=-b, high=b, size=(dim, dim)) + 1j * rng.uniform(
             low=-b, high=b, size=(dim, dim)
         )
         new_frame = new_frame - new_frame.conj().transpose()
+        new_frame = RotatingFrame(new_frame)
```
ihincks, line 766:
```suggestion
        new_frame = RotatingFrame(new_frame - new_frame.conj().transpose())
```

test/dynamics/models/test_generator_model.py
```diff
@@ -758,28 +757,32 @@ def test_array_inputs_pseudo_random(self):
             low=-b, high=b, size=(dim, dim)
         )
         old_frame = old_frame - old_frame.conj().transpose()
+        old_frame = RotatingFrame(old_frame)
```
ihincks, line 760:
```suggestion
        old_frame = RotatingFrame(old_frame - old_frame.conj().transpose())
```

ihincks (approved):

PR #177, opened 2023-01-25
Update pulse conversion docs to be fully explicit

### LINKED ISSUE #168: Update documentation for InstructionToSignals to be much more explicit
With #164 , we're working towards making the `InstructionToSignals` class behave in a way more reflective of how schedules are implemented on real backends. As this behaviour is non-obvious and non-trivial, after #164, we should update the documentation for this class to be much more explicit about the rules its following to generate the samples used in simulation from schedules.

DanPuzzuoli:
### Summary

Closes #168

The goal with this PR is to improve the documentation for the pulse -> signal converter, to make it clear to a reader how pulse schedules are being interpreted.


qiskit_dynamics/pulse/__init__.py
```diff
@@ -19,18 +19,17 @@
...
+schedules into signals such that they can be simulated.
 
 Converters
 ==========
 
 The conversion from a pulse schedule to a list of signals is done with the
-:class:`InstructionToSignals` converter. The example below shows a schedule and
-the resulting signals converted using
+:class:`InstructionToSignals` converter. The example below shows a schedule and the resulting
+signals converted using
```
ihincks, line 32: I might put a few words between the codeblock and the jupyter execute to implicitly indicate that these two lines of code don't produce a plot, e.g. "which results in a schedule such as the following"

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -44,13 +44,15 @@ class InstructionToSignals:
     :class:`ScheduleBlock` must first be converted to :class:`Schedule` using the
```
ihincks, line 44: a few lines up, wrap Signals in a :class:, or lower-case it

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -79,14 +81,60 @@ def __init__(
         self._carriers = carriers or {}
 
     def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
-        """
+        r"""Convert a schedule to a corresponding list of DiscreteSignal instances.
```
ihincks, line 84: :class:

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -79,14 +81,60 @@ def __init__(
...
+        analog carrier frequency for each channel, which is fixed for the full duration. For a given
+        channel, the :math:`k^{th}` envelope sample for the corresponding :class:`.DiscreteSignal`
+        is determined according to the following formula:
+
+        .. math::
+            f_k \exp(i(2\pi \Delta\nu(k) k dt + \phi(k) + 2 \pi \phi_a(k))),
+
+        where:
+
+        * :math:`f_k` is the waveform value at the :math:`k^{th}` time step as specified by ``Play``
```
ihincks, line 97: Is there a reason `f_k` and `\phi(k)` use different indexing notation? Consider unifying if not.

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -79,14 +81,60 @@ def __init__(
...
+          instructions.
+        * :math:`\Delta\nu(k)` is the frequency deviation at time step :math:`k` from the analog
```
ihincks, line 99: Just to satisfy my own curiosity, \Delta\nu(k) is always fixed for the duration of a pulse, and only changes between pulses?

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -79,14 +81,60 @@ def __init__(
...
+          carrier as the result of ``SetFrequency`` and ``ShiftFrequency`` instructions. As evident
+          by the formula, carrier frequency deviations as a result of these instructions are handled
+          digitally, with the analog carrier frequency being fixed for the entirety of the schedule.
+        * :math:`dt` is the sample rate as specified by the ``dt`` instantiation argument.
+        * :math:`\phi(k)` is the channel phase at time step :math:`k`, as determined by
+          ``ShiftPhase`` and ``SetPhase`` instructions.
+        * :math:`\phi_a(k)` is the phase correction term at time step :math:`k`, impacted by
+          ``SetFrequency`` and ``ShiftFrequency`` instructions, described below.
```
ihincks, line 107: Does it say somewhere which \phi comes from the pulse profile? (rather than phase advancement instructions, etc)

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: yeah \Delta\nu(k) can only change with one of the `SetFrequency` or `ShiftFrequency` instructions, and they can't happen during `Play` instruction.

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: The `\phi` is purely determined by `ShiftPhase` and `SetPhase`, and `phi_a` is purely determined by `SetFrequency` and `ShiftFrequency` instructions. This is the only way to manipulate these parameters. The `Play` instructions themselves have no phase, they just use whatever the current channel phase is.

ihincks: Oh so f_k is complex-valued. Got it.

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: No great reason. I think I maybe did this because of the way I think about them due to internal implementation - the waveform samples (the `f_k`) is stored in an actual array that has an explicit value stored at every time step, whereas I chose a function notation for the phases and frequency difference as they're things that get periodically updated and not explicitly stored at every time step.

Changed `f_k` to `f(k)`.

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: I actually purposefully left this without `:class:` based on a rule I try to follow that the first line should be readable as plain text for when users call `help`. I'm sure this isn't followed everywhere in Dynamics.

qiskit_dynamics/pulse/__init__.py
DanPuzzuoli: Changed this.

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: lower cased it.

qiskit_dynamics/pulse/__init__.py
```diff
@@ -19,18 +19,17 @@
 
 .. currentmodule:: qiskit_dynamics.pulse
 
-This module contains tools to interface qiskit pulse with qiskit dynamics. Indeed,
-Qiskit-dynamics simulates time evolution using the :class:`Signal` class. However,
-Qiskit pulse specifies pulse instructions using a schedule. This module contains
-the required converters to convert Qiskit pulse schedules into signals such that
-they can be simulated.
+This module contains tools to interface qiskit pulse with qiskit dynamics. Indeed, Qiskit Dynamics
```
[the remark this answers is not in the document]
DanPuzzuoli: With help from @mtreinish I've just added the `intersphinx` extension to the docs and have directly linked to the main qiskit pulse API documentation, as well as some of the classes.

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: Ah I see what you're saying, and why this might be confusing then, though this is the convention used in qiskit pulse.

ihincks (approved):

PR #178, opened 2023-02-01
Adjustments to some parameter handling in DynamicsBackend

DanPuzzuoli:
### Summary

This PR contains several modifications to `DynamicsBackend` that fix small issues discovered while preparing a tutorial for the class.

### Details and comments

The changes:
- The default absolute tolerance on `_get_dressed_state_decomposition` has been reduced to `1e-5`, to be consistent with the Hermiticity checking in the `quantum_info` module. The previous absolute tolerance of `1e-8` was erroneously raising errors when dealing with Hamiltonians specified in Hz/s containing parameters in the GHz range.
    - No tests have been added for this as it's just a tolerance change.
- The way the default measurement instructions are built in the `DynamicsBackend` constructor has been changed to be more in line with my understanding of what the "right" way of adding instructions is: calling `target.add_instruction`. (Previously, the instructions were being added directly to the `instruction_schedule_map`, which wasn't properly populating all fields in the `target`).
    - No specific tests have been added for this, but all `DynamicsBackend` tests invariably indirectly test this.
- Experiment metadata is now being extracted and added to the `ExperimentResult` objects. `qiskit-experiments` uses metadata, and without this, analysis tasks were failing.
    - A new test has been added for this to check output metadata when submitting multiple circuits.
- `qiskit-experiments` expects jobs to have `time_per_step` method that returns time stamps, which `DynamicsJob` was not constructing. I've added some timestamp storing internally to `DynamicsJob` that can be retrieved by the `time_per_step`.
    - A test has been added confirm this and its behaviour. 
- Lastly, I found a bug in which, if a schedule being simulated has no instructions before measurement, the returned statevector would contain `nan` if using JAX and `jax_odeint`. This was due to `jax_odeint` being called with `t_span=[0., 0.]`, and this is another case in which `odeint` being called with non-unique time values, some of the returned states will be `nan`. I've updated `trim_t_results_jax` to detect when this is occurring and correct the returned state, which resolved the bug. (This came up as one of the experiments from `qiskit-experiments` appears to measure SPAM error by submitting a schedule with nothing occurring before measurement.)
    - I've added tests for both `trim_t_results_jax` and `jax_odeint` that would have detected this problem.
- This PR also adds a test that the `RoughFrequencyCal` experiment from `qiskit-experiments` can be run using `DynamicsBackend`, and returns a sensible result. This is contained in the new test file `test.dynamics.backend.test_qiskit_experiments`, and `qiskit-experiments` has been added to `requirements-dev.txt`.

### DIFF
--- a/test/dynamics/backend/test_qiskit_experiments.py
+++ b/test/dynamics/backend/test_qiskit_experiments.py
@@ -63,7 +63,6 @@ class TestExperimentsIntegration(QiskitDynamicsTestCase, TestJaxBase):
     def test_RoughFrequencyCal_calibration(self):
         """Test RoughFrequencyCal outputs a sensible answer."""
 
-        # cals = Calibrations.from_backend(self.simple_backend)
         cals = Calibrations()
 
         dur = Parameter("dur")
--- a/test/dynamics/backend/test_qiskit_experiments.py
+++ b/test/dynamics/backend/test_qiskit_experiments.py
@@ -66,14 +66,12 @@ class TestExperimentsIntegration(QiskitDynamicsTestCase, TestJaxBase):
         cals = Calibrations()
 
         dur = Parameter("dur")
-        amp = Parameter("amp")
         sigma = Parameter("sigma")
-        beta = Parameter("beta")
         drive = pulse.DriveChannel(Parameter("ch0"))
 
         # Define and add template schedules.
         with pulse.build(name="x") as x:
-            pulse.play(pulse.Drag(dur, amp, sigma, beta), drive)
+            pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("beta")), drive)
 
         with pulse.build(name="sx") as sx:
             pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("beta")), drive)
--- a/test/dynamics/backend/test_qiskit_experiments.py
+++ b/test/dynamics/backend/test_qiskit_experiments.py
@@ -1,0 +1,101 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+# pylint: disable=invalid-name
+
+"""
+Integration tests that ensure this module interacts properly with qiskit-experiments.
+"""
+
+import numpy as np
+
+from qiskit import pulse
+from qiskit.circuit import Parameter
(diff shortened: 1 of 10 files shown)


DanPuzzuoli:
An additional thing to note: I don't know if `black` got updated while making this PR, but when running the formatter it ended up reformatting a ton of additional files. Changes outside what I described above are all formatting changes.

test/dynamics/backend/test_qiskit_experiments.py
```diff
@@ -0,0 +1,100 @@
...
+        """Build simple simulator for multiple tests."""
+
+        # build the solver
+        solver = Solver(
+            static_hamiltonian=2 * np.pi * 5.0 * np.array([[-1.0, 0.0], [0.0, 1.0]]) / 2,
+            hamiltonian_operators=[2 * np.pi * np.array([[0.0, 1.0], [1.0, 0.0]]) / 2],
+            rotating_frame=2 * np.pi * 5.0 * np.array([[-1.0, 0.0], [0.0, 1.0]]) / 2,
+            hamiltonian_channels=["d0"],
+            channel_carrier_freqs={"d0": 5.0},
+            dt=1.0 / 4.5,
```
to24toro, line 47: [Q] What is this origin of "4.5" ?

test/dynamics/backend/test_qiskit_experiments.py
```diff
@@ -0,0 +1,100 @@
...
+        sigma = Parameter("sigma")
+        beta = Parameter("beta")
+        drive = pulse.DriveChannel(Parameter("ch0"))
+
+        # Define and add template schedules.
+        with pulse.build(name="x") as x:
+            pulse.play(pulse.Drag(dur, amp, sigma, beta), drive)
+
+        with pulse.build(name="sx") as sx:
+            pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("beta")), drive)
```
to24toro, line 79: [Q] Why do you define variables "amp" and "beta" again ? Maybe I don't know how to use parameter.

DanPuzzuoli: Good catch, I actually hadn't thought about this as I had just copied this code over from the qiskit-experiments [single qubit calibration tutorial](https://qiskit.org/documentation/experiments/tutorials/calibrating_real_device.html).

I just asked the dev team for experiments, and it was explained to me that it's necessary to define a new parameter so that the `X` and `SX` gate parameters are not linked. If you use the same `Parameter` instance in both, they will be linked, and it will mess with the tuning of each gate.

DanPuzzuoli: I've just changed the test to remove the lines
```
amp = Parameter('amp')
beta = Parameter('beta')
```
and changed the definition of `x` to directly instantiate those parameters when the constructor is called, similar to `sx`. It saves a couple lines of code.

test/dynamics/backend/test_qiskit_experiments.py
to24toro: Certainly, you're right.

to24toro (approved):

PR #179, opened 2023-02-02
Add support for python 3.11 and remove for 3.7

DanPuzzuoli:
### Summary

Adds Python 3.11 support, and removes Python 3.7 support.

### Details and comments

The removal of 3.7 corresponds with the removal of 3.7 support from JAX, so it's maybe only a matter of time before 3.7 use of Dynamics + JAX becomes a problem.


DanPuzzuoli:
Some required 3.7 test jobs won't leave the above list of tests for some reason. Going to remake the PR.

PR #180, opened 2023-02-02
 Add support for python 3.11 and remove for 3.7

### LINKED ISSUE #169: Add 3.11 support
We may need to drop 3.7 support as JAX has dropped it.

DanPuzzuoli:
### Summary

Closes #169.

Adds Python 3.11 support, and removes Python 3.7 support. This is a remake of #179 as some required 3.7 tests wouldn't leave the job list for some reason.

### Details and comments

The removal of 3.7 corresponds with the removal of 3.7 support from JAX, so it's maybe only a matter of time before 3.7 use of Dynamics + JAX becomes a problem.


DanPuzzuoli:
@mtreinish There are some lingering 3.7 tests that don't appear to be running - I remade the PR because I thought maybe it was the PR that had gotten into some weird state, but this one is also having that issue.

mtreinish:
Those jobs are listed as required in the branch protection rules. To remove them you'll have to go into the repo's branch protection settings and remove those jobs as being required to pass before merging (and likely add the 3.11 jobs too).

releasenotes/notes/python-311-support-7f5e6e135fd1e791.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Adds Python 3.11 support, and removes Python 3.7 support. Starting with this release, Qiskit
```
mtreinish, line 4: I would split out the 3.7 removal into a separate `upgrade` note. Since this is a removal any existing 3.7 users will need to know about.

mtreinish (commented):
The only other place you might to update is https://github.com/Qiskit/qiskit-dynamics/blob/main/pyproject.toml#L6 for the black configuration. But it doesn't really matter, it just might change how black formats code (although I think in practice it doesn't actually change anything).

DanPuzzuoli:
> The only other place you might to update is https://github.com/Qiskit/qiskit-dynamics/blob/main/pyproject.toml#L6 for the black configuration. But it doesn't really matter, it just might change how black formats code (although I think in practice it doesn't actually change anything).

I have changed that one already!

releasenotes/notes/python-311-support-7f5e6e135fd1e791.yaml
DanPuzzuoli: ah right - thanks, good catch.

mtreinish (approved):

PR #181, opened 2023-02-03
Unify notation when using pulse classes in pulse test

to24toro:
### Summary
Unify notation in test_pulse_to_signals when using qiskit.pulse. I selected notation of `pulse.XX` not `XX`.


DanPuzzuoli (approved):
Makes sense - good for consistency.

PR #186, opened 2023-02-13
Fix t_span construction in DynamicsBackend

### LINKED ISSUE #185: Case in which DynamicsBackend.run hangs
### Informations

- **Qiskit Dynamics version**: `main` branch
- **Python version**: 3.11
- **Operating system**: MacOS

This is a bizarre bug that I haven't been able to get to the bottom of. The code is below, but the setup is:
- A `Solver` is constructed with a 2 transmon model, and configured for pulse simulation.
- A `DynamicsBackend` is constructed with this solver instance.
- Some simple pulse schedules are created.

With the above setup:
- Calling `Solver.solve` on the pulse schedules works.
- Calling `DynamicsBackend.run` hangs - with some computation seeming to run indefinitely.

In the cases in which this happens, the weird thing about this is that `DynamicsBackend.run` seems to hang at the step that the ODE solver is called (when debugging and stepping over things it gets stuck here), however all `DynamicsBackend.run` does up to that point is create a `DynamicsJob` instance then call `Solver.solve`. I.e. for the purposes of this bug `DynamicsBackend.run` is basically a light wrapper around `Solver.solve`.

Setting up things up:
```
# Configure to use JAX internally
import jax
jax.config.update("jax_enable_x64", True)
jax.config.update('jax_platform_name', 'cpu')
Array.set_default_backend('jax')

import numpy as np

dim = 2

v0 = 5.1e9
anharm0 = -0.33e9
r0 = 0.1e9

v1 = 5.15e9
anharm1 = -0.33e9
r1 = 0.1e9

J = 0.002e9

a = np.diag(np.sqrt(np.arange(1, dim)), 1)
adag = np.diag(np.sqrt(np.arange(1, dim)), -1)
N = np.diag(np.arange(dim))

ident = np.eye(dim, dtype=complex)
full_ident = np.eye(dim**2, dtype=complex)

N0 = np.kron(ident, N)
N1 = np.kron(N, ident)

a0 = np.kron(ident, a)
a1 = np.kron(a, ident)

a0dag = np.kron(ident, adag)
a1dag = np.kron(adag, ident)


static_ham0 = 2 * np.pi * v0 * N0 + np.pi * anharm0 * N0 * (N0 - full_ident)
static_ham1 = 2 * np.pi * v1 * N1 + np.pi * anharm1 * N1 * (N1 - full_ident)

static_ham_full = static_ham0 + static_ham1 + 2 * np.pi * J * ((a0 + a0dag) @ (a1 + a1dag))

drive_op0 = 2 * np.pi * r0 * (a0 + a0dag)

from qiskit_dynamics import Solver

# build solver
dt = 1/4.5e9

solver = Solver(
    static_hamiltonian=static_ham_full,
    hamiltonian_operators=[drive_op0],
    rotating_frame=np.diag(static_ham_full),
    hamiltonian_channels=['d0'],
    channel_carrier_freqs={'d0': v0},
    dt=dt,
    #evaluation_mode="sparse"
)

from qiskit import pulse

schedules = []

gauss = pulse.library.Gaussian(
    128, 1., 256, name="Parametric Gauss"
)

with pulse.build() as schedule:
    with pulse.align_right():
        pulse.play(gauss, pulse.DriveChannel(0))
        pulse.acquire(duration=1, qubit_or_channel=0, register=pulse.MemorySlot(0))
```
Simulate the schedules using `Solver.solve` works:
```
from qiskit.quantum_info import Statevector

y0 = np.zeros(dim**2, dtype=complex)
y0[0] = 1.

solver.solve(
    t_span=np.array([0., schedule.duration*dt]), 
    y0=Statevector(y0), 
    signals=schedule,
    method='jax_odeint',
    atol=1e-6, rtol=1e-8
)
```
Building a `DynamicsBackend` with this `Solver` instance and calling `run` (with seemingly the same options) leads to a never-ending computation:
```
from qiskit_dynamics import DynamicsBackend

solver_options = {'method': 'jax_odeint', 'atol': 1e-6, 'rtol': 1e-8}

backend = DynamicsBackend(
    solver=solver,
    subsystem_dims=[dim, dim],
    solver_options=solver_options,
)

backend.run(schedule).result()
```

### Variations of the above code that work or don't

This was discovered when I was trying to see how a sparse simulation performed, hence the sparse configuration of the `Solver`. 

Additional tested situations:
- If I remove the `"sparse"` evaluation mode, and change the rotating frame to the full static Hamiltonian, it works. (I.e. standard dense solver configuration).
- However, a dense evaluation mode, still with the rotating frame being the diagonal of the static Hamiltonian, also fails.
- Keeping the sparse configuration, but setting `J = 0` (the coupling) also works. 
- Changing the units to GHz/ns instead of Hz/s works.

What's confusing is that the above would normally suggest some issue with the differential equation solving... however `Solver.solve` still _always_ works in all of the above configurations. Only `DynamicsBackend.run` hangs, _and_ it specifically hangs when the ODE solver is called!

I've also disabled jitting by changing the `Solver.solve` code to always skip the auto-jitting behaviour, and the issue persists, so it doesn't seem to be anything weird going on with the internal auto-jitting code.

### Further steps

I'll keep tryign to debug with jitting disabled to see where the difference might lie.

DanPuzzuoli:
### Summary

Closes #185 

See the above issue for a more detailed description, but this PR fixes a bug in which `DynamicsBackend` was using the wrong integration interval in `schedule` simulations: the times were passed as sample times, rather than as sample times `* dt` (the sample width). 

### Details and comments

The private function `dynamics_backend._get_acquire_data` has been updated to return a list of only acquire times for each schedule (as opposed to something with the shape of `t_span` that should be passed to `solve`). The documentation of this function has also been updated to clearly state this is what it is doing. After `_get_acquire_data` is called within `DynamicsBackend.run`, this is translated to the `t_span` argument by constructing the integration intervals with the correct samples times multiplied by `dt`.

At the moment, the PR does not directly test this change, as it occurs within inaccessible code. One option is to modify the `_get_acquire_data` function to return the constructed `t_span` value, and then we could add tests for that function, however since this function works only with schedules, it feels more natural for it to work purely with sample times.

This PR does however add tests for diagonally-specified rotating frames for the `_get_dressed_state_decomposition` utility function. While this turned out to be unrelated to this issue, I noticed this case was being missed in these tests.

### DIFF
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -595,7 +595,8 @@ def _get_acquire_instruction_timings(
             if acquire_time != schedule_acquire_times[0]:
                 raise QiskitError("DynamicsBackend.run only supports measurements at one time.")
 
-        acquire_time_list.append(schedule_acquire_times[0])
+        # use dt to convert acquire start time from samples to to the integration interval
+        t_span_list.append([0.0, dt*schedule_acquire_times[0]])
         measurement_subsystems = []
         memory_slot_indices = []
         for inst in schedule_acquires:
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -20,7 +20,7 @@ Pulse-enabled simulator backend.
 import datetime
 import uuid
 
-from typing import List, Optional, Union, Dict
+from typing import List, Optional, Union, Dict, Tuple
 import copy
 import numpy as np
 from scipy.integrate._ivp.ivp import OdeResult  # pylint: disable=unused-import
@@ -304,10 +304,15 @@ class DynamicsBackend(BackendV2):
 
         schedules, num_memory_slots_list = _to_schedule_list(run_input, backend=backend)
 
-        # get the acquires instructions and simulation times
-        t_span, measurement_subsystems_list, memory_slot_indices_list = _get_acquire_data(
-            schedules, backend.options.subsystem_labels
-        )
+        # get the acquires sample times and subsystem measurement information
+        (
+            acquire_time_list,
+            measurement_subsystems_list,
+            memory_slot_indices_list,
+        ) = _get_acquire_data(schedules, backend.options.subsystem_labels)
+
+        dt = self.options.solver._dt
+        t_span = [[0.0, x * dt] for x in acquire_time_list]
 
         # Build and submit job
         job_id = str(uuid.uuid4())
@@ -546,14 +551,27 @@ def _validate_run_input(run_input, accept_list=True):
         raise QiskitError(f"Input type {type(run_input)} not supported by DynamicsBackend.run.")
 
 
-def _get_acquire_data(schedules, valid_subsystem_labels):
+def _get_acquire_data(
+    schedules: List[Schedule], valid_subsystem_labels: List[int]
+) -> Tuple[List[int], List[List[int]], List[List[int]]]:
     """Get the required data from the acquire commands in each schedule.
 
-    Additionally validates that each schedule has acquire instructions occurring at one time,
-    at least one memory slot is being listed, and all measured subsystems exist in
-    subsystem_labels.
+    Additionally validates that each schedule has acquire instructions occurring at one time, at
+    least one memory slot is being listed, and all measured subsystems exist in
+    ``valid_subsystem_labels``.
+
+    Args:
+        schedules: A list of ``Schedule`` instances.
+        valid_subsystem_labels: Valid acquire channel indices.
+    Returns:
+        A Tuple of Lists containing, for each schedule: the sample time at which the acquire
+        instruction occurs, a list of the subsystems being measured, and a list of the memory slots
+        indices in which to store the results of each subsystem measurement.
+    Raises:
+        QiskitError: If a schedule contains no measurement, if a schedule contains measurements at
+            different times, or if a measurement has an invalid subsystem label.
     """
-    t_span_list = []
+    acquire_time_list = []
     measurement_subsystems_list = []
     memory_slot_indices_list = []
     for schedule in schedules:
@@ -576,7 +594,7 @@ def _get_acquire_data(schedules, valid_subsystem_labels):
             if acquire_time != schedule_acquire_times[0]:
                 raise QiskitError("DynamicsBackend.run only supports measurements at one time.")
 
-        t_span_list.append([0, schedule_acquire_times[0]])
+        acquire_time_list.append(schedule_acquire_times[0])
         measurement_subsystems = []
         memory_slot_indices = []
         for inst in schedule_acquires:
@@ -593,7 +611,7 @@ def _get_acquire_data(schedules, valid_subsystem_labels):
         measurement_subsystems_list.append(measurement_subsystems)
         memory_slot_indices_list.append(memory_slot_indices)
 
-    return t_span_list, measurement_subsystems_list, memory_slot_indices_list
+    return acquire_time_list, measurement_subsystems_list, memory_slot_indices_list
 
 
 def _to_schedule_list(
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -613,10 +613,7 @@ def _get_acquire_instruction_timings(
         measurement_subsystems_list.append(measurement_subsystems)
         memory_slot_indices_list.append(memory_slot_indices)
 
-    # convert acquire start times specified as samples to the integration interval
-    t_span = [[0.0, x * dt] for x in acquire_time_list]
-
-    return t_span, measurement_subsystems_list, memory_slot_indices_list
+    return t_span_list, measurement_subsystems_list, memory_slot_indices_list
 
 
 def _to_schedule_list(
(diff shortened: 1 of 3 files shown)


qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -304,10 +304,14 @@ def run(
...
-            schedules, backend.options.subsystem_labels
-        )
+        # get the acquires sample times and subsystem measurement information
+        (
+            acquire_time_list,
+            measurement_subsystems_list,
+            memory_slot_indices_list,
+        ) = _get_acquire_data(schedules, backend.options.subsystem_labels)
+
+        t_span = [[0.0, x * self.options.solver._dt] for x in acquire_time_list]
```
ihincks, line 314: non-blocking: consider the rename `_get_acquire_instruction_timings`, and consider passing `self.options.solver._dt` so that the output of the helper is more generic.

ihincks (dismissed):

qiskit_dynamics/backend/dynamics_backend.py
DanPuzzuoli: I've implemented both suggestions and added a test directly verifying that correct `t_span` values are being returned by the newly named `_get_acquire_instruction_timings`.

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -546,14 +550,28 @@ def _validate_run_input(run_input, accept_list=True):
...
-    Additionally validates that each schedule has acquire instructions occurring at one time,
-    at least one memory slot is being listed, and all measured subsystems exist in
-    subsystem_labels.
+    Additionally validates that each schedule has acquire instructions occurring at one time, at
+    least one memory slot is being listed, and all measured subsystems exist in
+    ``valid_subsystem_labels``.
+
+    Args:
+        schedules: A list of ``Schedule`` instances. valid_subsystem_labels: Valid acquire channel
+        indices. dt: The sample size.
```
ihincks, line 564:
```suggestion
        schedules: A list of schedules. 
        valid_subsystem_labels: Valid acquire channel indices. 
        dt: The sample size.
```

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -576,7 +594,7 @@ def _get_acquire_data(schedules, valid_subsystem_labels):
             if acquire_time != schedule_acquire_times[0]:
                 raise QiskitError("DynamicsBackend.run only supports measurements at one time.")
 
-        t_span_list.append([0, schedule_acquire_times[0]])
+        acquire_time_list.append(schedule_acquire_times[0])
```
ihincks, line 597:
```suggestion
        # use dt to convert acquire start time from samples to to the integration interval
        t_span_list.append([0.0, dt*schedule_acquire_times[0]])
```

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -546,14 +550,28 @@ def _validate_run_input(run_input, accept_list=True):
...
+        A Tuple of Lists containing, for each schedule: the list of integration intervals required
+        for each schedule (in absolute time, from 0.0 to the beginning of the acquire instructions),
+        a list of the subsystems being measured, and a list of the memory slots indices in which to
+        store the results of each subsystem measurement.
+    Raises:
+        QiskitError: If a schedule contains no measurement, if a schedule contains measurements at
+            different times, or if a measurement has an invalid subsystem label.
     """
-    t_span_list = []
+    acquire_time_list = []
```
ihincks, line 574:
```suggestion
    t_span_list = []
```

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -593,7 +611,10 @@ def _get_acquire_data(schedules, valid_subsystem_labels):
         measurement_subsystems_list.append(measurement_subsystems)
         memory_slot_indices_list.append(memory_slot_indices)
 
-    return t_span_list, measurement_subsystems_list, memory_slot_indices_list
+    # convert acquire start times specified as samples to the integration interval
+    t_span = [[0.0, x * dt] for x in acquire_time_list]
+
+    return t_span, measurement_subsystems_list, memory_slot_indices_list
```
ihincks, line 617:
```suggestion
    return t_span_list, measurement_subsystems_list, memory_slot_indices_list
```

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -546,14 +550,28 @@ def _validate_run_input(run_input, accept_list=True):
...
-    subsystem_labels.
+    Additionally validates that each schedule has acquire instructions occurring at one time, at
+    least one memory slot is being listed, and all measured subsystems exist in
+    ``valid_subsystem_labels``.
+
+    Args:
+        schedules: A list of ``Schedule`` instances. valid_subsystem_labels: Valid acquire channel
+        indices. dt: The sample size.
+    Returns:
+        A Tuple of Lists containing, for each schedule: the list of integration intervals required
```
ihincks, line 566:
```suggestion
        A tuple of lists containing, for each schedule: the list of integration intervals required
```

test/dynamics/backend/test_dynamics_backend.py
```diff
@@ -514,3 +515,34 @@ def test_simple_example(self):
         )
 
         self.assertDictEqual(output.data.counts, {"000": 513, "010": 511})
+
+
+class Test_get_acquire_instruction_timings(QiskitDynamicsTestCase):
```
ihincks, line 520: 👌

qiskit_dynamics/backend/dynamics_backend.py
DanPuzzuoli: oops, accidentally missed this one, but fixed it in the subsequent commit with linting.

ihincks (approved):

PR #193, opened 2023-02-17
DynamicsBackend tutorial

### LINKED ISSUE #192: Add tutorial for DynamicsBackend

DanPuzzuoli:
### Summary

Closes #192

This tutorial is written to show users how they can use the `DynamicsBackend` to specify pulse-level simulations in various ways:
- Running simulations directly specified as pulse schedules
- Running simulations specified as circuits, but where each element either has an attached calibration, or a corresponding pulse definition in the target
- Simulating processes in calibration and characterization experiments generated by qiskit-experiments

Each of the above requires different levels of configuration of the various attributes of `DynamicsBackend`, so this tutorial will guide users on the setup necessary to achieve their desired task.

### Details and comments

It is currently just a jupyter notebook in the `docs/tutorial` folder.


DanPuzzuoli:
I've just converted to RST. I've added pandas to `requirements-dev.txt`, but it's only used for table visualization, so it may make sense to remove this.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+definitions, and calibration and characterization experiments from Qiskit Experiments.
+
+The sections of this tutorial are as follows: 
+
+1. Configure Dynamics to use JAX.
+2. Instantiating a minimally-configured :class:`.DynamicsBackend` with a 2 qubit model.
+3. Simulating pulse schedules on the :class:`.DynamicsBackend`.
+4. Simulating circuits at the pulse level using the :class:`.DynamicsBackend`.
+5. Simulating single-qubit calibration processes via Qiskit Experiments.
+6. Simulating 2 qubit interaction characterization via the ``CrossResonanceHamiltonian`` experiment.
```
to24toro, line 16: The text of the item does not match the item which is responding to after. Is it intentional?

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+
+.. jupyter-execute::
+
+    result.get_counts(0)
+
+5. Simulating single-qubit calibration processes via Qiskit Experiments
+-----------------------------------------------------------------------
+
+Next, we calibrate ``X`` and ``SX`` gates on both qubits modeled in the :class:`.DynamicsBackend`
+using Qiskit Experiments, following the single-qubit calibration tutorial.
```
coruscating, line 268: It'd be helpful to link to the experiments tutorial here (though the URL will be changing soon so you can add this later).

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+    dur = Parameter("dur")
+    sigma = Parameter("σ")
+    drive = pulse.DriveChannel(Parameter("ch0"))
+    
+    # Define and add template schedules.
+    with pulse.build(name="x") as x:
+        pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("β")), drive)
+    
+    with pulse.build(name="sx") as sx:
+        #pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("β")), drive)
```
coruscating, line 329:
```suggestion
```

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+    data_fine_sx1 = amp_sx_cal1.run().block_for_results()
+
+.. jupyter-execute::
+
+    pd.DataFrame(**cals.parameters_table(qubit_list=[0, ()], parameters="amp"))
+
+6. Simulating 2 qubit interaction characterization via the ``CrossResonanceHamiltonian`` experiment
+---------------------------------------------------------------------------------------------------
+
+Finally, simulate the ``CrossResonanceHamiltonian`` characterization experiment.
```
coruscating, line 521: You can link directly to the experiments API docs for this. Also, maybe add a brief explanation for why the backend needs more configuration to run this experiment.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+
+    drag0_data.figure(0)
+
+
+.. jupyter-execute::
+
+    drag1_data.figure(0)
+
+
+5.6 Fine amplitude calibration
```
coruscating, line 465: These calibrations are all really cool to see, but you don't necessarily have to go through all of them since they're already in the calibrations tutorial and you can just refer the user to them. But for the ones that you keep, it would be helpful to link to the experiments's API docs.

coruscating (commented):
One thing that would be helpful for people coming from experiments is a comparison between `DynamicsBackend`, Aer simulators, and `PulseBackend` in Experiments (which uses Dynamics) and when to use each simulator. Not sure if this is in the scope for this tutorial though.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: Hmm I guess this more a general style discussion then, as we have these tables of contents in every tutorial/userguide in dynamics. I guess it is duplicated in the sidebar, but my preference would be to keep it so it can be read through as part of the intro.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+3. Simulating pulse schedules on the :class:`.DynamicsBackend`.
+4. Simulating circuits at the pulse level using the :class:`.DynamicsBackend`.
+5. Simulating single-qubit calibration processes via Qiskit Experiments.
+6. Simulating 2 qubit interaction characterization via the ``CrossResonanceHamiltonian`` experiment.
+
+1. Configure Dynamics to use JAX
+--------------------------------
+
+Note that the :class:`.DynamicsBackend` internally performs just-in-time compilation automatically
+when configured to use JAX.
```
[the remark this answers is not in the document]
DanPuzzuoli: Add a reference to this.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+    
+    a0 = np.kron(ident, a)
+    a1 = np.kron(a, ident)
+    
+    a0dag = np.kron(ident, adag)
+    a1dag = np.kron(adag, ident)
+    
+    
+    static_ham0 = 2 * np.pi * v0 * N0 + np.pi * anharm0 * N0 * (N0 - full_ident)
+    static_ham1 = 2 * np.pi * v1 * N1 + np.pi * anharm1 * N1 * (N1 - full_ident)
```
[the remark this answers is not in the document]
DanPuzzuoli: good catch - fixed the equation.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+differential equation.
+
+Note that, to enable the internal automatic jit-compilation, we choose a JAX integration method.
+
+.. jupyter-execute::
+
+    from qiskit_dynamics import DynamicsBackend
+    
+    # Consistent solver option to use throughout notebook
+    solver_options = {"method": "jax_odeint", "atol": 1e-6, "rtol": 1e-8}
```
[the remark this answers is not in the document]
DanPuzzuoli: Added a sentence stating that the full list of allowable `solver_options` are the arguments to `solve_ode` (which is a clickable link).

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+        
+    job = backend.run(schedules, shots=100)
+    result = job.result()
+
+Retrieve the counts for one of the experiments as would be done using the results object from a real
+backend.
+
+.. jupyter-execute::
+
+    result.get_counts(3)
```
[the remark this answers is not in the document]
DanPuzzuoli: Yeah I agree, while it's not that meaningful it's still nice to see, especially since part of the point is to highlight that all the usual instructions work.

I've added a cell drawing the schedule with index `3`.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+    
+    circ = QuantumCircuit(1, 1)
+    circ.h(0)
+    circ.measure([0], [0])
+    
+    circ.draw("mpl")
+
+Next, attach a calibration for the Hadamard gate on qubit :math:`0` to the circuit. Note that here
+are only demonstrating the mechanics of adding a calibration; we have not actually chosen the pulse
+to implement a Hadamard gate.
```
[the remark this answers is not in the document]
DanPuzzuoli: Yeah this is confusing - by attaching the schedule to the gate I'm implying that it's a pulse-level "implementation" of the gate, but when I said it is not "chosen to implement a Hadamard gate" I meant that it's not a _good_ implementation (in the sense that I don't even know what the unitary generated is).

I've modified the sentence to say that the schedule hasn't been calibrated to implement the Hadamard gate with high fidelity.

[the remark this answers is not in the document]
coruscating: Ohh it seems so obvious after you explained it 😆 But I think the rewording is super clear, thanks.

docs/tutorials/dynamics_backend.rst
coruscating: That's fair, it's definitely nice to have the outline laid out at the start. Maybe if the sidebar stayed to the side as the user scrolled down the page it can serve a distinct purpose from the table of contents. Anyways, that's beyond the scope of this PR.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,566 @@
...
+
+Alternatively to the above work flow, add the above schedule as the pulse-level definition of the
+Hadamard gate on qubit :math:`0`.
+
+.. jupyter-execute::
+
+    from qiskit.circuit.library import HGate
+    from qiskit.transpiler import InstructionProperties
+    
+    backend.target.add_instruction(HGate(), {(0,): InstructionProperties(calibration=h_q0)})
```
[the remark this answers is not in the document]
DanPuzzuoli: I've run into this duplicate `add_instruction` problem as well, but haven't tried to figure it out. Maybe you can retrieve the instruction instance you want to modify from the `target` and directly modify that.

In any case it's a good idea to give a bit more information here. I've expanded the preceding sentence slightly to indicate that the point of the target is to configure transpilation, and added an additional sentence linking to the `Target` class documentation.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: It would definitely be nice to have a link here, but I guess if it's changing there's nothing to be done at this point. Hopefully the sentence is descriptive enough.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: Just added `qiskit_experiments` to intersphinx (hopefully correctly) and turned `CrossResonanceHamiltonian` into a link. I've also expanded the bullet points below to give more context as to why the experiment requires this further configuration:
- Adding the gate to the target is required for validation
- The control_channel_map is required for the experiment to know which control channel to drive for a given control-target cross resonance pair

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: You're right. I've deleted the frequency calibrations as well as the fine calibrations. I've kept the rough amp and rough drag cals. I could maybe even drop the rough drag cals - I think it should be enough for the `CrossResonanceHamiltonian` experiment to have amplitudes roughly calibrated. That being said I do like to see both amplitude and drag.

DanPuzzuoli:
@coruscating In addition to the above replies, with the addition of cross-linking of `qiskit_experiments`, I've tried to get all class references to Terra and Experiments to be proper links (hopefully not missing any).

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+solver>` for more details.
+
+.. jupyter-execute::
+
+    from qiskit_dynamics import Solver
+    
+    # build solver
+    dt = 1/4.5e9
+    
+    solver = Solver(
```
nkanazawa1989, line 118: To me it looks bit weird (and confusing) it is called _solver_. These constructor arguments make me think it is a _model_ rather than a solver, and solver is something that consumes the model with solver options (run options in terms of qiskit) -- but I know the solver functionality is indeed integrated into this class. This is an issue of class design and not the scope of this PR -- but maybe you can update the dynamics backend (e.g. `DynamicBackend.from_hamiltonian`) so that it can take these parameters and internally instantiate the solver instance to hide implementation details.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+.. jupyter-execute::
+
+    from qiskit_dynamics import DynamicsBackend
+    
+    # Consistent solver option to use throughout notebook
+    solver_options = {"method": "jax_odeint", "atol": 1e-6, "rtol": 1e-8}
+    
+    backend = DynamicsBackend(
+        solver=solver,
+        subsystem_dims=[dim, dim], # for computing measurement data
```
nkanazawa1989, line 144: What happens if you set [2, 2] here? f-state is measured as 1?

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+    
+    schedules = []
+    
+    for amp in np.linspace(0., 1., 10):
+        gauss = pulse.library.Gaussian(
+            num_samples, amp, sigma, name="Parametric Gauss"
+        )
+    
+        with pulse.build() as schedule:
+            with pulse.align_right():
```
nkanazawa1989, line 176: this alignment is bit problematic because drive channel and measure channel are both ALAP scheduled in parallel. `align_sequential` should work in this example (because there is no parallel pulse).

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+                pulse.play(gauss, pulse.DriveChannel(0))
+                pulse.shift_phase(0.5, pulse.DriveChannel(0))
+                pulse.shift_frequency(0.1, pulse.DriveChannel(0))
+                pulse.play(gauss, pulse.DriveChannel(0))
+                pulse.acquire(duration=1, qubit_or_channel=0, register=pulse.MemorySlot(0))
```
nkanazawa1989, line 181: Reader may wonder how acquire with 1 dt work without any readout stimulus.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+
+.. jupyter-execute::
+
+    with pulse.build() as h_q0:
+        pulse.play(
+            pulse.library.Gaussian(duration=256, amp=0.2, sigma=50, name="custom"),
+            pulse.DriveChannel(0)
+        )
+    
+    circ.add_calibration("h", [0], h_q0)
```
nkanazawa1989, line 237:
```suggestion
    circ.add_calibration("h", qubits=[0], schedule=h_q0)
```
assumed user is not familiar with the add_calibration API.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+Rebuild the same circuit, however this time we do not need to add the calibration for the Hadamard
+gate to the circuit object.
+
+.. jupyter-execute::
+
+    circ2 = QuantumCircuit(1, 1)
+    circ2.h(0)
+    circ2.measure([0], [0])
+    
+    %time result = backend.run(circ2).result()
```
nkanazawa1989, line 270: Does it work with 2+ qubits? I think it internally calls pulse scheduler but there is a [known bug](https://github.com/Qiskit/qiskit-terra/issues/9488).

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+.. jupyter-execute::
+
+    from qiskit.circuit.library import XGate, SXGate, RZGate
+    from qiskit.circuit import Parameter
+    from qiskit.providers.backend import QubitProperties
+    
+    target = backend.target
+    
+    # qubit properties
+    target.qubit_properties = [QubitProperties(frequency=v0), QubitProperties(frequency=v1)]
```
nkanazawa1989, line 302: These are undressed and inaccurate. Do you want to mention that? Also this description makes me think the frequency value set here will update experiment condition which actually doesn't (unless you update `channel_carrier_freqs` in the solver constructor). I think this should be mentioned otherwise user will easily misunderstand the mechanism.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+    
+    # add instructions
+    target.add_instruction(XGate())
+    target.add_instruction(SXGate())
+    
+    # Add RZ instruction as phase shift for drag cal
+    phi = Parameter("phi")
+    with pulse.build() as rz0:
+        pulse.shift_phase(phi, pulse.DriveChannel(0))
```
nkanazawa1989, line 311: Since this is a two qubit system, you also need to update control channel phase to run CR experiment, i.e. `ControlChannel(1)` for rz0.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+    
+    cals.add_schedule(x, num_qubits=1)
+    cals.add_schedule(sx, num_qubits=1)
+    
+    # add parameter guesses
+    for sched in ["x", "sx"]:
+        cals.add_parameter_value(80, "σ", schedule=sched)
+        cals.add_parameter_value(0.5, "β", schedule=sched)
+        cals.add_parameter_value(320, "dur", schedule=sched)
+        cals.add_parameter_value(0.5, "amp", schedule=sched)
```
nkanazawa1989, line 353:
```suggestion
    cals = Calibrations()
    
    dur = Parameter("dur")
    sigma = Parameter("σ")
    drive = pulse.DriveChannel(Parameter("ch0"))
    
    # Define and add template schedules.
    with pulse.build(name="x") as x:
        pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("β")), drive)
    
    with pulse.build(name="sx") as sx:
        pulse.play(pulse.Drag(dur, Parameter("amp"), sigma, Parameter("β")), drive)
    
    cals.add_schedule(x, num_qubits=1)
    cals.add_schedule(sx, num_qubits=1)
    
    # add parameter guesses
    for sched in ["x", "sx"]:
        cals.add_parameter_value(80, "σ", schedule=sched)
        cals.add_parameter_value(0.5, "β", schedule=sched)
        cals.add_parameter_value(320, "dur", schedule=sched)
        cals.add_parameter_value(0.5, "amp", schedule=sched)
```
You can skip these lines by instantiating the calibration class with [a template library](https://github.com/Qiskit/qiskit-experiments/blob/0a9da6a2656515b20fcc6338207b41d1ee4274ee/qiskit_experiments/calibration_management/basis_gate_library.py#L180). Code here is correct anyways.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+- Defining the control channel map, which is a dictionary mapping control-target qubit index pairs
+  (given as a tuple) to the control channel index used to drive the corresponding cross-resonance
+  interaction. This is required by the experiment to determine which channel to drive for each
+  control-target pair.
+
+.. jupyter-execute::
+
+    # add the gate to the target
+    from qiskit_experiments.library import CrossResonanceHamiltonian
+    backend.target.add_instruction(
```
nkanazawa1989, line 455: This is really needed? I think the experiment [dynamically adds calibration](https://github.com/Qiskit/qiskit-experiments/blob/0a9da6a2656515b20fcc6338207b41d1ee4274ee/qiskit_experiments/library/characterization/cr_hamiltonian.py#L395) and transpiler and scheduler can schedule it without having defs in the Target. If it doesn't probably it's a bug.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+
+.. jupyter-execute::
+
+    cr_ham_experiment = CrossResonanceHamiltonian(
+        qubits=(0, 1), 
+        flat_top_widths=np.linspace(0, 5000, 17), 
+        backend=backend
+    )
+    
+    cr_ham_experiment.set_transpile_options(inst_map=cals.default_inst_map)
```
nkanazawa1989, line 474: Since the dynamics backend is a V2 wrapper, this will run into the [known bug](https://github.com/Qiskit/qiskit-terra/issues/9489). You can export the instmap from calibration and use [this method](https://github.com/Qiskit/qiskit-terra/blob/126c9ba1e3b1dc1486b0191de759bd756ff91a28/qiskit/transpiler/target.py#L416) to mutate backend target.

docs/tutorials/dynamics_backend.rst
```diff
@@ -0,0 +1,489 @@
...
+    jax.config.update("jax_enable_x64", True)
+    jax.config.update("jax_platform_name", "cpu")
+    from qiskit_dynamics.array import Array
+    Array.set_default_backend("jax")
+
+2. Instantiating a minimally-configured :class:`.DynamicsBackend` with a 2 qubit model
+--------------------------------------------------------------------------------------
+
+To create the :class:`.DynamicsBackend`, first specify a :class:`.Solver` instance using the model
+details. For the model we will use a 2 qubit transmon model, with Hamiltonian:
```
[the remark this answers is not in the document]
DanPuzzuoli: I've expanded this paragraph to state:

1. The model depends on the device you wish to simulate
2. The model we choose here is for a pair of fixed frequency transmons with a fixed coupling
3. I've also added a link to the Qiskit textbook page as a resource for information on how these Hamiltonians are derived

There is no RWA being performed with this model, at least not in the sense of "RWA" that I'm familiar with (entering the rotating frame, then doing an approximation, starting from this model). 

Based on this comment though, I've added commentary on the choice of rotating frame when building the `Solver` (note that no approximation is being made here, it's just a choice for how the simulation is done). I've tried to make it clear that this choice only impacts performance of the solver, and does not impact the measurement outcomes of simulations.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: I agree that the `Solver` terminology is a bit weird. The core design of dynamics (from the first release) is model classes, `HamiltonianModel`, `LindbladModel`, ..., which you then pass to the core solver functions `solve_ode` and `solve_lmde`. This ended up being slightly awkward to use however, so we made the `Solver` class as a high level wrapper of these things (it builds the right kind of model internally, then calls the underlying solver functions with it). I think in the future, as we refactor/refine the package, this awkwardness should be addressed. I have some ideas for how this might look, but yeah as you say it is well outside the scope of this PR :).

For the `DynamicsBackend.from_hamiltonian` comment: we did discuss this earlier in the `DynamicsBackend` development process. The conclusion was that, if we made the constructor take in all of the model arguments directly, it would be very redundant. I.e. it ends up looking exactly like the constructor for `Solver` (which can take a lot of arguments), plus only a few other arguments. I think this had the effect of obscuring what was different between the `Solver` and `DynamicsBackend` constructors, so we ended up just deciding to have the main constructor for `DynamicsBackend` consume a `Solver` for clarity, and to avoid replicating the `Solver.__init__` doc string. I realize you're suggesting something slightly different: to add another static method that does this, but I think the reasoning still applies.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: No, an error will be raised in that case, as there is a validation check ensuring that `np.prod(subsystem_dims)` is equal to `solver.model.dim`. Customization of how measurement outcomes are computed can be done with other `DynamicsBackend` options, but `subsystem_dims` is purely meant to be an exact specification of how the model in `solver` breaks into subsystems.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: Changed it to `align_sequential`.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: I've just added a note above here talking about this. It provides some explanation of how measurement is performed in `DynamicsBackend`, explaining why the `duration=1` was chosen, and why no `MeasureChannel` was used.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: Ah thank you, good catch, I've added this.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: If I don't add this an error is raised.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: Thank you - I've dropped this cell and have replaced it with one using the fixed frequency transmon template library. It changes the plots a bit due to the different default durations and sigma, but it works all the way through.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: What is the correct way of doing this? I've tried:
```
backend.target.update_from_instruction_schedule_map(cals.get_inst_map())
```

but I get the error:
```
KeyError: "Provided qarg: '(0,)' not in this Target for x"
```

docs/tutorials/dynamics_backend.rst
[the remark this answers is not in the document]
nkanazawa1989: Thanks Dan. This sounds reasonable. I'm curious when the resonator subspace is decoupled in this popular setting. I think we don't need to consider that subspace unless we populate the resonator, but I was thinking we need RWA to make displacement transform. What am I missing in this context?

docs/tutorials/dynamics_backend.rst
nkanazawa1989: You also need to properly configure the meas_map (backend.configuration().meas_map) to schedule, i.e. if you specify `[[0,1]]` in the meas_map you need to provide a schedule to measure q0 and q1 simultaneously (this is a typical setting in IBM backend).

docs/tutorials/dynamics_backend.rst
nkanazawa1989: hmm `Target` is not well supported by `Calibrations` at this moment. Indeed your calibration instance have several gates that is not predefined in your target. Probably this trick works

https://github.com/Qiskit/qiskit-experiments/blob/main/qiskit_experiments/calibration_management/basis_gate_library.py#L211

namely you can instantiate the template library with the limited set of basis gates (`basis_gates=["sx"]` in your example) that is defined by your target instance.

docs/tutorials/dynamics_backend.rst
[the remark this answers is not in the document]
DanPuzzuoli: I don't know - I usually just have physicists tell me this is the Hamiltonian :). It sounds reasonable to me that some RWA is being done to get to this point, but I'm not familiar with this analysis.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: There are more computations being done by `DynamicsBackend` (the measurement portion), but yeah `Solver` is the main thing.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: I didn't want to make this limitation. A user can configure that if they would like to.

docs/tutorials/dynamics_backend.rst
DanPuzzuoli: So if I add a dummy instruction, e.g. ``backend.target.add_instruction(ZGate())``, the `'coupling graph not connected'` error goes away, and the code runs, including the line:
```
backend.target.update_from_instruction_schedule_map(cals.get_inst_map())
```
I've done some cursory checks of `backend.target.instructions` and `backend.target.instruction_properties`, and it seems like the calibrations have been correctly updated.

So, that's all good, but I need to figure out how to get rid of the `CouplingError: 'coupling graph not connected'` error without the hack of adding some gate that applies to all qubits (like the `ZGate`). Any ideas?

DanPuzzuoli: After discussing with @wshanks:
- There appears to be no way around this.
- As the lesser hack, I've also added `CXGate` as an instruction in the target configuration cell, along with a note explaining that this is being done so that that device registers as "connected" in validation checks.

wshanks: Well, I didn't know a way around without more investigation at least. My summary of what is happening is:

* qiskit-experiments always runs `transpile` when generating circuits.
* Some pass in the default pass manager wants to look at the full device and expects that the qubits in the device to all be part of one connected graph.
* With BackendV2/Target, the only way to specify the connectivity is by defining two qubit gates between pairs of qubits (you can't specify connectivity directly outside of gates).
* Defining a global gate (without specific qubit indices) causes this connectivity check to get bypassed (this lets just adding `ZGate` or `CXGate` without qubit indices work).

The solutions that I can think of besides adding a global gate are:

* Add `CXGate` (or other two qubit gate) with a specific qubit pair with a comment that this is how you define the connectivity of the qubits in a V2 backend. Dan tried this but ran into another transpiler error about reversing the direction of a gate.
* Update qiskit-experiments to allow more flexibilty with the transpilation (custom pass manager). With this, dynamics could avoid whatever transpiler pass is causing the problem currently. This has been discussed but is a longer term solution.
* Figure out which transpiler pass is the problem and see if it can be bypassed by setting a transpile option on the experiment.

@nkanazawa1989 Do you have any thoughts on those?

Also, one follow up thought -- maybe DynamicsBackend could do something to make this easier for the user? Maybe DynamicsBackend could define some kind of placeholder gate between coupled qubits or otherwise enforce the coupling map? Perhaps this is overkill if the only goal is to avoid an error with a transpiler pass that expects a coupling map and there are no other reasons to need an accurate coupling map for the DynamicsBackend.

docs/tutorials/dynamics_backend.rst
nkanazawa1989: Yes. I'm fine with current code as long as it works.

docs/tutorials/dynamics_backend.rst
nkanazawa1989: Thanks @wshanks for investigation. Yes, your summary looks correct. Target generates the coupling map on the fly based on stored instructions (one provided by the backend is just ignored). I think this behavior is weird, because the coupling edge doesn't appear unless user adds calibration even though two nodes are physically connected.

Probably you could do small hack for `DynamicsBackend`. Namely, you can override [coupling_map](https://github.com/Qiskit/qiskit-terra/blob/fa7ac688fdce3ecc27fb7e13d8b2c3f8553b3ff8/qiskit/providers/backend.py#L387-L390) not to call `Target.build_coupling_map` method. You can generate the coupling map based on device Hamiltonian (usually parsing Hamiltonian is really tough). Alternatively you can simply assume full connectivity and returns `CouplingMap(itertools.product(range(n_qubit), range(n_qubit)))`.

DanPuzzuoli: Okay - I've just added a `coupling_map` option to the backend, and implemented the `coupling_map` property, in the following way:
- If the option is set to `"complete"`, calling `backend.coupling_map` will return a `CouplingMap` representing the complete graph. I had to modify your suggested code to avoid having edges of the form `(0, 0)`.
- If the option is set to `"target"`, `backend.coupling_map` will call `target.build_coupling_map()` as before.
- If the option is set to a `CouplingMap` instance, `backend.coupling_map` will just return that.
I've also added error raising and tests to check for the behaviour of this option. The default is to return the complete graph.

Unfortunately, after implementing this, and removing the `CXGate` hack, I still got the error that the `cr_gate` cannot be flipped. So, this doesn't appear to actually solve the issue. In the section for the `CrossResonanceHamiltonian`, I've re-added a line adding the `CRPulseGate` to the `target`. 

Any other ideas @wshanks @nkanazawa1989 ? If not I'm fine with leaving things the way they are currently.

wshanks: Hmm, you overrode the Backend's coupling map, but I would guess that for the transpiler issues we have been hitting here it is the Target's coupling map gets used.

DanPuzzuoli: The default behaviour is for the coupling map to be complete - I thought this would resolve the problem.

Ah but I guess you're saying that whatever is looking up the coupling map is directly accessing the coupling map `Target`?

I think I might just revert these changes and add the `CRPulseGate` to the instructions as before. It would be nice to not have to do this, but this seems like a much cleaner hack than messing around too much with the `DynamicsBackend` class itself.

DanPuzzuoli: Alright I've reverted this - unless there are serious objections to this I think the hack of adding this gate is the preferred solution.

wshanks: I think using a hack here is the best option that I know of. Ideally, qiskit-experiments will allow more flexibility with transpilation in the future which could avoid the issue. The one thing I still wonder about is if there is set of transpiler options that could disable whatever is causing the error (since we know the circuit should run fine on DynamicsBackend without any transpilling, we just need the `transpile` call that qiskit-experiments does not to change anything).

Also, I think there might be a transpiler bug here. I will look into it a little more and report something to terra if it makes sense.

DanPuzzuoli: @wshanks @nkanazawa1989 - if this works for you, please approve. It is the last PR before for the next release :).

wshanks (dismissed):

coruscating (dismissed):
LGTM besides a typo.

coruscating (approved):

docs/tutorials/dynamics_backend.rst
wshanks: Terra issue about the transpiler problem we were seeing: https://github.com/Qiskit/qiskit-terra/issues/9783

PR #197, opened 2023-03-03
[WIP]Arraylias integration

to24toro:
### Summary
At first, I try to migrate to arraylias in signals.py.
- Import `numpy_array()` from arraylias
- Set `unp = numpy_array()()`
- Be able to use `unp` like `np`.

It results in simplifying codes in dynamcis due to arraylias.

In addition, I changed only some parts of `np` to `unp`: 
E.g. np.append(A, B) makes A numpy.ndarray even if the type of A is jax.numpy.array. ``np.exp(A) is the same.
~On the other hand, for example, we don't need to change `np.real(A)`, which reflects the type of A.~
In JAX.jit, it is necessary to use jnp.real not np.real.

I haven't decided if all the files are migrated in only  this PR.

### Details and comments
JAX array almost supports numpy.array methods. But it doesn't all of them.
So it may be better if we can test whether all the methods used in a library are implemented both in numpy and jax. 
Of course, whether to incorporate this way into arraylias will be another issue.

### DIFF
--- a/qiskit_dynamics/signals/signals.py
+++ b/qiskit_dynamics/signals/signals.py
@@ -629,10 +629,10 @@ class DiscreteSignalSum(DiscreteSignal, SignalSum):
     def __init__(
         self,
         dt: float,
-        samples: Union[List, Array],
+        samples: List,
         start_time: float = 0.0,
-        carrier_freq: Union[List, np.array, Array] = None,
-        phase: Union[List, np.array, Array] = None,
+        carrier_freq: Union[List, np.array, jnp.array] = None,
+        phase: Union[List, np.array, jnp.array] = None,
         name: str = None,
     ):
         r"""Directly initialize a ``DiscreteSignalSum``\. Samples of all terms in the
@@ -649,12 +649,12 @@ class DiscreteSignalSum(DiscreteSignal, SignalSum):
             name: name of the signal.
         """
 
-        samples = Array(samples)
+        samples = unp.asarray(samples)
         if carrier_freq is None:
-            carrier_freq = np.zeros(samples.shape[-1], dtype=float)
+            carrier_freq = unp.zeros(samples.shape[-1], dtype=float)
 
         if phase is None:
-            phase = np.zeros(samples.shape[-1], dtype=float)
+            phase = unp.zeros(samples.shape[-1], dtype=float)
 
         DiscreteSignal.__init__(
             self,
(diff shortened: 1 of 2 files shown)


qiskit_dynamics/signals/signals.py
```diff
@@ -124,43 +123,44 @@ def is_constant(self) -> bool:
...
 
     @phase.setter
-    def phase(self, phase: Union[float, list, Array]):
+    def phase(self, phase: Union[float, list]):
         """Phase setter. List handling is to support subclasses storing a
         list of phases."""
-        if type(phase) == list:
-            phase = [Array(entry).data for entry in phase]
-        self._phase = Array(phase)
+        self._phase = unp.asarray(phase)
```
to24toro, line 148: It is one approach of restricting to only numpy.array or jax.numpy.array in type of attributes in dynamics.

to24toro:
If integrating with Arraylias, is there a possibility that type of `idx` could be jnp.array?
https://github.com/Qiskit/qiskit-dynamics/blob/671c5836ea7a36ccea7fe0659cb19eb259f5628c/qiskit_dynamics/signals/signals.py#L494

qiskit_dynamics/signals/signals.py
```diff
@@ -591,12 +582,11 @@ def envelope(t):
             self, envelope=envelope, carrier_freq=carrier_freqs, phase=phases, name=name
         )
 
-    def complex_value(self, t: Union[float, np.array, Array]) -> Union[complex, np.array, Array]:
+    def complex_value(self, t: Union[float, np.array, jnp.array]) -> Union[complex, np.array, jnp.array]:
```
to24toro, line 585: We need to decide the type hint of Arraylias.

qiskit_dynamics/signals/signals.py
```diff
@@ -801,15 +799,15 @@ def __init__(self, signal_list: List[Signal]):
 
         super().__init__(signal_list)
 
-        self._eval_complex_value = lambda t: [sig.complex_value(t) for sig in self.components]
-        self._eval_signals = lambda t: [sig(t) for sig in self.components]
+        self._eval_complex_value = lambda t: unp.asarray([sig.complex_value(t) for sig in self.components])
+        self._eval_signals = lambda t: unp.asarray([sig(t) for sig in self.components])
 
-    def complex_value(self, t: Union[float, np.ndarray, jnp.ndarray]) -> Union[np.ndarray, jnp.ndarray]:
```
to24toro: An error occurs in the case that we cannot use jax if we use a type hint of something relating to jax.

qiskit_dynamics/signals/signals.py
```diff
@@ -299,39 +297,25 @@ def __init__(
...
+        def envelope(t):
+            t = unp.asarray(t)
+            idx = unp.clip(
+                    unp.array((t - self._start_time) // self._dt, dtype=int),
                     -1,
                     len(self.samples),
                 )
-                return self._padded_samples[idx]
-
+            return self._padded_samples[idx]
```
to24toro, line 318: We need to improve this function. Because of it, cannot pass tests.

DanPuzzuoli: What is the actual problem here? What is failing?

to24toro:
I close once. I will change the merge direction and reopen it after discussion in detail.

PR #203, opened 2023-03-13
Fix error in schedule -> signal conversion for schedules with barrier instructions

### LINKED ISSUE #202: Add support for barriers in schedule -> signal converter
The following example is courtesy of @zlianghahaha : 

```
from qiskit_dynamics.pulse import InstructionToSignals
from qiskit import pulse
from qiskit.pulse.transforms.canonicalization import block_to_schedule

from qiskit.providers.fake_provider import FakeQuito

backend = FakeQuito()

with pulse.build(backend) as sched_block:
    pulse.play(pulse.Constant(duration=10, amp=0.5), pulse.DriveChannel(0))
    pulse.barrier(0, 1)
    pulse.play(pulse.Constant(duration=10, amp=-0.5), pulse.DriveChannel(1))

    
converter = InstructionToSignals(dt=1., carriers={'d0': 1., 'd1': 1.})
sched = block_to_schedule(sched_block)

converter.get_signals(sched)
```

This raises the error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 18
     15 converter = InstructionToSignals(dt=1., carriers={'d0': 1., 'd1': 1.})
     16 sched = block_to_schedule(sched_block)
---> 18 converter.get_signals(sched)

File ~/Documents/projects/qiskit-dynamics/qiskit_dynamics/pulse/pulse_to_signals.py:166, in InstructionToSignals.get_signals(self, schedule)
    158     signals[chan.name] = DiscreteSignal(
    159         samples=[],
    160         dt=self._dt,
    161         name=chan.name,
    162         carrier_freq=carrier_freq,
    163     )
    165 for start_sample, inst in schedule.instructions:
--> 166     chan = inst.channel.name
    167     phi = phases[chan]
    168     freq = frequency_shifts[chan]

AttributeError: 'RelativeBarrier' object has no attribute 'channel'
```

The error is being raised as the `RelativeBarrier` object is in `sched`, and doesn't have a `channel` attribute. Interestingly, I'm fairly sure this can be fixed by just doing nothing for barrier instructions. `block_to_schedule` already puts absolute timing in all of the other instructions (which is what we use), so there is no need to interpret the contents of barrier instructions in the converter.

As the converter already takes action based on the type of instruction (and only handles the types of instructions it needs to know about), I think this can be fixed by just changing the single line `chan = inst.channel.name`, either by eliminating it and directly using `inst.channel.name` in the few blocks that follow, or doing `chan = inst.channel.name if hasattr(inst, "channel") else None`.

DanPuzzuoli:
### Summary

Closes #202.

This fixes an error in the schedule -> signal converter when the schedule contained a barrier instruction.

### Details and comments

As described in #202, in the loop over instructions to construct the signals, there is a line `chan = inst.channel.name`, which assumes that `inst.channel` exists. Barrier instructions do not have a `channel` attribute, hence the error. This PR modifies this line to conditionally set `chan` based on whether `inst.channel` exists, and adds a test validating the correct parsing of a schedule with a barrier. Note that I've also removed the `phase` and `freq` variables set at the same time as `chan`, as they were only used in one subsequent line.

For completeness, the important thing to note here is that nothing actually needs to be done to "support" barrier instructions:
- Barrier instructions impact how `block_to_schedule` converts `ScheduleBlock` instances to `Schedule` instances.
- Once in `Schedule` form all instructions have explicit absolute timing, so the barrier instructions in the schedule are actually redundant.
- As the Dynamics schedule -> signal converter works with these absolute times, it can actually just ignore the barrier instructions when parsing a schedule.
- Hence, the problem arising in #202 is just a result of a trivial mishandling of the barrier instruction object.

### DIFF
--- a/qiskit_dynamics/pulse/pulse_to_signals.py
+++ b/qiskit_dynamics/pulse/pulse_to_signals.py
@@ -164,7 +164,7 @@ class InstructionToSignals:
 
         for start_sample, inst in schedule.instructions:
             # get channel name if instruction has it
-            chan = inst.channel.name if hasattr(inst, "channel") else None
+            chan = getattr(inst, "channel", None)
 
             if isinstance(inst, Play):
                 # get the instruction samples
--- a/qiskit_dynamics/pulse/pulse_to_signals.py
+++ b/qiskit_dynamics/pulse/pulse_to_signals.py
@@ -163,9 +163,9 @@ class InstructionToSignals:
             )
 
         for start_sample, inst in schedule.instructions:
-            chan = inst.channel.name
-            phi = phases[chan]
-            freq = frequency_shifts[chan]
+            # get channel name if instruction has it
+            chan = inst.channel.name if hasattr(inst, "channel") else None
+
             if isinstance(inst, Play):
                 # get the instruction samples
                 inst_samples = None
@@ -178,8 +178,8 @@ class InstructionToSignals:
                 times = self._dt * (start_sample + np.arange(len(inst_samples)))
                 samples = inst_samples * np.exp(
                     Array(
-                        2.0j * np.pi * freq * times
-                        + 1.0j * phi
+                        2.0j * np.pi * frequency_shifts[chan] * times
+                        + 1.0j * phases[chan]
                         + 2.0j * np.pi * phase_accumulations[chan]
                     )
                 )
--- a/qiskit_dynamics/pulse/pulse_to_signals.py
+++ b/qiskit_dynamics/pulse/pulse_to_signals.py
@@ -164,7 +164,7 @@ class InstructionToSignals:
 
         for start_sample, inst in schedule.instructions:
             # get channel name if instruction has it
-            chan = getattr(inst, "channel", None)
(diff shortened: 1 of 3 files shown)


ihincks (approved):
LGTM, optional suggestion included.

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -163,9 +163,9 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
             )
 
         for start_sample, inst in schedule.instructions:
-            chan = inst.channel.name
-            phi = phases[chan]
-            freq = frequency_shifts[chan]
+            # get channel name if instruction has it
+            chan = getattr(inst, "channel", None)
```
ihincks, line 167: oops, didn't see the name in there

```suggestion
            chan = inst.channel.name if hasattr(inst, "channel") else None
```

ihincks (approved):

PR #207, opened 2023-03-22
Add arraylias demo repo

to24toro:
### Summary
I tested arraylike demo in the notebook.

### Details and comments

In docs(https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike), 

Among others this includes the likes of:

- Scalars.

- (Nested) sequences.

- Objects implementing the __array__ protocol.

So, we can use ArrayLike as a type hint which includes List, numpy.array, jax.numpy.array, and tensorflow.Tensor.


DanPuzzuoli:
One concern this raises is that using numpy's `ArrayLike` may imply that anything implementing `__array__` is valid, whereas we want to imply something more vague than that: `ArrayLike` means anything that a given instance of a numpy alias recognizes :).

This makes me lean towards defining our own `ArrayLike`, and documenting that it is synonymous with whatever is returned by `Alias.registered_types()`. @chriseclectic do you think it would make sense to add a static class property to `Alias` generically called something like `RegisteredType`, and we can override it for `numpy_alias` to be called `ArrayLike`? So, e.g. in a file in Dynamics, we could do:

```
from qiskit_dynamics.arraylias_state import DYNAMICS_ALIAS
ArrayLike = DYNAMICS_ALIAS.RegisteredType

def some_function(a: ArrayLike, b: ArrayLike):
     ...
```

For comparison, JAX defines their own `ArrayLike` [here](https://github.com/google/jax/blob/8bb90b5fbeeb7780a920e28c62bdb3f318be3d8f/jax/_src/basearray.pyi#L213). This isn't really analogous to the intended vagueness of what we're trying to do, but I just thought it was interesting they used the same type-hint name.

to24toro:
My concerns consist of two points.
- `ArrayLike` defined in `numpy` contains a type `List`. But there may be a situation we want to limit the type to either `np.array` or `jnp.array`.
- In dynamics, we check array class by using `isinstance`. We cannot use ArrayLike` defined in `numpy` in this case.
E.g. we could define:
```
@runtime_checkable
class ArrayLike(Protocol):
    def __array__(self):
        pass
```

If we use this `ArrayLike`, we use it as a type hint that has __array__ and can use 
```
isinstance(np.array([1.2]), ArrayLike)
```

It is one way of defining our own `ArrayLike`.

PR #208, opened 2023-03-24
Arraylias integration - Signal -

to24toro:
### Summary
Continuation of https://github.com/Qiskit/qiskit-dynamics/pull/197.

### Details and comments
Current status.
- ~Use ArrayLike defined by numpy~. -> define an original ArrayLike.
- Can pass tests in py310.
- ~Fail to pass test in JAX.~ -> pass tests in JAX
- Add TestNumpyBase and distinguish between Numpy and JAX tests.


qiskit_dynamics/signals/signals.py
```diff
@@ -569,7 +546,7 @@ def __init__(self, *signals, name: Optional[str] = None):
             elif isinstance(sig, Signal):
                 components.append(sig)
             elif isinstance(sig, (int, float, complex)) or (
-                isinstance(sig, Array) and sig.ndim == 0
+                hasattr(sig, "__array__") and sig.ndim == 0
```
to24toro, line 549: `ArrayLike` cannot be used in `isinstance`. So we have no choice but to use `hasattr`.

test/dynamics/signals/test_signals.py
```diff
@@ -30,7 +31,7 @@
     pass
 
 
-class TestSignal(QiskitDynamicsTestCase):
+class TestSignal(QiskitDynamicsTestCase, TestNumpyBase):
```
to24toro, line 34: Added TestNumpyBase.

test/dynamics/signals/test_signals.py
```diff
@@ -873,31 +874,26 @@ def test_DiscreteSignalSum_iterator(self):
         self.assertAllClose(sum_val, self.discrete_sig_sum(3.0))
 
 
-class TestSignalJax(TestSignal, TestJaxBase):
+class TestSignalJax(TestJaxBase, TestSignal):
```
to24toro, line 877: Method inherited at first has higher priority. (TestSignalJax use `asarray` in TestJaxBase.)

qiskit_dynamics/signals/signals.py
```diff
@@ -299,38 +286,25 @@ def __init__(
...
-        samples = Array(samples)
+        samples = unp.asarray(samples)
 
         if len(samples) == 0:
-            zero_pad = np.array([0])
+            zero_pad = unp.asarray([0])
         else:
-            zero_pad = np.expand_dims(np.zeros_like(Array(samples[0])), 0)
-        self._padded_samples = np.append(samples, zero_pad, axis=0)
+            zero_pad = np.expand_dims(np.zeros_like(unp.asarray(samples[0])), 0)
```
DanPuzzuoli, line 294: Should this line `unp`? I think for safety it may make sense to change it to `unp`.

qiskit_dynamics/signals/signals.py
```diff
@@ -406,12 +380,12 @@ def dt(self) -> float:
...
 
     @property
-    def samples(self) -> Array:
+    def samples(self) -> ArrayLike:
         """
         Returns:
             samples: the samples of the piecewise constant signal.
         """
-        return Array(self._padded_samples[:-1])
+        return unp.asarray(self._padded_samples[:-1])
```
DanPuzzuoli, line 388: This can be changed to 

```suggestion
        return self._padded_samples[:-1]
```
The reason we had `Array(self._padded_samples[:-1])` before was that `self._padded_samples` is an `Array`, but `self._padded_samples[:-1]` is actually the same as `self._padded_samples.data[:-1]` (i.e. it returns whatever the internal array is, rather than an `Array(self._padded_samples.data[:-1])`. Here we don't need to wrap or convert anything after indexing.

qiskit_dynamics/signals/signals.py
```diff
@@ -441,24 +415,24 @@ def add_samples(self, start_sample: int, samples: List):
...
+        samples = unp.asarray(samples)
 
         if len(samples) < 1:
             return
 
         if start_sample < len(self.samples):
             raise QiskitError("Samples can only be added afer the last sample.")
 
-        zero_pad = np.expand_dims(np.zeros_like(Array(samples[0])), 0)
+        zero_pad = np.expand_dims(np.zeros_like(unp.asarray(samples[0])), 0)
```
DanPuzzuoli, line 426: I think like the above case these `np`s should be changed to `unp`s. I think this is fine for JAX, but I wonder if it would pose problems for tensorflow?

DanPuzzuoli (commented):
I think it's looking good - I only had very minor comments about the actual content of `signals.py`.

Other comments:
- Should `TestPulseToSignals` also be updated to use `TestNumpyBase`? Not sure if this is necessary here.
- The test file `test.dynamics.signals.test_signals_algebra.TestSignalAlgebraJaxTransformations` is failing. Not sure what's going on here.

qiskit_dynamics/signals/signals.py
to24toro: modified at https://github.com/Qiskit/qiskit-dynamics/pull/208/commits/10742b902c76863ba805c983c4dca91db22afd81.

qiskit_dynamics/signals/signals.py
to24toro: modified at https://github.com/Qiskit/qiskit-dynamics/pull/208/commits/10742b902c76863ba805c983c4dca91db22afd81.

qiskit_dynamics/signals/signals.py
to24toro: Thanks. removed `unp.asarray` at https://github.com/Qiskit/qiskit-dynamics/pull/208/commits/d448b0ef14b6fff9b9890e787cb2c0698d590307.

to24toro:
Thanks, Dan.

I response to your comments.

> The test file test.dynamics.signals.test_signals_algebra.TestSignalAlgebraJaxTransformations is failing. Not sure what's going on here.

I haven't checked this test yet. The reason why the test fails is due to using Array().data. I modified at https://github.com/Qiskit/qiskit-dynamics/pull/208/commits/024acf90cef9695b281740fe1ea4ebf05ea401ea.

> Should TestPulseToSignals also be updated to use TestNumpyBase? Not sure if this is necessary here.

I can't say for sure, but I think there is a case of adding TestNumpyBase. Should we  develop a pulse part in this PR?

DanPuzzuoli:
> Thanks, Dan.
> 
> I response to your comments.
> 
> > The test file test.dynamics.signals.test_signals_algebra.TestSignalAlgebraJaxTransformations is failing. Not sure what's going on here.
> 
> I haven't checked this test yet. The reason why the test fails is due to using Array().data. I modified at [024acf9](https://github.com/Qiskit/qiskit-dynamics/commit/024acf90cef9695b281740fe1ea4ebf05ea401ea).
> 
> > Should TestPulseToSignals also be updated to use TestNumpyBase? Not sure if this is necessary here.
> 
> I can't say for sure, but I think there is a case of adding TestNumpyBase. Should we develop a pulse part in this PR?

I just checked out your latest version and `test.dynamics.signals.test_signals_algebra.TestSignalAlgebraJaxTransformations` is now working. Something interesting to note: If I checkout `main` and run all of the tests in `test.dynamics.signals.test_signals_algebra`, it takes about 7 seconds, whereas on this PR branch it only takes about 3 seconds, so there is a nice speedup happening there! I ran all of the tests in `test.dynamics.signals` in both branches and there was about a 4 second difference which is fully accounted for by the speedup in `test_signals_algebra`. I'm guessing `test_signals_algebra` has more array operations happening than the other ones, so that would make sense.

I don't think it's necessary to develop the pulse part in this PR, that would be too much.

DanPuzzuoli (approved):

PR #210, opened 2023-03-30
Temporary upper bound on JAX version <=0.4.6

### LINKED ISSUE #205: Update dynamics_backend.rst tutorial after next terra release
In the `DynamicsBackend` tutorial, it was necessary to add the `CRPulseGate` instruction to the `target` to make sure the transpiler functions as desired when running the `CrossResonanceHamiltonian` experiment. This is a work around that shouldn't b necessary.

This behaviour is fixed in [Qiskit terra PR 9783](https://github.com/Qiskit/qiskit-terra/issues/9783), so we should remove this line from the tutorial after the next terra release.

DanPuzzuoli:
### Summary

Closes #205.

JAX 0.4.4 introduced a bug that was caught by a Dynamics test ([discussed here](https://github.com/google/jax/discussions/14561)). This was temporarily fixed by #189 , which bypassed the new code. As of JAX 0.4.7, this bypass can no longer be used. This PR sets the upper bound to 0.4.6, the latest version that the bypass works.

To get the tests functioning, this PR also includes corrections to the `DynamicsBackend` tutorial.

### DIFF
--- a/docs/tutorials/dynamics_backend.rst
+++ b/docs/tutorials/dynamics_backend.rst
@@ -473,11 +473,6 @@ values for the single qubit gates calibrated above.
 
     from qiskit_experiments.library import CrossResonanceHamiltonian
 
-    backend.target.add_instruction(
-        instruction=CrossResonanceHamiltonian.CRPulseGate(width=Parameter("width")), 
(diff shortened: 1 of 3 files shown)


to24toro (approved):
Thanks. LGTM.

PR #211, opened 2023-04-04
Fix multiset ordering bug in perturbation module

DanPuzzuoli:
### Summary

This is a technical bug that I just discovered while trying to rerun the [speed benchmarks for the perturbation theory algorithms paper](https://github.com/DanPuzzuoli/multivariable_dyson_magnus) that the perturbation module is based on. It turns out this bug has existed since this module was introduced (in version 0.3.0), however it was missed up until this point as:
1. It requires an example of sufficient complexity (>=10 perturbations), and
2. The speed benchmarks, which were the only example we had run with >=10 perturbations, were run using an old commit in which the bug hadn't yet been introduced.

The issue arises in the file `perturbation/dyson_magnus.py`, which contains a bunch of manipulation of lists of multisets. Many of the functions contain arguments that are assumed to be lists of multisets that are "ordered and complete". Here, "complete" means the list is closed under taking submultisets (i.e. for any multiset in the list, all sub multisets are in the list), and "ordered" means ordered according to a < b iff:
1. Either, len(a) < len(b), or
2. if len(a) == len(b), sortedlist(a) "<" sortedlist(b), where sortedlist(a) means `a` represented as a sorted list (with repeated entries), and "<" means lexicographic ordering of these lists. (Implicit in this is the assumption that the elements of `a` and `b` are themselves ordered, which is fine as we explicitly demand the multisets be composed of integers.)

This may seem kind of convoluted, but for technical reasons it is convenient to order in this way when constructing the perturbation theory computations. (More succinctly, the ordering within multisets of equal size is the "lexicographic" ordering of their sorted-list representation.)

The bug was ultimately introduced when we started using the external `multiset` package, as opposed to a `Multiset` class I had personally written. There is no problem with the `multiset` package itself, I had just made an error during the conversion in which I incorrectly implemented the above ordering criteria. The incorrect implementation was based on a string representation of the multiset, which resulted `Multiset([10]) < Multiset([1]) == True`. This led to these lists being incorrectly ordered when dealing with at least 10 perturbations, and ultimately led to malfunctioning of the rest of the code which assumes ordering as described above. The ordering, however, was fine so long as less than 10 perturbations were used, which is how this error went unnoticed for so long.

### Details and comments

I have corrected the `_sorted_multisets` function in `multiset_utils.py` with the correct ordering. I've had to use a workaround in which a dummy class is defined that implements the `__lt__` dunder method, and then the multisets are sorted using this as the key. This was suggested by @ihincks as the standard way to pass an ordering to the `key` argument of `list.sort` (that can't be represented as a single valued key function). I've also added a test that failed before this change, but passes now.

I'm currently labelling this as a draft as I want to take a bit of time to look through the code to see if any un-explicit ordering assumptions are being made. I think one example is line [dyson_magnus.py line 838](https://github.com/Qiskit/qiskit-dynamics/blob/002f2a2a65e3947768dac08a762a1fdfb1579f0f/qiskit_dynamics/perturbation/dyson_magnus.py#L838). It currently reads
```
            lmult_rule.append((np.array([1.0, 1.0]), np.array([[-1, term_idx], [term_idx, -1]])))
```
but I think it should be
```
            lmult_rule.append((np.array([1.0, 1.0]), np.array([[-1, term_idx], [perturbation_labels.index(term), -1]])))
```

### DIFF
--- a/qiskit_dynamics/perturbation/multiset_utils.py
+++ b/qiskit_dynamics/perturbation/multiset_utils.py
@@ -89,7 +89,6 @@ def _sorted_multisets(multisets: Iterable[Multiset]) -> List[Multiset]:
     """Sort in non-decreasing order according to the ordering described in the dummy class
     _MultisetSort.
     """
-
     return sorted(multisets, key=_MultisetSort)
 
 
--- a/qiskit_dynamics/perturbation/multiset_utils.py
+++ b/qiskit_dynamics/perturbation/multiset_utils.py
@@ -15,7 +15,7 @@
 Utility functions for working with multisets.
 """
 
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Iterable
 import itertools
 
 from multiset import Multiset
--- a/qiskit_dynamics/perturbation/multiset_utils.py
+++ b/qiskit_dynamics/perturbation/multiset_utils.py
@@ -46,15 +46,55 @@ def _multiset_to_sorted_list(multiset: Multiset) -> List:
     return sorted_list
 
 
-def _sorted_multisets(multisets: List[Multiset]) -> List[Multiset]:
-    """Sort in non-decreasing order according to:
+class _MultisetSort:
+    """Dummy class for usage as a key when sorting Multiset instances. This assumes the entries
+    if the multisets can themselves be sorted.
+    """
+    __slots__ = "multiset",
 
-    ms1 <= ms2 if len(ms1) < len(ms2), or if
-    len(ms1) == len(ms2) and if
-    str(_multiset_to_sorted_list(ms1)) <= str(_multiset_to_sorted_list(ms2)).
+    def __init__(self, multiset: Multiset):
+        self.multiset = multiset
+    
+    def __lt__(self, other: Multiset) -> bool:
+        """Implements an ordering on multisets.
+        
+        This orders first according to length (the number of elements in each multiset). If self and
+        other are the same length, self < other if, when written as fully expanded and sorted lists,
+        self < other in lexicographic ordering. E.g. it holds that Multiset({0: 2, 1: 1}) <
+        Multiset({0: 1, 1: 2}), as the list versions are [0, 0, 1], and [0, 1, 1]. Here, the first
+        element of each is 0, so there is no comparison, but the second element of the first is <
+        the second, hence the ordering on the multisets.
+        """
+        if len(self.multiset) < len(other.multiset):
+            return True
+
+        if len(other.multiset) < len(self.multiset):
+            return False
+
+        unique_self = set(self.multiset.distinct_elements())
+        unique_other = set(other.multiset.distinct_elements())
+        unique_entries = list(unique_self.union(unique_other))
+        unique_entries.sort()
+
+        for element in unique_entries:
+            self_count = self.multiset[element]
+            other_count = other.multiset[element]
+
+            if self_count < other_count:
+                return False
+
+            if self_count > other_count:
+                return True
+
+        return False
+
+
+def _sorted_multisets(multisets: List[Multiset]) -> List[Multiset]:
+    """Sort in non-decreasing order according to the ordering described in the dummy class
+    _MultisetSort.
     """
 
-    return sorted(multisets, key=lambda x: str(len(x)) + ", " + str(_multiset_to_sorted_list(x)))
+    return sorted(multisets, key=_MultisetSort)
 
 
 def _clean_multisets(multisets: List[Multiset]) -> List[Multiset]:
--- a/qiskit_dynamics/perturbation/multiset_utils.py
+++ b/qiskit_dynamics/perturbation/multiset_utils.py
@@ -46,7 +46,7 @@ def _multiset_to_sorted_list(multiset: Multiset) -> List:
     return sorted_list
 
 
-class _MultisetSort:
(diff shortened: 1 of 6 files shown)


qiskit_dynamics/perturbation/multiset_utils.py
```diff
@@ -46,15 +46,55 @@ def _multiset_to_sorted_list(multiset: Multiset) -> List:
...
+            if self_count < other_count:
+                return False
+
+            if self_count > other_count:
+                return True
+
+        return False
+
+
+def _sorted_multisets(multisets: List[Multiset]) -> List[Multiset]:
```
ihincks, line 92: If accepted, ensure `Iterable` is imported.
```suggestion
def _sorted_multisets(multisets: Iterable[Multiset]) -> List[Multiset]:
```

qiskit_dynamics/perturbation/multiset_utils.py
```diff
@@ -46,15 +46,55 @@ def _multiset_to_sorted_list(multiset: Multiset) -> List:
...
+            other_count = other.multiset[element]
 
-    ms1 <= ms2 if len(ms1) < len(ms2), or if
-    len(ms1) == len(ms2) and if
-    str(_multiset_to_sorted_list(ms1)) <= str(_multiset_to_sorted_list(ms2)).
+            if self_count < other_count:
+                return False
+
+            if self_count > other_count:
+                return True
```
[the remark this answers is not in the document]
DanPuzzuoli: > This is probably stylistic more than anything, but it does reduce the number of ifs (same can be applied to higher up in the function). BTW, without understanding the context at all, I'm slightly surprised to see True being returned if self_count > other_count. Intentional?

Yes this is intentional. This is the "lexicographic" ordering. E.g. for multisets `a = Multiset({0: 2, 1: 1})` and `b = Multiset({0: 1, 1: 2})` (the dictionaries are the count dictionaries), when you convert them into sorted lists you get: `sortedlist(a) = [0, 0, 1]` and `sortedlist(b) = [0, 1, 1]`. This loop + if statements are equivalent to:
```
for a_entry, b_entry in zip(sortedlist(a), sortedlist(b)):
    if a_entry != b_entry:
        return a_entry < b_entry
```
i.e. it's like alphabetical ordering on the lists. The code in `__lt__` does this without actually converting these to lists by utilizing the multiset counts.

Ultimately we could have reversed this, but I naturally chose this ordering as it's how I typically generate lists of multisets by hand, e.g.:
```
Multiset([0, 0]), Multiset([0, 1]), Multiset([1, 1])
```
i.e. by incrementing entries analogous to increasing numbers in a place-value number system.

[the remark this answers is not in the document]
ihincks: Ah I see. I was specifically wondering about `>` vs `<`, but you have explained that you indeed want `>` because it indicates that lhs has more smaller things, ie is smaller.

DanPuzzuoli:
This PR is ready for review.

Changes:
- Corrects the multiset sorting function as described in the initial comment.
- Modifies lines 837 and 838 in `dyson_magnus.py` along the lines as described in the initial comment. This ensures the correct index of the `perturbation_labels` is used in the lmult rule, and now only adds the rule if `term` is in `perturbation_labels`. This added line corresponds to the RHS term for the differential equation of a first order Dyson term. If `term` is not in `perturbation_labels`, this Dyson term will be 0 for all times, so there is no need to include it as part of the RHS function (the derivative is always 0). This will save some unnecessary matrix multiplications (albeit in unusual scenarios).
- Adds sorting tests for `_sorted_multisets` that would have caught the original bug.
- Adds tests validating point 2 above:
   - One checks correct exclusion of the unnecessary matrix multiplications.
   - The other checks correct multiplication rule construction in `perturbation_labels` is not in "canonical" order (this is the only list of multisets that technically shouldn't be assumed to be in canonical order).
- Adds an "integration" test that also would have caught the original bug. The test case `test_dyson_analytic_case1_1d` is one of the cheapest-to-compute test cases for `solve_lmde_perturbation`, and this PR adds `test_dyson_analytic_case1_1d_relabeled`, which reproduces this test but relabels the perturbations with indices `1` and `10`, which is one case where the bug was expressed.

I've verified that all added tests fail if the changes in points 1 and 2 are reverted to what is currently in `main`.

qiskit_dynamics/perturbation/multiset_utils.py
```diff
@@ -46,15 +46,51 @@ def _multiset_to_sorted_list(multiset: Multiset) -> List:
     return sorted_list
 
 
-def _sorted_multisets(multisets: List[Multiset]) -> List[Multiset]:
-    """Sort in non-decreasing order according to:
-
-    ms1 <= ms2 if len(ms1) < len(ms2), or if
-    len(ms1) == len(ms2) and if
-    str(_multiset_to_sorted_list(ms1)) <= str(_multiset_to_sorted_list(ms2)).
+class _MultisetSort:
```
ihincks, line 49: nit: consider renaming to _MultisetSortKey

ihincks (approved):
the parts i understand lgtm

qiskit_dynamics/perturbation/multiset_utils.py
DanPuzzuoli: Changed.

ihincks (approved):

PR #214, opened 2023-04-05
Arraylias integration - Rotating frame -

to24toro:
### Summary
I started to migrate Arraylias to rotating_frame.py.
The first difficulty is dealing with csr_matrix and BCOO.

### Details and comments
[TODO]
- how to deal with `Operator`.  I think the way is whether converting to `op.data` or  registering `Operator` to ArrayLike.
-  how to use matmul in `sparse matrix` or `BCOO`.
- how to set `spmatrix`. Is it necessary to add some concrete classes of a sparse matrix.


DanPuzzuoli:
The `to_array`, `to_csr`, `to_BCOO`, and `to_numeric_matrix_type` are things that we'll need to think carefully about when converting to arraylias. Dynamics has been very loose with array types, and that has been partly enabled by these functions, which are used throughout the package to funnel a variety of types (including lists, lists of arrays, etc) into array and sparse types. It would be nice to not need these methods, but I'm not sure to what extent we can avoid using them. 

One option would be to just start being very strict about types. E.g. the `RotatingFrame` methods could all require the inputs be `ArrayLike` (assuming `csr_matrix` and `BCOO` become registered Arraylias types). The reason I hesitate with fully going this direction is that it's natural to specify operators in models in a variety of ways, either as raw arrays/sparse matrices, qiskit Operator/QuantumState instances, qutip Qobj, or lists of these. We _could_ just force users to convert these to their desired array types in most places, and make the `to_*` functions into publically available helper functions. I wouldn't want to do this _everywhere_, e.g. in `Solver.solve`, there is some high level handling of `QuantumState` types that I wouldn't want to get rid of.

Aside from these concerns, I don't think we want to register `csr_matrix` or `BCOO` as straight `'numpy'`/`'jax'` types. Both require special syntax, so I think it will be necessary to register them as their own libs. E.g., in `arraylias_state.py`, I think we can do something like (with suitable `try` blocks in case JAX isn't installed):
```
DYNAMICS_ALIAS.register_type(csr_matrix, lib="scipy_sparse")

"""
register required custom versions of functions for csr type here
"""

DYNAMICS_ALIAS.register_type(BCOO, lib="jax_sparse")
"""
register required custom versions of functions for BCOO type here
"""
```

I think for scipy, there are also multiple different sparse types that we'll need to register (they get transformed into each other). The `scipy.sparse` function `issparse` is used to check for all of these types. Maybe we can look there for all of the types.

Some considerations with functions that need to be registered:
- For some reason, the sparse types in scipy use `*` for matrix multiplication instead of `@`. There are places in `RotatingFrame` where the `issparse` checks for a sparse type, and a line is implemented using `*`. I think this type checking can be eliminated with Arraylias if we convert matrix multiplications in `RotatingFrame` to be written with `unp.matmul`, and we register `lambda x, y: x * y` as the definition of `matmul` for `lib='scipy_sparse'`.
- The `BCOO` type has been set up interestingly in JAX to use the `sparsify` transformation. This takes a JAX function meant to act on dense arrays, and converts it into one that also works with sparse arrays. (It doesn't work with absolutely everything in JAX, but it does for everything we'll care about.) Checkout the [JAX documentation on this](https://jax.readthedocs.io/en/latest/jax.experimental.sparse.html), and also the [operator_collections.py](https://github.com/Qiskit/qiskit-dynamics/blob/main/qiskit_dynamics/models/operator_collections.py) file for examples in Dynamics. I think the things being registered in `operator_collections.py` can probably be registered in the main `arraylias_state.py` file to be used throughout the rest of Dynamics.
- The `asarray` method will need to be registered in both cases, and I think this is what your `'csr_matrix'` registered function should be.

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -86,7 +88,8 @@ def __init__(
             # if Hermitian convert to anti-Hermitian
             frame_operator = _enforce_anti_herm(frame_operator, atol=atol, rtol=rtol)
 
-            self._frame_diag = Array(frame_operator)
+            # TODO: need to define asarray for Operator
```
to24toro, line 91: need to define asarray for Operator. Or
```
if op is `Operator`:
  op = op.data
```

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -188,6 +193,7 @@ def operator_into_frame_basis(
 
         if isinstance(op, list):
             return [self.operator_into_frame_basis(x, convert_type=False) for x in op]
+        # Is it unnecessary to separate cases if they are integrated into arraylias?
```
to24toro, line 196: [Q] Can we put it together using arraylias?

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -338,10 +345,7 @@ def _conjugate_and_add(
                 return operator
             else:
                 if op_to_add_in_fb is not None:
-                    if issparse(operator):
-                        op_to_add_in_fb = csr_matrix(op_to_add_in_fb)
-                    elif type(operator).__name__ == "BCOO":
-                        op_to_add_in_fb = to_BCOO(op_to_add_in_fb)
+                    op_to_add_in_fb = unp.asarray(op_to_add_in_fb)
```
to24toro, line 348: We can use `unp.asarray` by defining `asarray` for `BCOO` and `csr_matrix`.

qiskit_dynamics/type_utils.py
```diff
@@ -353,7 +355,7 @@ def isinstance_qutip_qobj(obj):
 
 
 # pylint: disable=too-many-return-statements
-def to_array(op: Union[Operator, Array, List[Operator], List[Array], spmatrix], no_iter=False):
```
to24toro, line 356: The `spmatrix` class is abstract one. 
- leave it to the type hint.
- delete it and we need to register additional concrete class such as `csc_matrix` to ArrayLike.

qiskit_dynamics/arraylias_state.py
```diff
@@ -16,14 +16,36 @@
...
+    pass
 
 
 DYNAMICS_ALIAS = numpy_alias()
 
 # Set qiskit_dynamics.array.Array to be dispatched to numpy
 DYNAMICS_ALIAS.register_type(Array, "numpy")
 
+# register required custom versions of functions for csr type here
+DYNAMICS_ALIAS.register_type(csr_matrix, lib="scipy_sparse")
```
to24toro, line 33: Is it necessary to add other classes which is a subclass of `spmatrix`?
- DYNAMICS_ALIAS.register_type(csc_matrix, lib="scipy_sparse")
- DYNAMICS_ALIAS.register_type(coo_matrix, lib="scipy_sparse")

qiskit_dynamics/models/rotating_frame.py
```diff
@@ -188,6 +193,7 @@ def operator_into_frame_basis(
 
         if isinstance(op, list):
             return [self.operator_into_frame_basis(x, convert_type=False) for x in op]
+        # Is it unnecessary to separate cases if they are integrated into arraylias?
         elif type(op).__name__ == "BCOO":
             return self.frame_basis_adjoint @ jsparse_matmul(op, self.frame_basis.data)
```
to24toro, line 198: [Q] Can we put it together using arraylias?

to24toro:
close and move to https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/282.

PR #215, opened 2023-04-21
Bump Sphinx Theme to 1.11

Eric-Arellano:
### Summary

Major changes include:

1. Fixing IBM fonts not loading
2. Fixing jQuery missing and some JavaScript errors
3. Making the right page table of contents sticky when you scroll down the page.

Screenshot 2023-04-21 at 11 09 30 AM

### Details and comments

This also fixes `pyproject.toml` to add the missing `build-backend` key. And it removes `wheel` from what is required, since you don't actually need that.

Finally, it fixes `tox -e docs` not working on macOS by setting `usedevelop = False` like we had to do in Terra.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -39,7 +39,8 @@ extensions = [
     'reno.sphinxext',
     'sphinx.ext.intersphinx',
     'nbsphinx',
-    'sphinxcontrib.bibtex'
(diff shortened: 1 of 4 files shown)


DanPuzzuoli (approved):
Thanks for this update, it's nice to be able to do `tox -e docs` again :).

PR #232, opened 2023-06-16
Add warning for JAX versions on import of Dynamics

### LINKED ISSUE #231: Incompatibility with jax 0.4.11
With jax >=0.4.11, `import qiskit_dynamics` fails trying to import `DeviceArray` from `jax` which was removed in 0.4.11. The end of the traceback is:

```
  File "/home/conda/feedstock_root/build_artifacts/qiskit-dynamics_1686354044341/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/python3.11/site-packages/qiskit_dynamics/dispatch/backends/jax.py", line 20, in <module>
    from jax.interpreters.xla import DeviceArray
ImportError: cannot import name 'DeviceArray' from 'jax.interpreters.xla'
```

I saw this [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=720801&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d&l=525) in the conda-forge build of qiskit-dynamics. The documentation of the removal from jax is [here](https://jax.readthedocs.io/en/latest/changelog.html#jax-0-4-11-may-31-2023).

DanPuzzuoli:
### Summary

Closes #231 

This PR updates the JAX type registration in `dispatch`, and also adds a warning at JAX import time about the version bounds for JAX.

### Details and comments

The docs build has also been set to run with JAX version 0.4.3 to avoid this warning.

### DIFF
--- a/qiskit_dynamics/dispatch/backends/jax.py
+++ b/qiskit_dynamics/dispatch/backends/jax.py
@@ -20,6 +20,19 @@ try:
     from jax import Array
     from jax.core import Tracer
 
+    # warning based on JAX version
+    from packaging import version
+    import warnings
+    if version.parse(jax.__version__) >= version.parse("0.4.4"):
+        warnings.warn(
+            "The functionality in the perturbation module of Qiskit Dynamics requires a JAX "
+            "version <= 0.4.6, due to a bug in JAX versions > 0.4.6. For versions 0.4.4, 0.4.5, "
+            "and 0.4.6, using the perturbation module functionality requires setting "
+            "os.environ['JAX_JIT_PJIT_API_MERGE'] = '0' before importing JAX."
+        )
+
+
+
     JAX_TYPES = (Array, Tracer)
 
     try:
--- a/qiskit_dynamics/dispatch/backends/jax.py
+++ b/qiskit_dynamics/dispatch/backends/jax.py
@@ -26,7 +26,11 @@ try:
 
     if version.parse(jax.__version__) >= version.parse("0.4.4"):
         import os
-        if version.parse(jax.__version__) > version.parse("0.4.6") or os.environ.get('JAX_JIT_PJIT_API_MERGE', None) != '0':
+
+        if (
+            version.parse(jax.__version__) > version.parse("0.4.6")
+            or os.environ.get("JAX_JIT_PJIT_API_MERGE", None) != "0"
+        ):
             warnings.warn(
                 "The functionality in the perturbation module of Qiskit Dynamics requires a JAX "
                 "version <= 0.4.6, due to a bug in JAX versions > 0.4.6. For versions 0.4.4, "
--- a/qiskit_dynamics/dispatch/backends/jax.py
+++ b/qiskit_dynamics/dispatch/backends/jax.py
@@ -25,12 +25,14 @@ try:
     import warnings
 
     if version.parse(jax.__version__) >= version.parse("0.4.4"):
-        warnings.warn(
-            "The functionality in the perturbation module of Qiskit Dynamics requires a JAX "
-            "version <= 0.4.6, due to a bug in JAX versions > 0.4.6. For versions 0.4.4, 0.4.5, "
-            "and 0.4.6, using the perturbation module functionality requires setting "
-            "os.environ['JAX_JIT_PJIT_API_MERGE'] = '0' before importing JAX."
-        )
+        import os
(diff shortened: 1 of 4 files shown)


qiskit_dynamics/dispatch/backends/jax.py
```diff
@@ -17,26 +17,29 @@
...
 
+    # in versions <= 0.4.10
     try:
-        # This class is not in older versions of Jax
-        from jax.interpreters.partial_eval import DynamicJaxprTracer
+        # pylint: disable=ungrouped-imports
+        from jax.interpreters.xla import DeviceArray
 
-        JAX_TYPES += (DynamicJaxprTracer,)
+        JAX_TYPES += (DeviceArray,)
```
wshanks, line 42: I don't know jax well enough to judge -- is it okay to drop the extra tracer types in this PR?

qiskit_dynamics/dispatch/backends/jax.py
```diff
@@ -17,26 +17,29 @@
...
 
-    try:
-        # This class was introduced in 0.4.0
-        from jax import Array
+    if version.parse(jax.__version__) >= version.parse("0.4.4"):
+        warnings.warn(
+            "The functionality in the perturbation module of Qiskit Dynamics requires a JAX "
+            "version <= 0.4.6, due to a bug in JAX versions > 0.4.6. For versions 0.4.4, 0.4.5, "
+            "and 0.4.6, using the perturbation module functionality requires setting "
+            "os.environ['JAX_JIT_PJIT_API_MERGE'] = '0' before importing JAX."
```
wshanks, line 32: Perhaps you could check `os.environ` and not warn if this is set?

tox.ini
```diff
@@ -41,8 +41,8 @@ commands = black {posargs} qiskit_dynamics test
 usedevelop = False
 deps =
     -r{toxinidir}/requirements-dev.txt
-    jax<=0.4.6
-    jaxlib<=0.4.6
+    jax<=0.4.3
```
wshanks, line 44: You might want a comment here to explain the lower pin. Or maybe you don't need the lower pin if you follow my suggestion of checking the environment variable before warning.

wshanks (dismissed):
I had a couple suggestions but this looks fine to me (other than the tracer type changes which I am not sure about).

qiskit_dynamics/dispatch/backends/jax.py
DanPuzzuoli: Ah that's a really good idea, I'll make this change.

qiskit_dynamics/dispatch/backends/jax.py
DanPuzzuoli: Yeah I'm not 100% sure but with the arraylias work this appears to be fine (also all the Dynamics tests passed with this change). They used to have a zoo of tracer and abstract array types but in more recent version they've introduced the `Array` and `Tracer`, which i'm assuming the other types inherit from.

tox.ini
DanPuzzuoli: I'll implement your suggestion, making this moot.

DanPuzzuoli:
@wshanks I've implemented your suggestion for the warning to not trigger if the JAX version is 0.4.4, 0.4.5, or 0.4.6 and they've already set the `os` environment variable. I've reverted the docs to use JAX 0.4.6 as the warning will no longer be triggered.

DanPuzzuoli:
So I still ended up needing to set `os.environ["JAX_JIT_PJIT_API_MERGE"] = "0"` in  `docs/conf.py` to avoid the warning in the docs build, but luckily that worked. I've put that in with a comment, and have added a description of this PR to issue #190 (where I'm keeping track of all the little changes that we've had to make because of these JAX issues).

I was actually also able to completely remove adding `DeviceArray` to `JAX_TYPES` - `Array` and `Tracer` turn out to be all that's necessary these days.

qiskit_dynamics/dispatch/backends/jax.py
DanPuzzuoli: See the latest comment on the PR, but I've actually removed `DeviceArray` completely now. Only `Array` and `Tracer` are required for any JAX version above `0.4.0`.

wshanks (approved):
Looks good!

PR #239, opened 2023-07-05
Add probability normalization to DynamicsBackend sampling routine

### LINKED ISSUE #237: The probability solved in Dynamicsbackend exceeds 1.
### Informations

- **Qiskit Dynamics version**: 0.4.0
- **Python version**: 3.10.10
- **Operating system**: M1 MacOS
- **Qiskit Terra**: You need to pull https://github.com/Qiskit-Extensions/qiskit-experiments/pull/1199.

Dynamicsbackend is utilized for a pulse simulation and useful for the simulation of the calibration.
When we pull the above PR and simulate the rough amplitude calibration of a cross resonance gate, we have the probability dynamics backend calculated exceeds 1.

### Suggested solutions
This error raises because there may be a rounding error in the way the probability is calculated and the deviation from one is too large to surpress an error when rng.choice is called.
We need to fix the way of normalizing stetes (https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/37d12ac33fffd57daea3c9d5f9cda3f00acc37ea/qiskit_dynamics/backend/dynamics_backend.py#L795)
We could fix it to normalize the probabilities instead of normalizing the state.

DanPuzzuoli:
### Summary

Closes #237

### Details and comments

When looking at the implementation I realized that two normalizations are necessary; it isn't possible to only normalize the probability before sampling. The state needs to be normalized before calculating probabilities so that the `QuantumState.probabilities_dict` function doesn't raise an error, and the resultant probabilities need to then be normalized to avoid numerical error in the probability computation. 

As a result of this, I've kept `options.normalize_states`, and modified the code that samples the outcomes to also normalize the probabilities if `options.normalize_states == True`. I've modified the description of the `normalize_states` option to describe this, and added a "fixes" release note for this.

### DIFF
--- a/qiskit_dynamics/backend/backend_utils.py
+++ b/qiskit_dynamics/backend/backend_utils.py
@@ -146,7 +146,7 @@ def _get_memory_slot_probabilities(
 
 
 def _sample_probability_dict(
-    probability_dict: Dict, shots: int, seed: Optional[int] = None
+    probability_dict: Dict, shots: int, normalize_probabilities: bool = True, seed: Optional[int] = None,
 ) -> List[str]:
     """Sample outcomes based on probability dictionary.
 
@@ -154,6 +154,8 @@ def _sample_probability_dict(
         probability_dict: Dictionary representing probability distribution, with keys being
             outcomes, values being probabilities.
         shots: Number of shots.
+        normalize_probabilities: Whether or not to normalize the probabilities to sum to 1 before
+            sampling.
         seed: Seed to use in rng construction.
 
     Return:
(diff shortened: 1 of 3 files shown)


to24toro:
I am a little bit uncomfortable that `normalize_state` also computes the normalization of probability. Do you have any reason to continue using normalize_state as the variable name?

DanPuzzuoli:
My thinking on the name is:
1. Whatever we call this option, it makes sense for it to normalize at multiple points (the states before computing probabilities, and also the probabilities right before the sampling happens).
2. The current option for specifying normalization is `normalize_states`, so it seemed simpler to just keep that name, rather than changing it to `normalize_probabilities`.

If you'd like I'm open to changing it to `normalize_probabilities` though; I do think it's a better name. If so I can make this change and also add an `upgrade` release note indicating that this change has occurred. I don't think with this I will bother following the deprecation procedure as I highly doubt anyone is currently using this option.

to24toro:
Thank you for your comment.
If few users will use and change this option to False, we can leave it as is for a while. We can consider about this naming issue again when demand arises to use normalize_state and normalize_probability separately.

to24toro (approved):

PR #242, opened 2023-07-12
Add warning if digital carrier exceeds Nyquist frequency in pulse -> signal conversion

### LINKED ISSUE #241: Add warning about sampling rate when using `ShiftFrequency` and `SetFrequency` in pulse schedules
An issue arose [here](https://github.com/Qiskit-Extensions/qiskit-experiments/pull/1009#issuecomment-1631290173) when `SetFrequency` and `ShiftFrequency` were used to shift a channel frequency well away from what was specified in the `channel_carrier_freqs` argument to `Solver`. 

The specified channel frequency couldn't be represented digitally with the given `dt`, and the simulations therefore produced strange results.

@nkanazawa1989 suggested raising a warning in the schedule interpretation step when the difference between the schedule frequency and the `channel_carrier_freqs` frequency is larger than the Nyquist frequency (and therefore can't be accurately represented). One minor technical issue with adding this is that this code can be compiled with JAX, so we'll need to take some technical care with handling this.

DanPuzzuoli:
### Summary

Closes #241

A warning is added to the pulse -> signal converter if the digital carrier frequency (resulting from `SetFrequency` and `ShiftFrequency` instructions) exceeds the Nyquist frequency set by the sample size `dt`.

### Details and comments

The warning is suppressed if the conversion is called within JAX tracing, as in this case the boolean function `np.abs(frequency_shift) > 0.5 / dt` cannot be evaluated.

A test has been added, and @nkanazawa1989 I've verified that the warning is raised in your example that motivated this PR.

### DIFF
--- a/qiskit_dynamics/pulse/pulse_to_signals.py
+++ b/qiskit_dynamics/pulse/pulse_to_signals.py
@@ -139,7 +139,7 @@ class InstructionToSignals:
           :math:`\Delta\nu \mapsto \mu - \nu`, where :math:`\nu` is the analog carrier frequency.
           Similarly to ``ShiftFrequency``, the shift rule for :math:`\phi_a` is defined to maintain
           carrier wave continuity.
-        
+
         If, at any sample point :math:`k`, :math:`\Delta\nu(k)` is larger than the Nyquist sampling
         rate given by ``dt``, a warning will be raised.
 
@@ -200,7 +200,7 @@ class InstructionToSignals:
 
             if isinstance(inst, SetPhase):
                 phases[chan] = inst.phase
-            
+
             if isinstance(inst, ShiftFrequency):
                 frequency_shifts[chan] = frequency_shifts[chan] + Array(inst.frequency)
                 phase_accumulations[chan] = (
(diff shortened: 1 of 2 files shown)


qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -188,14 +198,15 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
             if isinstance(inst, ShiftPhase):
                 phases[chan] += inst.phase
 
+            if isinstance(inst, SetPhase):
```
nkanazawa1989, line 201: Is this related to the warning?

nkanazawa1989 (approved):
Thanks @DanPuzzuoli this is really helpful improvement. I was thinking shift and set frequency were directly applied to the analog carrier, and had trouble with interpretation of weird simulation results. I think I am not only person suffering this problem :)

> The warning is suppressed if the conversion is called within JAX tracing, as in this case the boolean function np.abs(frequency_shift) > 0.5 / dt cannot be evaluated.

This means the warning is also raised at run time even with the JAX backend? Then this PR looks good to me.

qiskit_dynamics/pulse/pulse_to_signals.py
DanPuzzuoli: This is not related to the warning: I saw that the `SetPhase` and `ShiftPhase` handling were not grouped together, so I moved this block up.

DanPuzzuoli:
> Thanks @DanPuzzuoli this is really helpful improvement. I was thinking shift and set frequency were directly applied to the analog carrier, and had trouble with interpretation of weird simulation results. I think I am not only person suffering this problem :)
> 
> > The warning is suppressed if the conversion is called within JAX tracing, as in this case the boolean function np.abs(frequency_shift) > 0.5 / dt cannot be evaluated.
> 
> This means the warning is also raised at run time even with the JAX backend? Then this PR looks good to me.

The error will not be raised _ever_ if the converter is called within a traced function. For now this probably won't impact any users; it will still be raised when using the `DynamicsBackend` using JAX as you are in your notebook. 

As an aside: This is an unfortunate element of  the way JAX works - you can't have warnings/errors that depend on the value of variables that are being traced. Branching logic that depends on the value of a variable will cause an error during tracing, and once compiled, the original code is no longer executed anyway, so no warnings can be raised there.

to24toro (approved):

PR #247, opened 2023-07-18
Fix bug with carrier_freq being a JAX tracer if envelope is constant in Signal

### LINKED ISSUE #245: Make Signal carrier_freq arg compatible with JAX compilation
As discussed in #243, the following function is not JAX compatible:

```
def signal_from_input(pulse_input):
    amp = Array(pulse_input[0])
    w = pulse_input[1]
    sig = Signal(amp, carrier_freq = w)
    return sig(1.)
```

This is due to the code checking if `carrier_freq == 0.0`. 

I'm pretty surprised this doesn't work - I would have thought this issue would have been caught very early in the development of the project.

DanPuzzuoli:
### Summary

Closes #245

### Details and comments

I ended up putting the `carrier_freq == 0.0` check into a `try` block to solve this. It's not the prettiest solution, but it seemed to be the simplest and most concise. I originally had checks for JAX types, then checked if `carrier_freq` was a tracer, but it was actually awkward to implement this.

The current code is essentially equivalent to:
- if `carrier_freq` is not a JAX tracer object (or a tracer object inside of an `Array`), and `carrier_freq == 0.0` then set `self._is_constant = True`.

The awkwardness with directly implementing this is that the code also needs to work if JAX isn't installed, so the "if `carrier_freq` is not a JAX tracer object" is itself an awkward thing to check.

I've also added tests verifying the original issue is resolved: 
- A check that `carrier_freq` can be properly traced within a JAX function if the envelope is constant.
- A modification to an existing test to verify that `is_constant` is properly set within JAX tracing if `carrier_freq` is concretely `0`.

### DIFF
--- a/qiskit_dynamics/signals/signals.py
+++ b/qiskit_dynamics/signals/signals.py
@@ -26,7 +26,6 @@ from matplotlib import pyplot as plt
 
 try:
     import jax.numpy as jnp
-    import jax
 except ImportError:
     pass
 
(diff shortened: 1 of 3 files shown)


to24toro (approved):
LGTM!

PR #249, opened 2023-07-21
Drop support for backendV2 in from_backend and _to_schedule_list

to24toro:
### Summary
closes https://github.com/Qiskit-Extensions/qiskit-dynamics/issues/236.

### Details and comments
The type hint of an argument `backend` in `from_backend` supports both backendV1 and backendV2. However, we found it confuses people because almost all backendV2 cannot be used for dynamics backend in reality.
 
In addition, I changed the type hint of an argument `backend` from `BackendV2` to `BackendV1` in `_to_schedule_list`. In `_to_schedule_list`, `build_schedule` is used. Scheduling with backendV2 fails because of some bugs in terra. I am assigned to work on it but cannot solve it yet.

In the future, supporting `BackendV2` again in these functions is essential for users.

### DIFF
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -544,7 +544,7 @@ class DynamicsBackend(BackendV2):
     @classmethod
     def from_backend(
         cls,
-        backend: Union[BackendV1, BackendV2],
+        backend: BackendV1,
         subsystem_list: Optional[List[int]] = None,
         rotating_frame: Optional[Union[Array, RotatingFrame, str]] = "auto",
         evaluation_mode: str = "dense",
@@ -958,7 +958,7 @@ def _get_acquire_instruction_timings(
 
 
 def _to_schedule_list(
-    run_input: List[Union[QuantumCircuit, Schedule, ScheduleBlock]], backend: BackendV2
+    run_input: List[Union[QuantumCircuit, Schedule, ScheduleBlock]], backend: BackendV1
 ):
     """Convert all inputs to schedules, and store the number of classical registers present
     in any circuits.


qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -958,7 +958,7 @@ def _get_acquire_instruction_timings(
 
 
 def _to_schedule_list(
-    run_input: List[Union[QuantumCircuit, Schedule, ScheduleBlock]], backend: BackendV2
+    run_input: List[Union[QuantumCircuit, Schedule, ScheduleBlock]], backend: BackendV1
```
DanPuzzuoli, line 961: This is actually called only on `DynamicsBackend`, which is `BackendV2`. I'm assuming it works on `DynamicsBackend` otherwise `DynamicsBackend` would never work. Is this issue in Terra in the `main` branch or on a current release?

to24toro: Oh, sorry I misunderstood it. We cannot change this type hint.
Anyway, `DynamicsBackend` would never work if it has the same architecture as backendV2. I haven't solved this issue in the current release 0.25 because I found other bugs in the current code in terra.  However, I am not sure if `DynamicsBackend` would not work in fact because for example, `DynamicsBackend`  enable to measure all qubits independently which is different from the real ibm backend.

p.s.
Scheduling with backendV2 needs `meas_map` and `DynamicsBackend`  has to have `meas_map`.

DanPuzzuoli:
The only other thing I'm wondering with this PR is, now that we are removing `BackendV2` from the type hints, should add a line to the `from_backend` doc string that this will work for `BackendV2` if it has a `configuration` and `defaults`. Just as we don't want users to think this will work for all `BackendV2`, causing confusion, we also want them to be aware that it will work with the ones that have `configuration` and `defaults` for backwards compatibility.

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -609,6 +609,9 @@ def from_backend(
 
         Args:
             backend: The ``Backend`` instance to build the :class:`.DynamicsBackend` from.
+               Note that even if the `Backend` is :class:`~qiskit.providers.backend.BackendV2`,
+               only :class:`~qiskit.providers.backend.BackendV2` with ``configuration`` and
+               ``defaults`` can be used.
```
DanPuzzuoli, line 614:
```suggestion
               Note that while the type hint indicates that `backend` should be a :class:`~qiskit.providers.backend.BackendV1` instance,
               this method also works for :class:`~qiskit.providers.backend.BackendV2` instances that have been set up with sufficiently populated ``configuration`` and ``defaults`` for backwards compatibility.
```

Just a slight suggestion on wording. This may mess up the linting.

DanPuzzuoli (changes requested):

qiskit_dynamics/backend/dynamics_backend.py
to24toro: Thank you for your correction.

DanPuzzuoli (approved):

PR #252, opened 2023-07-31
Fix ClassicalRegister counting in DynamicsBackend

### LINKED ISSUE #251: DynamicsBackend not counting number of cregs properly
This example in this issue https://github.com/Qiskit-Extensions/qiskit-experiments/issues/1235 contains an error resulting from `DynamicsBackend` miscounting the number of classical registers in a circuit/schedule. `DynamicsBackend` counts the number of classical registers in a circuit via

```
circ.cregs[0].size
```

The above error occurs as the `cregs` is of the form

```
[ClassicalRegister(1, 'c'), ClassicalRegister(1, 'c_tomo')]
```

So, `DynamicsBackend` miscounts the number as `1` but it should be `2`. This is obviously an easy fix, changing the count to

```
sum(creg size for creg in circ.cregs)
```

The only thing I'm worried about here is that I don't know how these `ClassicalRegister` lists are handled and would like to confirm. What I'm guessing is the case is:

1. Counts will always be returned as a single bit string of length `sum(creg size for creg in circ.cregs)`. (I guess implicit in this is the assumption that the classical registers correspond to `MemorySlot`s?)
2. The memory slots for the classical registers are indexed "naturally", e.g. the bits in `cregs[0]` are indexed `0, ..., len(cregs[0])-1`, the bits in `cregs[1]` are indexed `len(cregs[0]), ..., len(cregs[0]) + len(cregs[1]) - 1`, etc.
3. The bit string is ordered with reverse indexing.

DanPuzzuoli:
### Summary

Closes #251 

### Details and comments

Also adds a test with a trivial simulation to validate that a list of classical registers gets properly counted.

### DIFF
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -965,7 +965,7 @@ def _to_schedule_list(
         elif isinstance(sched, Schedule):
             schedules.append(sched)
         elif isinstance(sched, QuantumCircuit):
-            num_memslots[-1] = sched.cregs[0].size
+            num_memslots[-1] = sum(creg.size for creg in sched.cregs)
             schedules.append(build_schedule(sched, backend, dt=backend.options.solver._dt))
         else:
             raise QiskitError(f"Type {type(sched)} cannot be converted to Schedule.")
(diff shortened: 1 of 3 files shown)


to24toro (approved):
Thanks. LGTM!

PR #260, opened 2023-09-30
Add global instances of NumPy/SciPy aliases

DanPuzzuoli:
### Summary

Adds the global instances of numpy/scipy aliases into dynamics, as well as an updated way to write tests for multiple array backends.

### Details and comments

In the end the biggest change is to the test infrastructure provided in `test/dynamics/common.py`. The existing solution we have been using for writing test classes for multiple array backends uses inheritance. This has worked pretty well, but given that are going to be making large scale changes to tests, it would be nice if this could be implemented with a single decorator, which would make it easy to add/remove array backends from test cases without having to create a bunch of test classes manually.

The new decorator `test_array_backends` in `common.py` aims to achieve this by automating the creation of the subclasses for testing different array backends. It is a little bit "hacky" as it needs to dynamically create new classes, and the implemented solution is the one way I could find to do this. One minor issue is that the subclasses generated by `test_array_backends` have methods `asarray`, etc, that the original test class needs to make use of but does not define. (Basically the original test class does not need to be a well-defined class - it can be written to rely on the functionality that gets added to it by the decorator.) A result of this is that `no-member` warnings get raised by the linter for the test class that is being decorated by `test_array_backends`. It would be nice to figure out a way to do this more "properly" to get rid of this, but I think I can also live with it.


to24toro:
@DanPuzzuoli 
Is this branch going to be merged to the main branch?

test/dynamics/common.py
```diff
@@ -129,3 +201,52 @@ def setUpClass(cls):
                 import qutip  # pylint: disable=import-outside-toplevel,unused-import
         except Exception as err:
             raise unittest.SkipTest("Skipping qutip tests.") from err
+
+
+# to be removed for 0.5.0
+class TestJaxBase(unittest.TestCase):
```
to24toro, line 207: This name baffles me with ``JaxTestBase``, but if it's temporary, I guess We'll just have to put up with it.

DanPuzzuoli: Yes sorry - I didn't want to change `TestJaxBase` as all of the existing testing infrastructure, which we need to work through and change, depends on it. As such I wanted to create a new base class for JAX tests. In the end yes, `TestJaxBase` will be removed.

DanPuzzuoli:
> @DanPuzzuoli Is this branch going to be merged to the main branch?

Yes - this is the start of making arraylias-related changes to `main`.

to24toro (changes requested):

DanPuzzuoli:
Thanks for the suggestion @to24toro - I think that makes the function a bit cleaner.

Changes:
- I changed the `lib` method in the test cases to `array_library` to be very explicit, and have made them class methods.
- I changed the `libraries` method of `test_array_backends` to `array_libraries`.
- I wrote a doc string for the file explaining the intended usage of this new infrastructure.

test/dynamics/common.py
```diff
@@ -12,17 +12,40 @@
...
+
+``QiskitDynamicsTestCase`` adds general functionality used across tests.
+
+The ``test_array_backends`` decorator, along with the classes with names ``<array_library>TestBase``
+enable writing test cases that are agnostic to the underlying array library. The general setup is:
+- Each ``<array_library>TestBase`` implements an ``array_library`` class method returning a string,
+  an ``asarray`` method for specifically defining arrays for that library, and an
+  ``assertArrayType`` method for validating that an array is from that library. These classes can
+  also implement any required setup and teardown methods for working with that library (e.g.
+  ``JAXTestBase`` skips tests if running) on windows.
```
to24toro, line 25:
```suggestion
  ``JAXTestBase`` skips tests if running on windows).
```
Is this modification correct?

to24toro (changes requested):

test/dynamics/common.py
DanPuzzuoli: Yeah that is correct, thanks!

to24toro (approved):

PR #261, opened 2023-10-03
Fix citation of JuliaQuantumControl

goerz:
As the author of the QuantumControl.jl package, I would prefer if you would cite it in the JOSS paper with the names of the people involved in the project.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -173,7 +173,8 @@
 }
 
 @misc{julia_qc,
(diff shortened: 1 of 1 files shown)


goerz:
Re: https://github.com/openjournals/joss-reviews/issues/5853

DanPuzzuoli:
Thanks for fixing this - I'm not sure what to do when citing packages if there isn't an explicit bib entry provided.

PR #269, opened 2023-10-16
Arraylias integration - Signal class -

to24toro:
### Summary
I started to integrate arraylias to dynamics.
This PR includes the integration of the Signal class in dynamics.


test/dynamics/signals/test_signals.py
```diff
@@ -728,6 +738,7 @@ def test_conjugate(self):
         )
 
 
+@partial(test_array_backends, array_libraries=["numpy", "jax", "array_numpy", "array_jax"])
```
to24toro, line 741: This decorator works as creating TestDiscreteSignalSum of "numpy", "jax", "array_numpy", "array_jax", which inherits TestSignalSum. But there is no class whose name is TestSignalSum because it has already been decorated by test_array_backends.

to24toro:
The current implementation fails the tutorial in optimizing_pulse_sequence.rst. The main reason for this failure is that `numpy.array` sometimes gets in the mix within `jit(value_and_grad(objective))`. 
- In the existing tutorial code, `dynamics.Array` is used, but for example, even if `dynamics.Array` backend is `jax`, it will be converted to `numpy.array` by `unp.asarray()`. I think this is at https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/3a373cfb834cd2f2e15245ab924aa3568b8db326/qiskit_dynamics/array/array.py#L188 .
If we are to continue supporting `dynamics.Array`  by the completion of arraylias integration, we will need to change or improve the `dynamics.Array` .

- Suppose we can change `samples` used in `DIscreteSignal` to `jax.numpy.array` through `unp.asarray`. In that case, an error will occur if there is a `numpy.array` elsewhere. Moreover, taking `Convolution` as an example, in the following code:

```python
if isinstance(signal, DiscreteSignal):
  # Perform a discrete time convolution.
  dt = signal.dt
  func_samples = np.asarray([self._func(dt * i) for i in range(signal.duration)])
  func_samples = func_samples / sum(func_samples)
  sig_samples = signal(dt * np.arange(signal.duration))
```

even if np -> unp is changed, `unp.arange(signal.duration)` becomes a `numpy.array` because `signal.duration` is assumed to be an `int`. The point is that types like `int` or `float` get coverted to `numpy` by `unp.arange()`. In the other words, `numpy` takes precedence over `jax`, leading to the issue.

qiskit_dynamics/signals/signals.py
```diff
@@ -24,13 +24,10 @@
...
 
-try:
-    import jax.numpy as jnp
-except ImportError:
-    pass
-
 from qiskit import QiskitError
-from qiskit_dynamics.array import Array
+from qiskit_dynamics.arraylias import ArrayLike
+from qiskit_dynamics.arraylias import DYNAMICS_NUMPY_ALIAS as alias
```
DanPuzzuoli, line 29: I think maybe we should call this `numpy_alias` instead of just `alias` now that there will be one for NumPy and one for SciPy.

DanPuzzuoli (changes requested):
The `signals` and `test/dynamics/signals` folders look good to me aside from the two comments I made:
- change `alias` to `numpy_alias`
- Remove all redundant type hinting for "numbers" and `List` where `ArrayLike` appears.

DanPuzzuoli:
I'm trying to figure out how to get the tests to pass - the issue with a lot of the failures seems to be that:
```
unp.asarray(Array(x, backend='jax'))
```
ends up calling `numpy.array(x)`.

This issue comes up in tests like `test.dynamics.models.test_generator_model.TestGeneratorModelSparseJax.test_jit_grad` in which a value `a` is being traced, and the function contains the code:
```
Signal(Array(a))
```
This test (and I'm guessing many others) can be fixed by simply changing the above to:
```
Signal(a)
```
This is how we'll want to do things going forward anyway, but this doesn't help with the fact that `Signal(Array(x, backend='jax'))` is broken when `x` is a tracer.

Fixing this by modifying `Array` appears to be nontrivial - it seems that `numpy` gets upset when `np.array` returns something that is not an `array`. If this is the case it may be necessary to register `Array` as it's own library, that uses `numpy` for everything _except_ `asarray`.

DanPuzzuoli:
I've figured out how to bypass the above issue and directly fix most tests [here](https://github.com/DanPuzzuoli/qiskit-dynamics/tree/signal-array-fix). The only thing I haven't fixed yet is stuff to do with pulse -> signal conversion and JAX compatibility - but I think it may just be the same issue.

I'll attempt again to see if we can "properly" fix this problem (i.e. get `unp.asarray` to work with `Array` with `backend=="jax"`). It would be nice to have backwards compatibility with code that uses `Array`, but I'm also okay with it breaking if the workaround to get things to work properly gets too complicated.

qiskit_dynamics/signals/signals.py
to24toro: fixed at https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/79c45ceacadcaba54e2bb49824a34a19b4719ec8 .

qiskit_dynamics/signals/signals.py
```diff
@@ -72,9 +69,9 @@ class Signal:
 
     def __init__(
         self,
-        envelope: Union[Callable, complex, float, int, Array],
-        carrier_freq: Union[float, List, Array] = 0.0,
-        phase: Union[float, List, Array] = 0.0,
+        envelope: Union[Callable, complex, float, int, ArrayLike],
+        carrier_freq: Union[float, List, ArrayLike] = 0.0,
+        phase: Union[float, List, ArrayLike] = 0.0,
```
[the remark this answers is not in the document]
to24toro: You are right. I fixed at https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/a3509f51a75c690bd89967820c5aeea6a96b1124 and https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/921126b104b3ce99fe2799a04f62ed40876a5210 .

to24toro:
The tests which have not been passed are two types:
- When jitting, https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/1e5fae945caac10b99e7c325e25bf38067971b9c/qiskit_dynamics/pulse/pulse_to_signals.py#L169 is the resource of errors because this `list` is converted to `numpy.array`. The solution is implementing setting something like Array.default_backend() and when it is "jax", we need to pass jnp.array([]) to `samples`.
- `np.append(ham_sig_vals, dis_sig_vals, axis=-1)` in "qiskit_dynamics/models/operator_collections.py". Should it be solved when working on operator_collections?

qiskit_dynamics/pulse/pulse_to_signals.py
```diff
@@ -226,7 +227,7 @@ def get_signals(self, schedule: Schedule) -> List[DiscreteSignal]:
             if sig.duration < max_duration:
                 sig.add_samples(
                     start_sample=sig.duration,
-                    samples=np.zeros(max_duration - sig.duration, dtype=complex),
+                    samples=unp.zeros(max_duration - sig.duration, dtype=complex),
```
DanPuzzuoli, line 230: Very minor but I think this can just remain as `np`, as `max_duration - sig.duration` should always be a standard python `int`, so this will always just be dispatched to `numpy` anyway.

qiskit_dynamics/signals/signals.py
```diff
@@ -303,38 +290,27 @@ def __init__(
         """
         self._dt = dt
 
-        samples = Array(samples)
+        samples = unp.asarray(samples)
 
         if len(samples) == 0:
-            zero_pad = np.array([0])
+            zero_pad = unp.asarray([0])
```
DanPuzzuoli, line 296: similar to the other comment - I think here it still makes sense to keep it as `np`, since this will just get dispatched to numpy.

qiskit_dynamics/signals/signals.py
```diff
@@ -303,38 +290,27 @@ def __init__(
...
         else:
-            zero_pad = np.expand_dims(np.zeros_like(Array(samples[0])), 0)
-        self._padded_samples = np.append(samples, zero_pad, axis=0)
+            zero_pad = unp.expand_dims(unp.zeros_like(unp.asarray(samples[0])), 0)
```
DanPuzzuoli, line 298: can you try changing this to:
```suggestion
            zero_pad = unp.expand_dims(unp.zeros_like(samples[0]), 0)
```
samples is already being passed through `unp.asarray` earlier in the method. The previous code had to wrap this in `Array` again as indexing would remove `Array` wrapping, but that shouldn't be necessary here.

qiskit_dynamics/signals/signals.py
```diff
@@ -445,24 +421,24 @@ def add_samples(self, start_sample: int, samples: List):
...
+        samples = unp.asarray(samples)
 
         if len(samples) < 1:
             return
 
         if start_sample < len(self.samples):
             raise QiskitError("Samples can only be added afer the last sample.")
 
-        zero_pad = np.expand_dims(np.zeros_like(Array(samples[0])), 0)
+        zero_pad = unp.expand_dims(unp.zeros_like(unp.asarray(samples[0])), 0)
```
DanPuzzuoli, line 432: Same as a previous comment, try changing this to:
```suggestion
        zero_pad = unp.expand_dims(unp.zeros_like(samples[0]), 0)
```

qiskit_dynamics/signals/signals.py
```diff
@@ -494,12 +470,10 @@ def __len__(self):
...
         return len(self.components)
 
-    def __getitem__(
-        self, idx: Union[int, List, np.array, slice]
-    ) -> Union[Signal, "SignalCollection"]:
+    def __getitem__(self, idx: Union[ArrayLike, slice]) -> Union[Signal, "SignalCollection"]:
         """Get item with NumPy-style subscripting, as if this class were a 1d array."""
 
-        if type(idx) == np.ndarray and idx.ndim > 0:
+        if type(idx) != int and type(idx) != slice and type(idx) != list and idx.ndim > 0:
```
DanPuzzuoli, line 476: Hmm - the only issue with this current version is that when `idx.ndim > 0` is checked there is no guarantee that `idx` has an `ndim` attribute. I wonder if you can do:
```suggestion
        if type(idx) != slice and unp.asarray(idx).ndim > 0:
```
this will cover the `int` case, and we'll be calling `list` on a `list` for the `list` case, so that's fine too.

qiskit_dynamics/signals/signals.py
```diff
@@ -573,7 +547,7 @@ def __init__(self, *signals, name: Optional[str] = None):
             elif isinstance(sig, Signal):
                 components.append(sig)
             elif isinstance(sig, (int, float, complex)) or (
-                isinstance(sig, Array) and sig.ndim == 0
+                not isinstance(sig, (int, float, complex)) and sig.ndim == 0
```
DanPuzzuoli, line 550: Maybe this whole `elif` can be changed to:
```
else:
    try:
        if unp.asarray(sig).ndim == 0:
            components.append(Signal(sig))
    except:
```

This way the final resort is to check if it can be converted to an array or not. If not, the error at the end will be raised. Perhaps the error message at the end should also be changed to:
```
"Components of a SignalSum must be instances of a Signal subclass or a scalar."
```

qiskit_dynamics/signals/signals.py
```diff
@@ -607,12 +572,12 @@ def envelope(t):
...
 
-    def complex_value(self, t: Union[float, np.array, Array]) -> Union[complex, np.array, Array]:
+    def complex_value(self, t: ArrayLike) -> ArrayLike:
         """Return the sum of the complex values of each component."""
-        if Array.default_backend() == "jax":
-            t = Array(t)
-        exp_phases = np.exp(np.expand_dims(t, -1) * self._carrier_arg + self._phase_arg)
-        return np.sum(self.envelope(t) * exp_phases, axis=-1)
+        exp_phases = unp.exp(
+            unp.expand_dims(unp.asarray(t), -1) * self._carrier_arg + self._phase_arg
```
DanPuzzuoli, line 578: Is `unp.asarray(t)` necessary here, since whatever `t` is, it will just get dispatched by `unp.expand_dims` anyway? Or will this not work if `t` is a list or something.

qiskit_dynamics/signals/signals.py
```diff
@@ -637,12 +602,14 @@ def flatten(self) -> Signal:
...
 
-        ave_freq = np.sum(self.carrier_freq) / len(self)
+        ave_freq = unp.sum(self.carrier_freq) / len(self)
         shifted_arg = self._carrier_arg - (1j * 2 * np.pi * ave_freq)
 
         def merged_env(t):
-            exp_phases = np.exp(np.expand_dims(Array(t), -1) * shifted_arg + self._phase_arg)
-            return np.sum(self.envelope(t) * exp_phases, axis=-1)
+            exp_phases = unp.exp(
+                unp.expand_dims(unp.asarray(t), -1) * shifted_arg + self._phase_arg
```
DanPuzzuoli, line 610: Similarly here, is `unp.asarray(t)` necessary?

qiskit_dynamics/signals/signals.py
```diff
@@ -675,12 +642,12 @@ def __init__(
             name: name of the signal.
         """
 
-        samples = Array(samples)
+        samples = unp.asarray(samples)
         if carrier_freq is None:
-            carrier_freq = np.zeros(samples.shape[-1], dtype=float)
+            carrier_freq = unp.zeros(samples.shape[-1], dtype=float)
```
DanPuzzuoli, line 647: This can stay as `np` as `samples.shape[-1]` will always be a standard python integer.
```suggestion
            carrier_freq = np.zeros(samples.shape[-1], dtype=float)
```

qiskit_dynamics/signals/signals.py
```diff
@@ -675,12 +642,12 @@ def __init__(
...
 
         if phase is None:
-            phase = np.zeros(samples.shape[-1], dtype=float)
+            phase = unp.zeros(samples.shape[-1], dtype=float)
```
DanPuzzuoli, line 650:
```suggestion
            phase = np.zeros(samples.shape[-1], dtype=float)
```

qiskit_dynamics/signals/signals.py
```diff
@@ -750,7 +717,7 @@ def from_SignalSum(
 
         if sample_carrier:
             freq = 0.0 * freq
-            exp_phases = np.exp(np.expand_dims(Array(times), -1) * signal_sum._carrier_arg)
+            exp_phases = unp.exp(unp.expand_dims(unp.asarray(times), -1) * signal_sum._carrier_arg)
```
DanPuzzuoli, line 720: Is `unp.asarray(times)` necessary?

qiskit_dynamics/signals/signals.py
```diff
@@ -895,9 +858,9 @@ def signal_add(sig1: Signal, sig2: Signal) -> SignalSum:
             and sig1.start_time == sig2.start_time
             and sig1.duration == sig2.duration
         ):
-            samples = np.append(sig1.samples, sig2.samples, axis=1)
-            carrier_freq = np.append(sig1.carrier_freq, sig2.carrier_freq)
-            phase = np.append(sig1.phase, sig2.phase)
+            samples = unp.append(sig1.samples, sig2.samples, axis=1)
+            carrier_freq = unp.append(sig1.carrier_freq, sig2.carrier_freq)
+            phase = unp.append(sig1.phase, sig2.phase)
```
DanPuzzuoli, line 863: I suspect these lines may be prone to the dispatching errors we had to deal with in other places. Should these be changed to use `_numpy_multi_dispatch`?

qiskit_dynamics/signals/signals.py
```diff
@@ -951,29 +914,29 @@ def signal_multiply(sig1: Signal, sig2: Signal) -> SignalSum:
...
-            new_samples_conj = Array(
+            new_samples_conj = unp.asarray(
                 0.5
                 * (sig1.samples[:, :, None] * sig2.samples[:, None, :].conj()).reshape(
                     (sig1.samples.shape[0], sig1.samples.shape[1] * sig2.samples.shape[1]),
                     order="C",
                 )
             )
-            samples = np.append(new_samples, new_samples_conj, axis=1)
+            samples = unp.append(new_samples, new_samples_conj, axis=1)
```
DanPuzzuoli, line 931: `_numpy_multi_dispatch`?

qiskit_dynamics/signals/signals.py
```diff
@@ -951,29 +914,29 @@ def signal_multiply(sig1: Signal, sig2: Signal) -> SignalSum:
...
 
             new_freqs = sig1.carrier_freq + sig2.carrier_freq
             new_freqs_conj = sig1.carrier_freq - sig2.carrier_freq
-            freqs = np.append(Array(new_freqs), Array(new_freqs_conj))
+            freqs = unp.append(unp.asarray(new_freqs), unp.asarray(new_freqs_conj))
```
DanPuzzuoli, line 935: `_numpy_multi_dispatch`?

qiskit_dynamics/signals/signals.py
```diff
@@ -951,29 +914,29 @@ def signal_multiply(sig1: Signal, sig2: Signal) -> SignalSum:
...
 
             new_phases = sig1.phase + sig2.phase
             new_phases_conj = sig1.phase - sig2.phase
-            phases = np.append(Array(new_phases), Array(new_phases_conj))
+            phases = unp.append(unp.asarray(new_phases), unp.asarray(new_phases_conj))
```
DanPuzzuoli, line 939: `_numpy_multi_dispatch`?

DanPuzzuoli (changes requested):
All looking good to me, but some final nitpicking details to sort out.

qiskit_dynamics/signals/signals.py
to24toro: I fixed at https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/2e1436e838dec1acbb5dd478b39a773d353741b4.

qiskit_dynamics/signals/signals.py
to24toro: I fixed at https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/2e1436e838dec1acbb5dd478b39a773d353741b4.

qiskit_dynamics/signals/signals.py
to24toro: I forgot to change it to `_numpy_multi_dispatch`.
I fixed at https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/2e1436e838dec1acbb5dd478b39a773d353741b4.

qiskit_dynamics/signals/signals.py
to24toro: I fixed at https://github.com/Qiskit-Extensions/qiskit-dynamics/pull/269/commits/2e1436e838dec1acbb5dd478b39a773d353741b4.

qiskit_dynamics/signals/transfer_functions.py
```diff
@@ -117,11 +118,13 @@ def _apply(self, signal: Signal) -> Signal:
...
-            func_samples = func_samples / sum(func_samples)
-            sig_samples = signal(dt * np.arange(signal.duration))
+            func_samples = unp.asarray([self._func(dt * i) for i in range(signal.duration)])
+            func_samples = func_samples / unp.sum(func_samples)
+            sig_samples = signal(dt * unp.arange(signal.duration))
 
-            convoluted_samples = list(np.convolve(func_samples, sig_samples))
+            convoluted_samples = _numpy_multi_dispatch(
+                func_samples, sig_samples, path="convolve"
+            )  # unp.convolve(func_samples, sig_samples)
```
DanPuzzuoli, line 127:
```suggestion
            )
```

DanPuzzuoli (changes requested):

DanPuzzuoli (approved):

DanPuzzuoli (approved):

PR #272, opened 2023-10-17
Change test_array_backends mixin classes to subclass QiskitDynamicsTestCase

DanPuzzuoli:
### Summary

Changes the classes utilized by `test_array_backends` to be subclasses of `QiskitDynamicsTestCase`. The `arraylias` module tests have also been updated (this is the only test folder that currently uses this).


to24toro (approved):

to24toro:
A modification where `Arraytest` inherits `QiskitDynamicsTestCase`  caused a `no-member` error in the existing tests that inherits `QiskitDynamicsTestCase` since the method of `QiskitDynamicsTestCase` is not found in the tests.(e.g. self.assertAllClose()). At least, I believe we have to acknowledge this error.

It might be good to provide a means of Arraylais to make tests for some arrays work.

PR #281, opened 2023-10-30
Updating solvers module to use arraylias

DanPuzzuoli:
### Summary

Files/sections:
- ~scipy solver~
- ~diffrax~
- ~fixed_step~
- ~jax_odeint~
- ~lanczos~
- solver_functions
- solver_classes
- ~solver_utils~
- perturbative solvers
    - It may make sense to group this with the perturbation module

### Details and comments

I'm currently stuck on `solver_functions.py` as it uses `RotatingFrame`, and the tests use `GeneratorModel`. It may be necessary to update the `models` module before this can be completed.


qiskit_dynamics/solvers/solver_classes.py
```diff
@@ -758,20 +773,16 @@ def validate_and_format_initial_state(y0: any, model: Union[HamiltonianModel, Li
...
     # if Simulating density matrix or SuperOp with a HamiltonianModel, simulate the unitary
     if y0_cls in [DensityMatrix, SuperOp] and isinstance(model, HamiltonianModel):
         y0 = np.eye(model.dim, dtype=complex)
     # if LindbladModel is vectorized and simulating a density matrix, flatten
-    elif (
-        (y0_cls is DensityMatrix)
-        and isinstance(model, LindbladModel)
-        and "vectorized" in model.evaluation_mode
-    ):
+    elif (y0_cls is DensityMatrix) and isinstance(model, LindbladModel) and model.vectorized:
```
to24toro, line 785:
```suggestion
    elif (y0_cls is DensityMatrix) and is_lindblad_model_vectorized(model):
```

qiskit_dynamics/solvers/solver_utils.py
```diff
@@ -35,17 +35,15 @@
...
 def is_lindblad_model_vectorized(obj: any) -> bool:
     """Return True if obj is a vectorized LindbladModel."""
-    return isinstance(obj, LindbladModel) and ("vectorized" in obj.evaluation_mode)
+    return isinstance(obj, LindbladModel) and obj.vectorized
 
 
 def is_lindblad_model_not_vectorized(obj: any) -> bool:
     """Return True if obj is a non-vectorized LindbladModel."""
-    return isinstance(obj, LindbladModel) and ("vectorized" not in obj.evaluation_mode)
+    return isinstance(obj, LindbladModel) and not obj.vectorized
```
to24toro, line 43:
```suggestion
    return not is_lindblad_model_vectorized(obj)
```

test/dynamics/solvers/test_solver_classes.py
```diff
@@ -746,16 +738,48 @@ def func(a):
...
+                method="jax_odeint",
+            ).y[-1]
+            return yf
+
+        self.jit_grad(func)(1.0)
+
+
+@partial(test_array_backends, array_libraries=["jax"])
+class TestSolverConstructionJAXTransformations:
+    """Test construction of Solver within a function to be JAX transformed.."""
```
to24toro, line 750:
```suggestion
    """Test construction of Solver within a function to be JAX transformed."""
```

test/dynamics/solvers/test_solver_functions_interface.py
```diff
@@ -180,14 +180,43 @@ def setUp(self):
...
 
-        self.vec_lindblad_model = self.lindblad_model.copy()
-        self.vec_lindblad_model.evaluation_mode = "dense_vectorized"
+        self.vec_lindblad_model = LindbladModel(
+            hamiltonian_operators=[Operator.from_label("X")],
+            hamiltonian_signals=[Signal(1.0, 5.0)],
+            static_hamiltonian=Operator.from_label("Z"),
+            static_dissipators=[Operator.from_label("Y")],
+            vectorized=True,
+        )
```
to24toro, line 189: It reflects on the discussion that the model class is immutable.

to24toro (changes requested):
Thank you, @DanPuzzuoli .
I separated my PR comments from `perturbation` updatePR.
Plese see if when you have time.

to24toro:
It reflects on the discussion that the model class is immutable.

to24toro:
https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/88b2dfd357005f5241d0c059a814360059a91fae/qiskit_dynamics/models/__init__.py#L107

Also, need to fix it?

qiskit_dynamics/solvers/solver_utils.py
DanPuzzuoli: These are not actually the same. If `obj` is a `HamiltonianModel`, then 
```
(isinstance(obj, LindbladModel) and not obj.vectorized) == False
```
but
```
not is_lindblad_model_vectorized(obj) == True
```

`obj` needs to be a `LindbladModel` for both of these functions to return `True`.

DanPuzzuoli:
> https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/88b2dfd357005f5241d0c059a814360059a91fae/qiskit_dynamics/models/__init__.py#L107
> 
> Also, need to fix it?

yes good catch - for now I don't want to fix documentation, I think it should all be fixed at the end when all of the changes have been made (to avoid having to make further unexpected changes).

test/dynamics/solvers/test_solver_functions_interface.py
to24toro: No, just comment.

qiskit_dynamics/solvers/solver_utils.py
to24toro: Certainly, you're right.

to24toro (approved):
Sorry for late approval.

PR #284, opened 2023-11-02
Added zeros and ones funcs to common

to24toro:
### Summary
Added funcs like np.ones and np.zeros to Array test classes. At least, it will be used for tests in RWA.


to24toro:
I feel like it's unnecessary, so let's close it.

PR #286, opened 2023-11-07
Add sparse modules to update dynamics by arraylias

to24toro:
### Summary
I added some modules of jax_sparse and numpy_sparse. I wrote the code, looking at the model class and don't know if it is suite for other classes.
In order to make Operator input work, I think it is necessary to add Operator as a new type to dynamics arraylias. Maybe, we need to add list too, but I cannot conclude the necessarily of the list now because it is on going about integrating the model.


DanPuzzuoli:
> In order to make Operator input work, I think it is necessary to add Operator as a new type to dynamics arraylias.

Hmm possibly. I do know though that `Operator` implements `__array__`, so:
```
unp.asarray(Operator.from_label('X'))
```
works, and even
```
unp.asarray([Operator.from_label('X'), Operator.from_label('Y')])
```
already works as is. Maybe this is enough?

qiskit_dynamics/arraylias/register_functions/to_sparse.py
```diff
@@ -0,0 +1,61 @@
...
+Registering asarray functions to alias
+"""
+
+import numpy as np
+from scipy.sparse import csr_matrix
+from qiskit_dynamics.type_utils import isinstance_qutip_qobj
+
+
+def register_tosparse(alias):
+    """register to_sparse functions to each array libraries"""
```
DanPuzzuoli, line 25:
```suggestion
    """Register to_sparse functions to each array library."""
```

qiskit_dynamics/arraylias/register_functions/to_dense.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+Registering to_dense functions to alias
+"""
+
+import numpy as np
+from arraylias.exceptions import LibraryError
+
+
+def register_todense(alias):
+    """register to_dense functions to each array libraries"""
```
DanPuzzuoli, line 24:
```suggestion
    """Register to_dense functions to each array library."""
```

qiskit_dynamics/arraylias/register_functions/rmatmul.py
```diff
@@ -0,0 +1,49 @@
...
+
+"""
+Registering rmatmul functions to alias
+"""
+
+import numpy as np
+
+
+def register_rmatmul(alias):
+    """register rmatmul functions to each array libraries"""
```
DanPuzzuoli, line 24:
```suggestion
    """Register rmatmul functions to each array library."""
```

qiskit_dynamics/arraylias/register_functions/multiply.py
```diff
@@ -0,0 +1,42 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Registering multiply functions to alias
```
DanPuzzuoli, line 16:
```suggestion
Register multiply functions to alias.
```

qiskit_dynamics/arraylias/register_functions/multiply.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+
+
+def register_multiply(alias):
+    """register multiply functions to each array libraries"""
```
DanPuzzuoli, line 21:
```suggestion
    """Register multiply functions to each array library."""
```

qiskit_dynamics/arraylias/register_functions/matmul.py
```diff
@@ -0,0 +1,38 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Registering matmul functions to alias
```
DanPuzzuoli, line 16:
```suggestion
Register matmul functions to alias.
```

qiskit_dynamics/arraylias/register_functions/matmul.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+
+
+def register_matmul(alias):
+    """register matmul functions to each array libraries"""
```
DanPuzzuoli, line 21:
```suggestion
    """Register matmul functions to required array libraries."""
```

DanPuzzuoli (changes requested):
Overall looks good - excited for these changes.

Comments:
- I made a bunch of suggestions - just making sure the file and function doc strings are full sentences.
- Can we get rid of `to_numeric_matrix_type` by registering a suitable default for `asarray` instead?
- Do we need `rmatmul` anymore? I think we had set this up to originally so we could dispatch on the second argument of `matmul`, but we may just be able to use the `_numpy_multi_dispatch` function with `matmul` instead.

Beyond this, I guess an issue with scipy sparse is that we store lists of them in `numpy` arrays, e.g. 
```
np.array([csr_matrix(...), csr_matrix(...)])
```
so the library inference doesn't work. One solution to this could be to define a new class with a name like `SparseMatrixList` that just stores the numpy object array of sparse matrices, is registered as a `scipy_sparse` type, and passes all ufuncs/numpy function calls through to the underlying numpy object array.

to24toro:
Thanks. It's no need to add "Operator"(https://github.com/Qiskit/qiskit/blob/af643eb09b22fc3833084b3112890b1aebdcbf92/qiskit/quantum_info/operators/operator.py#L120)

qiskit_dynamics/arraylias/register_functions/to_numeric_matrix_type.py
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+Registering to_numeric_matrix_type functions to alias
+"""
+
+from arraylias.exceptions import LibraryError
+from qiskit_dynamics.type_utils import isinstance_qutip_qobj
+
+
+def register_to_numeric_matrix_type(alias):
```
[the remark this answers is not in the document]
to24toro: Yes. I'm just considering it, but I cannot decide it because it is used in model files I have not seen yet.

to24toro:
> Do we need rmatmul anymore? I think we had set this up to originally so we could dispatch on the second argument of matmul, but we may just be able to use the _numpy_multi_dispatch function with matmul instead.

Yes, it makes sense but I don't think we can use _numpy_multi_dispatch which only judges whether the array type is "numpy" or "jax" not "sparse.

qiskit_dynamics/arraylias/register_functions/to_numeric_matrix_type.py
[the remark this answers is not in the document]
to24toro: Certainly, I also feel we can integrate it to asarray.

to24toro:
> Beyond this, I guess an issue with scipy sparse is that we store lists of them in numpy arrays, e.g.
np.array([csr_matrix(...), csr_matrix(...)])
so the library inference doesn't work. One solution to this could be to define a new class with a name like SparseMatrixList that just stores the numpy object array of sparse matrices, is registered as a scipy_sparse type, and passes all ufuncs/numpy function calls through to the underlying numpy object array.


I think the reason we want to add `list` type to alias is to treat such as List[spase_matrix]. Therefore, it is a good idea.

qiskit_dynamics/arraylias/register_functions/asarray.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+from scipy.sparse import csr_matrix, issparse
+
+
+def register_asarray(alias):
+    """register asarray functions to each array libraries"""
+
+    @alias.register_default(path="asarray")
+    def _(arr):
+        return arr
```
DanPuzzuoli, line 28:
```suggestion
        return np.asarray(arr)
```

The default should be to return a numpy array. This fixes one of the test errors as well as the error in the docs build. Unfortunately the docs build still fails due to warnings of casting complex -> real, but that's a separate issue.

qiskit_dynamics/arraylias/register_functions/to_dense.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def register_todense(alias):
+    """Register to_dense functions to each array library."""
+
+    @alias.register_default(path="to_dense")
+    def _(arr):
+        if arr is None:
+            return None
+        return arr
```
DanPuzzuoli, line 30: I'm wondering if this should also be
```suggestion
        return np.asarray(arr)
```

We expect the output of `to_dense` to either be a numpy or JAX array, so currently with `return arr` here if a user enters an unidentified type it will just be returned.

qiskit_dynamics/arraylias/register_functions/to_numeric_matrix_type.py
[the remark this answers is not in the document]
DanPuzzuoli: Perhaps for now just remove it and we'll see if it can be replaced effectively with `asarray` in the model files.

DanPuzzuoli (changes requested):
Aside from the direct comments, another change I would suggest is to put the `register_fallback` and `register_default` functions together in the files (currently the `default` functions appear at the beginning, and the `fallback` functions appear at the end).

I believe these suggestions should fix the current issues. There is another problem with tests raising warnings, but I think these are the result of a JAX update possibly. I will see if I can resolve these in a separate PR.

DanPuzzuoli (changes requested):
This LGTM. The only thing missing is some tests - can you add a few test cases for the new types/functions?

qiskit_dynamics/arraylias/register_functions/asarray.py
```diff
@@ -27,10 +27,6 @@ def register_asarray(alias):
     def _(arr):
         return np.asarray(arr)
 
-    @alias.register_fallback(path="asarray")
-    def _(arr):
-        return np.asarray(arr)
-
```
to24toro: I feel we can remove this fallback functions.

DanPuzzuoli: Yes this seems okay since we have registered an `asarray` for every library.

test/dynamics/arraylias/register_functions/test_matmul.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class TestMultiplyFunction(QiskitDynamicsTestCase):
+    """Test cases for matmul functions registered in dynamics_numpy_alias."""
+
+    def test_scipy_sparse(self):
+        """Test matmul for scipy_sparse."""
+        x = csr_matrix([[1, 0], [0, 1]])
+        y = csr_matrix([[2, 2], [2, 2]])
+        self.assertAllClose(csr_matrix.toarray(unp.matmul(x, y)), [[2, 2], [2, 2]])
```
DanPuzzuoli, line 33: validate type of `unp.matmul(x, y)`

test/dynamics/arraylias/register_functions/test_matmul.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_jax_sparse(self):
+        """Test matmul for jax_sparse."""
+        try:
+            from jax.experimental.sparse import BCOO
+
+            x = BCOO.fromdense([[1, 0], [0, 1]])
+            y = BCOO.fromdense([[2, 2], [2, 2]])
+            self.assertAllClose(BCOO.todense(unp.matmul(x, y)), [[2, 2], [2, 2]])
```
DanPuzzuoli, line 42: same thing here, validate type of `unp.matmul(x, y)`

test/dynamics/arraylias/register_functions/test_matmul.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+import unittest
+import numpy as np
+from scipy.sparse import csr_matrix
+
+from qiskit_dynamics import DYNAMICS_NUMPY as unp
+from ...common import QiskitDynamicsTestCase
+
+
+class TestMultiplyFunction(QiskitDynamicsTestCase):
```
DanPuzzuoli, line 26: `TestMatmulFunction`

test/dynamics/arraylias/register_functions/test_rmatmul.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+import unittest
+import numpy as np
+from scipy.sparse import csr_matrix
+
+from qiskit_dynamics import DYNAMICS_NUMPY as unp
+from ...common import QiskitDynamicsTestCase
+
+
+class TestMultiplyFunction(QiskitDynamicsTestCase):
```
DanPuzzuoli, line 26: `TestRmatmulFunction`

DanPuzzuoli (changes requested):
Thanks Kento - the tests look good. I've left a few comments.

The main thing is to add validation for the output type of the dispatched functions. I've commented in a few places, but I think this should be done everywhere.

DanPuzzuoli (approved):

PR #292, opened 2023-12-06
Implement DynamicsBackend.solve function and unit test

### LINKED ISSUE #288: Add explicit interface to DynamicsBackend for just returning results of ODE simulation
Currently `DynamicsBackend.run`, is the only method of `DynamicsBackend` for running simulations. In accordance with the standard backend interfaces, it performs a state simulation and returns a probability dict.

It would be good to have a public interface just for ODE simulation with arbitrary "states" (including unitary simulation) - basically the equivalent of `Solver.solve`, but in which a user can take advantage of circuit transpilation. I think it probably makes sense to add a new `DynamicsBackend.solve` method for this. (The other option would be to add options to `DynamicsBackend.run`, but this would completely change the output type, and therefore complicate the interface, and also take it out of line with general `backend.run`.)

donsano33:
### Summary

Add public function `DynamicsBackend.solve` enabling ODE simulation with the preconfigured solver member of the `DynamicsBackend`.

### Details and comments

I opted for the most barebone implementation, basically just mirroring the interface for the `Solver.solve` function with an additional `validate` parameter.

Depending on how sophisticated one wants to implement this function, a variety of extensions could be thought of, adding ease of usage but also complexity.
One possible extension could be (similar to `DynamicsBackend.run`) to optionally use the placement of the Acquire pulses to dynamically figure out the t_span (if measurements are in the schedule / circuit).

Let me know if you want something of that kind to be added.

Closes #288

### DIFF
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -345,17 +345,25 @@ class DynamicsBackend(BackendV2):
     def solve(
         self,
         t_span: Array,
-        y0: Union[Array, QuantumState, BaseOperator],
+        y0: Optional[Union[Array, QuantumState, BaseOperator]] = None,
         solve_input: Optional[
             Union[
                 List[QuantumCircuit],
                 List[Union[Schedule, ScheduleBlock]],
             ]
         ] = None,
-        convert_results: bool = True,
-        validate: bool = True,
+        convert_results: Optional[bool] = True,
+        validate: Optional[bool] = True,
     ) -> Union[OdeResult, List[OdeResult]]:
-        """Solve a list of simulations
+        """Simulate a list of :class:`~qiskit.circuit.QuantumCircuit`,
+           :class:`~qiskit.pulse.Schedule`, or :class:`~qiskit.pulse.ScheduleBlock` instances and
+           return the ``OdeResult``.
+
+        This method is analogous to :meth:`.Solver.solve`, however it additionally utilizes
+        transpilation and the backend configuration to convert
+        :class:`~qiskit.circuit.QuantumCircuit` instances into pulse-level schedules for simulation.
+        The options for the solver will be drawn from ``self.options.solver_options``, and if
+        ``y0`` is not specified, it will be set from ``self.options.initial_state``.
 
         Args:
             t_span: Time interval to integrate over.
@@ -373,6 +381,13 @@ class DynamicsBackend(BackendV2):
         if validate:
             _validate_run_input(solve_input)
         schedules, _ = _to_schedule_list(solve_input, backend=self)
+
+        # use default y0 if not given as parameter
+        if y0 is None:
+            y0 = self.options.initial_state
+        if y0 == "ground_state":
+            y0 = Statevector(self._dressed_states[:, 0])
+
         solver_results = self.options.solver.solve(
             t_span=t_span,
             y0=y0,
@@ -405,7 +420,7 @@ class DynamicsBackend(BackendV2):
         """
 
         if validate:
-            _validate_run_input(run_input, accepted_types=(QuantumCircuit, Schedule, ScheduleBlock))
+            _validate_run_input(run_input)
 
         # Configure run options for simulation
         if options:
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -51,7 +51,6 @@ from qiskit.quantum_info.states.quantum_state import QuantumState
 from qiskit_dynamics import RotatingFrame
 from qiskit_dynamics.array import Array
 from qiskit_dynamics.solvers.solver_classes import Solver
-from qiskit_dynamics.signals import Signal
 
 from .dynamics_job import DynamicsJob
 from .backend_utils import (
@@ -372,9 +371,7 @@ class DynamicsBackend(BackendV2):
             OdeResult: object with formatted output types.
         """
         if validate:
-            _validate_run_input(
-                solve_input, accepted_types=(QuantumCircuit, Schedule, ScheduleBlock)
-            )
+            _validate_run_input(solve_input)
         schedules, _ = _to_schedule_list(solve_input, backend=self)
         solver_results = self.options.solver.solve(
             t_span=t_span,
@@ -918,15 +915,15 @@ def default_experiment_result_function(
         raise QiskitError(f"meas_level=={backend.options.meas_level} not implemented.")
 
 
-def _validate_run_input(run_input, accepted_types, accept_list=True):
+def _validate_run_input(run_input, accept_list=True):
     """Raise errors if the run_input is not one of QuantumCircuit, Schedule, ScheduleBlock, or
     a list of these.
     """
     if isinstance(run_input, list) and accept_list:
         # if list apply recursively, but no longer accept lists
         for x in run_input:
-            _validate_run_input(x, accepted_types, accept_list=False)
-    elif not isinstance(run_input, accepted_types):
+            _validate_run_input(x, accept_list=False)
+    elif not isinstance(run_input, (QuantumCircuit, Schedule, ScheduleBlock)):
         caller_func_name = inspect.stack()[1].function
         raise QiskitError(
             f"Input type {type(run_input)} not supported by DynamicsBackend.{caller_func_name}."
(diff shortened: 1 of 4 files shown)


DanPuzzuoli (changes requested):
Hi @donsano33 , thank you for the contribution! It will be nice to have an easy interface for getting ODE state solutions from the `DynamicsBackend` utilizing transpilation.

One change I think should be made is that `DynamicsBackend.solve` should actually not accept `Signal` inputs. If a user wants to pass in `Signal` inputs they can directly access the `Solver` via `backend.options.solver` and pass them through the existing `Solver.solve` interface. This will keep the interface for `DynamicsBackend` solely about taking in inputs that real backends support (schedules and circuits).

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -871,16 +929,19 @@ def default_experiment_result_function(
         raise QiskitError(f"meas_level=={backend.options.meas_level} not implemented.")
 
 
-def _validate_run_input(run_input, accept_list=True):
+def _validate_run_input(run_input, accepted_types, accept_list=True):
```
donsano33, line 932: @DanPuzzuoli  As I am removing the ability to input signals in the `DynamicsBackend.solve` function, should I remove this change as well?
It makes the `validate_run_input` function more versatile but now lacks a usecase as the list of allowed types is the same for all functions calling it.

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -338,6 +343,59 @@ def _set_solver(self, solver):
...
+                List[QuantumCircuit],
+                List[Union[Schedule, ScheduleBlock]],
+                List[Signal],
+                Tuple[List[Signal], List[Signal]],
+            ]
+        ] = None,
+        convert_results: bool = True,
+        validate: bool = True,
+    ) -> Union[OdeResult, List[OdeResult]]:
+        """Solve a list of simulations
```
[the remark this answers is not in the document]
donsano33: What do you think of:
"Solve a list of schedules/circuits analogous to the :meth: `.Solver.solve` interface."

qiskit_dynamics/backend/dynamics_backend.py
DanPuzzuoli: Yeah if it isn't being used you can remove this change.

qiskit_dynamics/backend/dynamics_backend.py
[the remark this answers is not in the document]
donsano33: I added the suggested docstring (very many words ;)) and implemented `y0` as optional, opting to `self.options.initial_state`.
This is set to `self._dressed_states` if `y0 == "ground_state`.

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -871,16 +930,20 @@ def default_experiment_result_function(
         raise QiskitError(f"meas_level=={backend.options.meas_level} not implemented.")
 
 
-def _validate_run_input(run_input, accept_list=True):
+def _validate_run_input(run_input, accept_list=True, caller_func_name: str = None):
```
DanPuzzuoli, line 933: Nice

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -338,6 +342,61 @@ def _set_solver(self, solver):
...
+    def solve(
+        self,
+        t_span: Array,
+        y0: Optional[Union[Array, QuantumState, BaseOperator]] = None,
+        solve_input: Optional[
+            Union[
+                List[QuantumCircuit],
+                List[Union[Schedule, ScheduleBlock]],
+            ]
+        ] = None,
```
DanPuzzuoli, line 354:
```suggestion
        solve_input: List[Union[QuantumCircuit, Schedule, ScheduleBlock]],
```
This is the form of the type-hint for `DynamicsBackend.run`. I don't think we should allow passing `None` or have a default value here. This may require moving this above `y0` in the argument list. I think it'd be okay to move it to the first argument.

qiskit_dynamics/backend/dynamics_backend.py
```diff
@@ -338,6 +342,61 @@ def _set_solver(self, solver):
...
+        convert_results: Optional[bool] = True,
+        validate: Optional[bool] = True,
+    ) -> Union[OdeResult, List[OdeResult]]:
+        """Simulate a list of :class:`~qiskit.circuit.QuantumCircuit`,
+           :class:`~qiskit.pulse.Schedule`, or :class:`~qiskit.pulse.ScheduleBlock` instances and
+           return the ``OdeResult``.
```
DanPuzzuoli, line 360: I'm surprised the linter didn't complain about this.

```suggestion
        """Simulate a list of :class:`~qiskit.circuit.QuantumCircuit`,
        :class:`~qiskit.pulse.Schedule`, or :class:`~qiskit.pulse.ScheduleBlock` instances and
        return the ``OdeResult``.
```

releasenotes/notes/dynamics-backend-solve-fd42dac884ff80cc.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Adds the :meth:`.DynamicsBackend.solve` function for ODE simulations with arbitrary input types
+    `y0` (`Statevector`, `Operator`, ..), corresponding to the :meth:`.Solver.solve` function of the
+    backends preconfigured :class:`.Solver` member.
```
DanPuzzuoli, line 6:
```suggestion
    Adds the :meth:`.DynamicsBackend.solve` method for running simulations of circuits and schedules for arbitrary input types, and returning the ODE simulation results.
```

DanPuzzuoli (changes requested):
Over all looks good - just some final nitpicks of the wording/arguments.

The only major comment is that it'd be good for the test to actually validate the output of the simulation. I'm not sure if the schedule/circuit represents an X pulse, but can you add validation of the expected final states? There is a method in the base test class called `assertAllClose(x, y)` that does `self.assertTrue(np.allclose(x, y))` (and a bit more than this).

qiskit_dynamics/backend/dynamics_backend.py
donsano33: I agree, I put it as first argument.

qiskit_dynamics/backend/dynamics_backend.py
donsano33: Done

test/dynamics/backend/test_dynamics_backend.py
```diff
@@ -301,6 +302,42 @@ def test_pi_pulse(self):
...
+        input_variety = [x_sched0, x_circ0]
+
+        # solve for all combinations of input types and initial states
+        for solve_input, (y0, expected_result) in product(input_variety, y0_and_expected_results):
+            solver_results = self.simple_backend.solve(
+                t_span=[0, n_samples * self.simple_backend.dt], y0=y0, solve_input=[solve_input]
+            )
+            if isinstance(solver_results, list):
+                for solver_result in solver_results:
+                    assert solver_result.success
```
DanPuzzuoli, line 335:
```suggestion
                    self.assertTrue(solver_result.success)
```

test/dynamics/backend/test_dynamics_backend.py
```diff
@@ -301,6 +302,42 @@ def test_pi_pulse(self):
...
+                    self.assertAllClose(solver_result.y[-1], expected_result, atol=1e-2)
+            else:
+                assert solver_results.success
```
DanPuzzuoli, line 338:
```suggestion
                self.assertTrue(solver_results.success)
```

test/dynamics/backend/test_dynamics_backend.py
DanPuzzuoli: Not sure if `assertTrue` is correct here, either use this if `solver_result.success` is a bool or use `assertEqual` if it's some other type.

DanPuzzuoli (changes requested):
Looks good. The only final thing is that the tolerance of `1e-2` is very loose. I guess the trade off here is that you are running pulse of a decent length, but you don't want to decrease the solver tolerance or it will slow the simulation down quite a bit.

One alternative is to define the gate to be quite short, adjust the expected unitary accordingly, set the solver to use a lower tolerance, then lower the test case tolerance as well. E.g. you could reduce the number of samples to `5`, and set the expected unitary to be a rotation about X by `pi * 5/100`. I think I'd prefer this over a longer pulse but using a loose tolerance on the test. Otherwise looks good to me.

DanPuzzuoli (approved):

PR #293, opened 2023-12-07
Depend on qiskit directly

1ucian0:
### Summary

Currently, `qiskit-dynamics` depends on `qiskit-terra`. Starting in Qiskit 1.0.0 only the `qiskit` package will be published. This PR changes the requirement to use `qiskit` instead of `qiskit-terra`.

### Details and comments

https://github.com/Qiskit/qiskit/pull/11230


DanPuzzuoli (approved):
Thanks @1ucian0 !

PR #248, opened 2023-07-18
Remove subsystem_labels option from DynamicsBackend

DanPuzzuoli:
### Summary

Closes #235

### Details and comments (out of date, see next comment)

This is currently a draft to prototype a potential solution to #235, in which the `subsystem_labels` option is removed, and all qubits removed via the `subsystem_list` argument of `from_backend` are actually "kept" as trivial 1-dimensional systems.

So far:
- I've commented out all lines/blocks using `subsystem_labels`.
- Where necessary, new code has been added, e.g. construction of `subsystem_dims` in `from_backend` has been changed to include all pre-existing subsystems, but setting the dimension of the "removed" subsystems to `1` (instead of not including them at all).

Initial testing in a jupyter notebook seems to indicate that this behaves as expected:
- I can do `from_backend` with a non-trivial `subsystem_list`, and the backend is built without issue.
- I haven't yet done any verifications of correctness, but it seems like we can run sims and compute measurement results as well.

To do:
- Add number of qubits to the backend wherever it makes sense.
- Do some initial verification that the calculations are correct.
- The `subsystem_labels` option was used to raise errors of non-existent subsystems were measured. As this no longer exists, it would be nice to still warn users if their schedule measures a trivial subsystem. 
    - This could just be done by checking the dimension of the measured subsystem.
- Pre-existing tests need to be modified.

Question:
- This will remove the `subsystem_labels` option. I'm not sure how this could be deprecated, because it will literally serve no purpose and have no effect anymore.

PR #297, opened 2023-12-18
fix `sub_dims` bug in `_get_iq_data`

rupeshknn:
### Summary
Backends instantiated using `DynamicsBackend.from_backend` do not work with `meas_level` arg in `DynamicsBackend.run`

### Details and comments
```python
from qiskit import pulse
from qiskit import QuantumCircuit
from qiskit.qobj.utils import MeasLevel
from qiskit_ibm_provider import ibm_provider
from qiskit_dynamics import DynamicsBackend

provider = ibm_provider.IBMProvider()
kyoto = provider.get_backend('ibm_kyoto')
sim_backend = DynamicsBackend.from_backend(kyoto,subsystem_list=[0]) #single qubit backend


qc = QuantumCircuit(1,1)
qc.h(0)
qc.measure([0],[0])

with pulse.build() as h_q0:
    pulse.play(
        pulse.library.Gaussian(duration=256, amp=0.2, sigma=50, name="custom"),
        pulse.DriveChannel(0)
    )
qc.add_calibration("h", qubits=[0], schedule=h_q0)


result = sim_backend.run(qc, meas_level=MeasLevel.KERNELED)

# ValueError: maximum supported dimension for an ndarray is 32, found 127
```

which is raised by [qiskit_dynamics/backend/backend_utils.py:239 --> `probabilities_tensor = probabilities.reshape(list(reversed(subsystem_dims)))`](https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/main/qiskit_dynamics/backend/backend_utils.py#L239)

This PR solves this error by removing all 1d (trivial) systems from the subsystem dims within the `_get_iq_data` function.

More generally, however, `DynamicsBackend.from_backend` must assign a `Target` object with the correct number of qubits to the backend and have the `subsystem_dims` of only the required length.

### DIFF
--- a/qiskit_dynamics/backend/backend_utils.py
+++ b/qiskit_dynamics/backend/backend_utils.py
@@ -234,7 +234,7 @@ def _get_iq_data(
         QiskitError: If number of centers and levels don't match.
     """
     rng = np.random.default_rng(seed)
-    subsystem_dims = state.dims()
+    subsystem_dims = [dim for dim in state.dims() if dim != 1]
     probabilities = state.probabilities()
     probabilities_tensor = probabilities.reshape(list(reversed(subsystem_dims)))
 
--- a/test/dynamics/backend/test_backend_utils.py
+++ b/test/dynamics/backend/test_backend_utils.py
@@ -281,7 +281,7 @@ class Test_get_iq_data(QiskitDynamicsTestCase):
         self.assertDictEqual(counts, {"0": 100})
 
     def test_multi_qubit_predict(self):
-        """Multi_qubit predict test case."""
+        """Multi qubit predict test case."""
         iq_data = _get_iq_data(
             state=Statevector(np.array([0.5, 1, 0, 0]) / np.sqrt(1.25)),
             measurement_subsystems=[0, 1],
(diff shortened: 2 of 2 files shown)


rupeshknn:
I just realized, the problem isn't just with `meas_level=MeasLevel.KERNELED`. Even doing `result = sim_backend.run(qc)` doesn't work as similar logic is evoked in Qiskit quantum-info. The 127 qubit open-access backends have made it infeasible to use `DynamicsBackend.from_backend`

DanPuzzuoli:
Thanks @rupeshknn 

If I remember correctly setting the dimension to `1` for the trivial subsystems was done as a hack to get around some other issue. The removal of this doesn't seem to make any other issues suddenly appear?

DanPuzzuoli:
Yeah so the original change setting "trivial systems" to dimension `1` is here in PR #248. How does this PR handle the labelling of qubits once you've chosen some subset in `from_backend`?

DanPuzzuoli (changes requested):
Thank you. The only other thing I'd request is if you can add a test case to `test.dynamics.backend.test_dynamics_backend.TestDynamicsBackend_from_backend` that would have caught this. I guess there aren't any test cases that actually compute outcomes. You could add one that integrates over a very short time interval (to avoid costly simulation) and computes the results. Thoughts?

rupeshknn:
Hey Dan, for this PR, the bug isn't from `DynamicsBackend.from_backend`. It has to do with how IQ data is computed. I've already added a test that ensures trivial systems are not measured for IQ data-related means options. This works irrespective of whether the trivial subsystems were provided by the user or via `from_backend`.

So, I don't see the need for another test. if you disagree, could you point out what exactly would this test be checking for?

DanPuzzuoli:
> Hey Dan, for this PR, the bug isn't from `DynamicsBackend.from_backend`. It has to do with how IQ data is computed. I've already added a test that ensures trivial sub-systems are not measured for IQ data-related means options. This works irrespective of whether the trivial subsystems were provided by the user or via `from_backend`.
> 
> So, I don't see the need for another test. if you disagree, could you point out what exactly would this test be checking for?

Ah I see - yes you're right. Apologies, I rushed through looking at this in a scramble to reply to a bunch of things after getting back from holidays.

DanPuzzuoli (approved):

PR #320, opened 2024-02-08
Merge arraylias integration branch into main

DanPuzzuoli:
### Summary

I believe all tests, linting, and docs should build.

### Details and comments

Note: We should try to rebase and merge here, to preserve the history of the feature branch.

PR #326, opened 2024-02-20
Fix JAX pulse -> signals issue

DanPuzzuoli:
### Summary

This is an issue on the `main` branch that failed after merging the arraylias integration branch in PR #320 . The tests were passing in the PR so I'm assuming this error has been introduced after updating some dependency.

An error is now occurring with JAX when building a `SymbolicPulse` inside of a `jit` compilation - this error is causing the docs to fail, as well as the test 
```
python -m unittest test.dynamics.pulse.test_pulse_to_signals.TestPulseToSignalsJAXTransformations.test_InstructionToSignals_with_JAX
```
(among some others).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.8'
+        python-version: '3.10'
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


to24toro (approved):
Thank you for looking into the bugs.

to24toro (changes requested):
I will start to work on this problem at #327.

DanPuzzuoli:
Closed in favour of #327

PR #327, opened 2024-02-21
Remove Array class from pulse_to_signals

to24toro:
### Summary
Removed Array class.
It checks whether the pulse parameters type is Jax.core.Tracer.

### Details and comments
Related to #326.


DanPuzzuoli (approved):
Looks good. Thanks @to24toro !

PR #328, opened 2024-02-23
Remove old test class TestJaxBase

DanPuzzuoli:


to24toro:
Thanks Dan.
This PR looks good for me.
In addition, Can we remove ArrayTest class e.g. ArrayNumpyTestBase?

DanPuzzuoli:
> Thanks Dan. This PR looks good for me. In addition, Can we remove ArrayTest class e.g. ArrayNumpyTestBase?

That's a good question. We're currently using it for `Signal` testing, and I was thinking we could keep it there since we will be deprecating `Array` and not getting rid of it right away. On the other hand, the usage of `Array` throughout the package generally is already broken, so it may not be worth maintaining tests for the signal class. What do you think?

to24toro:
Thanks. I consider that it is better to drop Array test cases at the same time that we will deprecate Array class.

to24toro (approved):

PR #329, opened 2024-02-24
Fix complex cast warning

DanPuzzuoli:
### Summary

This PR modifies the code to eliminate complex -> real casting warnings that have popped up at various places.

### Details and comments

To do this I've changed code in a couple places in which a `tensordot` occurs between a real and complex array. The change is essentially converting `np.tensordot(x, y)` to `np.tensordot(x, y.real) + 1j * np.tensordot(x, y.imag)` (where `x` is typically real and `y` is the complex array).

This eliminated all of the warnings - I'm not 100% sure about efficiency but I'm hoping it shouldn't change anything, as I assume the cost of arithmetic of `np.tensordot(x, y)` for `y` complex is twice that of when `y` is imaginary.


to24toro (approved):
LGTM!

PR #330, opened 2024-02-26
Resolving nyquist frequency warning in tests

DanPuzzuoli:
### Summary

Fixed a test that was setting a frequency out of the nyquist frequency range of the channel.


to24toro (approved):
LGTM!

PR #332, opened 2024-02-28
Update models API docs

DanPuzzuoli:
### Summary

Updates the `models` module API docs to reflect the change of ``evaluation_mode`` to ``array_library`` and ``vectorized``.

### Details and comments

In the interest of minimizing further documentation maintenance, I also slightly changed the `BaseGeneratorModel` abstract base class to implement the `array_library` property. 
- Previously, this property was kept abstract, and was implemented by all subclasses. 
- Now, the `BaseGeneratorModel` class has an `__init__` that sets `self._array_library`, and implements a concrete `array_library` property.

This enabled deleting the `array_library` property for `GeneratorModel`, `HamiltonianModel`, and `LindbladModel`. As a result, changes to documentation of this property only need to be made in one place.

### DIFF
--- a/docs/userguide/how_to_configure_simulations.rst
+++ b/docs/userguide/how_to_configure_simulations.rst
@@ -204,7 +204,7 @@ with extra emphasis on the following:
 .. note::
 
     As stated in the
-    :ref:`evaluation modes section of the Models API documentation <evaluation modes>`,
+    :ref:`model evaluation section of the Models API documentation <model evaluation>`,
     when using a sparse evaluation mode, to preserve sparsity, it is recommended to
     only use *diagonal* rotating frames, which can be specified as a 1d array to the
     ``rotating_frame`` kwarg of :class:`.Solver` instantiation.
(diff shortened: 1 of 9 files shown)


to24toro (commented):
It looks almost good.
The below is a subtle suggestion.

https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/2b91912c78693495f5dbb2f3edcd8cc906b032e0/qiskit_dynamics/models/operator_collections.py#L204

-> ArrayLike that acts on states ``y`` via multiplication.

DanPuzzuoli:
> It looks almost good. The below is a subtle suggestion.
> 
> https://github.com/Qiskit-Extensions/qiskit-dynamics/blob/2b91912c78693495f5dbb2f3edcd8cc906b032e0/qiskit_dynamics/models/operator_collections.py#L204
> 
> -> ArrayLike that acts on states `y` via multiplication.

Good find - this is actually the scipy sparse version of the class so I changed "Array" -> "scipy sparse matrix".

to24toro (approved):

PR #340, opened 2024-03-06
Update array library and simulation configuration user guides

DanPuzzuoli:
### Summary

This PR updates two userguides:
1. The "how-to use JAX" userguide, which is renamed more generally to: `How-to use different array libraries and types with Qiskit Dynamics`
2. The "how-to customize simulations" userguide entry.

These are both updated for the new Arraylias integration.

### Details and comments

Major changes:
1. The first userguide has been greatly simplified. It used to be about how to use `Array` to enable JAX transformations, which is no longer necessary. It now just shows that you can use numpy or JAX, points to the aliases built in qiskit dynamics, and gives a demonstration of compiling a JAX simulation.
2. The second userguide has less significant changes, and is just a cosmetic update to align with the new `arraylias` integration.


to24toro (approved):
LGTM!

PR #344, opened 2024-03-15
Add stable docs build

DanPuzzuoli:
### Summary

Adding `stable/0.4` to the docs build.

### DIFF
--- a/.github/workflows/docs_stable.yml
+++ b/.github/workflows/docs_stable.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Build docs stable
         env:
           QISKIT_DOCS_BUILD_TUTORIALS: 'always'
-        run: EXPERIMENTS_DEV_DOCS=1 PROD_BUILD=1 tox -e docs
+        run: tox -e docs
       - name: Bypass Jekyll Processing # Necessary for setting the correct css path
         run: touch docs/_build/html/.nojekyll
       - name: Set current version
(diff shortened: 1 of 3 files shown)


.github/workflows/docs_stable.yml
```diff
@@ -0,0 +1,38 @@
...
+          python-version: '3.10'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -U virtualenv setuptools wheel tox
+          sudo apt-get install graphviz pandoc
+      - name: Build docs stable
+        env:
+          QISKIT_DOCS_BUILD_TUTORIALS: 'always'
+        run: EXPERIMENTS_DEV_DOCS=1 PROD_BUILD=1 tox -e docs
```
arnaucasau, line 28: Nit: we can remove this environment variables
```suggestion
        run: tox -e docs
```

arnaucasau (commented):
Great feature to add! All looks good to me. In addition to the new workflow, we need to add the following lines in `.github/workflows/docs.yml` to avoid removing the stable folder when re-deploying the current version docs.
```
     clean-exclude: |
       stable/*
```

Here you can find an example from qiskit-experiments: https://github.com/arnaucasau/qiskit-experiments/blob/3796627517c2480bddceb80346c28d91f9c4c2be/.github/workflows/docs_release.yml#L35

arnaucasau (approved):
LGTM! :)

PR #349, opened 2024-03-21
Remove class Array text from models and solvers 

to24toro:
### Summary
Found some texts including class `Array` representation and removed them.

### DIFF
--- a/qiskit_dynamics/models/generator_model.py
+++ b/qiskit_dynamics/models/generator_model.py
@@ -117,7 +117,7 @@ class GeneratorModel(BaseGeneratorModel):
 
         G(t) = \sum_i s_i(t) G_i + G_d
 
-    where the :math:`G_i` are matrices (represented by :class:`Operator` or array libraries),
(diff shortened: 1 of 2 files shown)


qiskit_dynamics/models/generator_model.py
```diff
@@ -117,7 +117,7 @@ class GeneratorModel(BaseGeneratorModel):
 
         G(t) = \sum_i s_i(t) G_i + G_d
 
-    where the :math:`G_i` are matrices (represented by :class:`Operator` or :class:`Array` objects),
+    where the :math:`G_i` are matrices (represented by :class:`Operator` or array libraries),
```
DanPuzzuoli, line 120:
```suggestion
    where the :math:`G_i` are matrices (represented by an :class:`Operator` or array),
```

DanPuzzuoli (changes requested):
Nice catch, just suggested a minor change to wording.

DanPuzzuoli (approved):

PR #353, opened 2024-04-09
Default t_span available for `DynamicsBackend.solve`

arthurostrauss:
### Summary
This PR introduces the possibility to use `DynamicsBackend.solve()`without having to specify a `t_span`argument explicitly. If it is not specified, the default `t_span`is inferred from the provided `solve_input` to create a `t_span`of the form [0, solve_input.duration] for each solve_input in the provided list. This can be useful for simulating easily a full `QuantumCircuit`s as the process of converting them to `Schedule` is already done behind the scenes.

PR #354, opened 2024-04-09
Fix qutip Qobj handling for qutip 5.0.0

DanPuzzuoli:
### Summary

As pointed out in #353 , the handling of the qutip `Qobj` type by `DYNAMICS_NUMPY.asarray` was broken with the release of qutip 5.0.0. In the past, qutip `Qobj` instances always contained customized `csr_matrix` instances, so we would always consume them by explicitly converting them to a `csr_matrix`. Qutip 5.0.0 adds a "data layer" abstraction, allowing `Qobj` instances to store different underlying array types (e.g. dense `np.ndarray`). This data layer has been written to be extensible, and hence moving forward we have no guarantees on the type of array data within a `Qobj`.

As a result of this, I've changed `DYNAMICS_NUMPY.asarray(qobj)` to return `qobj.full()`, which returns a dense `np.ndarray`. This works regardless of underlying array type, so seems like a safe future-proof way of solving this issue.

### Details and comments

I've also added an upgrade note outlining this change.

### DIFF
--- a/qiskit_dynamics/arraylias/register_functions/asarray.py
+++ b/qiskit_dynamics/arraylias/register_functions/asarray.py
@@ -29,13 +29,7 @@ def _isinstance_qutip_qobj(obj):
     Returns:
         Bool: True if obj is qutip Qobj
     """
-    if (
-        type(obj).__name__ == "Qobj"
-        and hasattr(obj, "_data")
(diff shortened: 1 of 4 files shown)


to24toro (approved):
LGTM!

PR #367, opened 2024-11-15
Change level 1 memory result data type to match Qiskit

wshanks:
In Qiskit, it is specified that memory should be a list (https://github.com/Qiskit/qiskit/blob/b9d5c9c6aeb4568b6e9fba8943517ebed300886d/qiskit/result/models.py#L38). While most code can work with either lists or numpy arrays, some can not, including BackendSamplerV2 in Qiskit (see
https://github.com/qiskit-community/qiskit-experiments/issues/1487). Here the numpy array data is converted to a nested list using `tolist()`.

### DIFF
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -920,7 +920,7 @@ def default_experiment_result_function(
             measurement_data = np.average(measurement_data, axis=0)
 
         # construct results object
-        exp_data = ExperimentResultData(memory=measurement_data)
+        exp_data = ExperimentResultData(memory=measurement_data.tolist())
         return ExperimentResult(
             shots=backend.options.shots,
             success=True,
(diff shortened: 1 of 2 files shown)


wshanks:
I didn't test this yet. Hoping it passes the CI without changes :crossed_fingers: but some tests might assume the array format. This is a little bit of a breaking change potentially but I don't know how much effort to put into backwards compatibility for this (like an option to keep the old format?).

IvanaGyro:
Is there any workaround for this issue now?

IvanaGyro (approved):
You may also want to convert the strings to hex string `memory = [hex(int(byte, 2)) for byte in measurement_data]` because SamplerV2 expects the measurement results representing in hex strings.

DanPuzzuoli:
> You may also want to convert the strings to hex string `memory = [hex(int(byte, 2)) for byte in measurement_data]` because SamplerV2 expects the measurement results representing in hex strings.

@IvanaGyro can you create an issue about this? I'm going to merge this this branch now.

DanPuzzuoli (approved):

wshanks:
I created #373 for @IvanaGyro's suggestion. It is necessary for level 2 data experiments to work, but I just did the simplest implementation. It might break some other things that I mention.

PR #372, opened 2025-01-23
Remove setting of target.num_qubits in DynamicsBackend init

### LINKED ISSUE #366: `DynamicsBackend` sets `Target.num_qubits` which is not settable in Qiskit 1.3
### Informations

- **Qiskit Dynamics version**: 0.5.1
- **Python version**: 3.12
- **Operating system**: Fedora Linux 41

`AttributeError` when trying to create a `DynamicsBackend`:

```

File ~/envs/qiskit/lib/python3.12/site-packages/qiskit_dynamics/backend/dynamics_backend.py:219, in DynamicsBackend.__init__(self, solver, target, **options)
    216     target.add_instruction(Measure(), measure_properties)
    218 target.dt = solver._dt
--> 219 target.num_qubits = len(self.options.subsystem_dims)
    221 self._target = target

AttributeError: attribute 'num_qubits' of 'qiskit._accelerate.target.BaseTarget' objects is not writable
```

Run the following code with Qiskit 1.3:

```python
from qiskit.quantum_info import Operator
from qiskit_dynamics import DynamicsBackend, Solver
import numpy as np

nu_z = 10.
nu_x = 1.
nu_d = 9.98 # Almost on resonance with the Hamiltonian's energy levels difference, nu_z

X = Operator.from_label('X')
Y = Operator.from_label('Y')
Z = Operator.from_label('Z')
s_p = 0.5 * (X + 1j * Y)

solver = Solver(
    static_hamiltonian=.5 * 2 * np.pi * nu_z * Z,
    hamiltonian_operators=[2 * np.pi * nu_x * X],
    hamiltonian_channels=["d0"],
    channel_carrier_freqs={"d0": nu_x},
    dt=0.01,
)

dynamics_backend = DynamicsBackend(solver=solver, subsystem_dims=[2])
```

Successful `DynamicsBackend` creation

### Suggested solutions

`num_qubits` can be passed to `Target` instead. I am not sure about the case when a target is passed in to `DynamicsBackend`. ~I think it should just leave the target alone and base the number of qubits on the subsystem dimensions. Perhaps it could make a new target and only copy the instructions related to the qubits in the subsystem.~ Thinking about it more, I think it would be most consistent to build a new qubit with just the instructions for the qubits in the subsystem to be simulated. Otherwise, number of qubits will be inconsistent in different parts of the code.

DanPuzzuoli:
### Summary

Closes #366 

There are many remaining issues with the backend stuff, but this appears to solve this specific problem and gets the relevant tests to pass.

### DIFF
--- a/qiskit_dynamics/backend/dynamics_backend.py
+++ b/qiskit_dynamics/backend/dynamics_backend.py
@@ -216,7 +216,6 @@ class DynamicsBackend(BackendV2):
             target.add_instruction(Measure(), measure_properties)
 
         target.dt = solver._dt
-        target.num_qubits = len(self.options.subsystem_dims)
 
         self._target = target
 
--- a/releasenotes/notes/backend-num-qubits-700d8c03e08b324a.yaml
+++ b/releasenotes/notes/backend-num-qubits-700d8c03e08b324a.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Setting of the ``num_qubits`` ``Target`` attribute in ``DynamicsBackend`` initialization has
+    been removed. Users will now need to set this at ``Target`` instantiation if needed.


DanPuzzuoli (commented):
This appears to resolve the backend test issues, though does not solve the issue with the tutorial.

PR #379, opened 2025-02-20
Update all DynamicsBackend code/tests to work with hex for meas level 2

### LINKED ISSUE #375: Measurement level 2 hex
PR #373 converted the meas level 2 results to be returned in hexadecimal. This has resulted in tests failing:

- Some are simply that the checks are done on binary strings, which is easy to fix, but:
- Others are due to the fact that the `DynamicsBakend` has the option to return measurements of higher energy states, resulting in non-binary result strings, which fail in the conversion (which assumes base 2).

Some solutions:

- Add the option to return results in the original bit strings
- Disable measurement of higher level energy states (I think not preferred)

For now I'm going to leave this for general discussions about pulse migration.

DanPuzzuoli:
### Summary

Closes #375 

### Details and comments

A current issue I'm having with resolving all test cases is that the `Result` object, in a test like 
```
test.dynamics.backend.test_dynamics_backend.TestDynamicsBackend.test_schedule_memory_slot_num
```
actually converts things back into integers (as strings) when calling `get_counts`.


DanPuzzuoli:
To fix the tests I decided to directly look at the internal memory of the `Result` objects (through `Result.to_dict()`), which are stored in hex, and validate correctness. We can't control the behaviour the `Result` object, and at least this way we can validate the outputs of `DynamicsBackend`.

